PR #19, opened 2022-03-29
Docs: Add simple documentation for `WindowProperties`, `Button` and `Text`

NinjaCheetah:
Mostly just parameters but this should be sufficient for using these structs until full documentation is written.


stackotter (approved):
Looks good 👍

PR #20, opened 2022-03-29
Add basic "Getting Started" documentation

NinjaCheetah:
Here's some basic documentation for getting started, basically all pulled from the README. I mainly thought this would be helpful because it means you can find all of the info you need to start working on the documentation site. Definitely can reword anything that needs to be reworded.


stackotter (approved):
Thanks, looks good!

PR #22, opened 2022-08-29
Create .spi.yml

finestructure:
This should address the Linux build failures using the image define here: https://gitlab.com/finestructure/spi-images


stackotter (approved):
Thanks for this! Hopefully it works 🤞

PR #28, opened 2023-05-13
SwiftUIs NavigationStack using GtkStack

Amzd:
Rough implementation of SwiftUIs NavigationStack using GtkStack

The main difference between Apples NavigationStack and this is that we don't have environment so you have to store a path in State and give it to both NavigationStack and NavigationLink.

Another API difference is that .navigationDestination is defined on NavigationStack instead of on View. This means SwiftUI code will look a bit different:

```diff
NavigationStack(path: $path) {
    List {
-        NavigationLink("Mint", value: Color.mint)
-        NavigationLink("Pink", value: Color.pink)
+        NavigationLink("Mint", value: Color.mint, path: $path)
+        NavigationLink("Pink", value: Color.pink, path: $path)
    }
-    .navigationDestination(for: Color.self) { color in
-        ColorDetail(color: color)
-    }
}
+.navigationDestination(for: Color.self) { color in
+    ColorDetail(color: color)
+}
```

TODO:

- [x] Fix NavigationLink
- [x] Remove unnecessary code 
- [x] Document
- [x] Create Example


Sources/SwiftCrossUI/Views/NavigationLink.swift
```diff
@@ -0,0 +1,25 @@
+/// A view that controls a navigation presentation.
```
stackotter: This description doesn't make much sense to me. Maybe something like `A navigation primitive that appends a value to the current navigation path on click`?

Sources/SwiftCrossUI/Views/NavigationPath.swift
```diff
@@ -0,0 +1,36 @@
...
+    public var isEmpty: Bool {
+        path.isEmpty
+    }
+
+    /// The number of elements in this path.
+    public var count: Int {
+        path.count
+    }
+
+    public init() {}
```
stackotter, line 16: Maybe document as `Creates an empty navigation path` just to be clear that it doesn't initialise it to `["root"]` or some other default value

Sources/SwiftCrossUI/Views/NavigationStack.swift
```diff
@@ -0,0 +1,188 @@
+import Gtk
+
+/// Type to indicate the root of the NavigationStack. This is private to prevent root accidentally showing instead of a detail view.
+private struct NavigationStackRootPath: Hashable {}
```
stackotter, line 4: If the root view is only ever the first element of the path, could you just show the root view when the path is empty instead? That would mean there doesn't need to be a special type for it.

Sources/SwiftCrossUI/Views/NavigationStack.swift
```diff
@@ -0,0 +1,188 @@
...
+        }
+    }
+
+    /// Associates a destination view with a presented data type for use within a navigation stack.
+    ///
+    /// - Parameters:
+    ///   - data: The type of data that this destination matches.
+    ///   - destination: A view builder that defines a view to display when the stack’s navigation state contains a value of type data. The closure takes one argument, which is the value of the data to present.
+    ///
+    /// Add this view modifer to describe the view that the stack displays when presenting a particular kind of data. Use a `NavigationLink` to present the data.
```
stackotter: I think discussion is meant to go before the `Parameters` section? That's what I always do at least for consistency, not sure whether DocC actually cares

Sources/SwiftCrossUI/Views/NavigationStack.swift
```diff
@@ -0,0 +1,188 @@
...
+        transitionDuration = previous.transitionDuration
+        body = NavigationStackContent(previous.body.elements) {
+            if let previous = previous.body.child($0) {
+                // Either root or previously defined destination returned a view
+                return EitherViewContent(previous)
+            } else if let new = destination($0) {
+                // This destination returned a detail view for the current element
+                return EitherViewContent(new)
+            } else {
+                // Possibly a comming .navigationDestination will handle this path element
```
stackotter: By `comming` I think you mean `coming`, but `future` might make more sense

Sources/SwiftCrossUI/Views/NavigationStack.swift
```diff
@@ -0,0 +1,188 @@
...
+
+    public var elements: [any Hashable]
+    public var child: (any Hashable) -> Child?
+
+    func childOrCrash(for element: any Hashable) -> Child {
+        assert(child(element) != nil, """
+            Failed to find detail view for "\(element)",
+            make sure you have called .navigationDestination for this type.
+        """)
+        return child(element)!
```
stackotter: To get rid of the force unwrap you should do this instead:
```swift
guard let child = child(element) else {
    fatalError("Your error message...")
}
return child
```

It also prevents `child(element)` from being computed twice.

Sources/SwiftCrossUI/Views/NavigationStack.swift
```diff
@@ -0,0 +1,188 @@
...
+                break
+            }
+            let index = content.elements.startIndex.advanced(by: i)
+            node.update(with: content.childOrCrash(for: content.elements[index]))
+        }
+
+        let remaining = content.elements.count - storage.nodes.count
+        if remaining > 0 {
+            // Remove queued pages
+            for widget in storage.widgetsQueuedForRemoval {
```
stackotter: It's probably not super important, but you could instead remove the queued widgets whenever `remaining != 0`; that would mean there is only ever one invisible widget at any given time.

For example, if you pop one path component and then pop another one, the first detail view will already be invisible by the time you pop the second one so it can be removed.

Sources/SwiftCrossUI/Views/NavigationStack.swift
```diff
@@ -0,0 +1,188 @@
...
+        // Update pages
+        for (i, node) in storage.nodes.enumerated() {
+            guard i < content.elements.count else {
+                break
+            }
+            let index = content.elements.startIndex.advanced(by: i)
+            node.update(with: content.childOrCrash(for: content.elements[index]))
+        }
+
+        let remaining = content.elements.count - storage.nodes.count
```
stackotter, line 157: If the path changes from `[science, physics]` to `[humanities, english]` in a single update I think this implementation wouldn't work. Which is fine if only `append` and `pop` are used by the user, but either the user could do some weird stuff (storing old paths and restoring the path to an old value) or we could open up more ways to work with navigation paths that break this.

I think it'd probably be best to implement `update` under the assumption that the new path is not related to the old one (for ultimate flexibility). I have a possible implementation in mind (similar to the current one):

1. Remove widgets queued for removal
2. Compute how long the common prefix of the previous and new path is (e.g. `[science, physics]`, `[science, chemistry, science]` and `[science, science]` all have a common prefix of length one).
3. Remove all widgets from the previous path that are after the common prefix
4. Add all widgets from the new path that are after the common prefix

For example: given `previousPath = [a, b, c, d]` and `newPath = [a, b, e, f, g]`, remove `c` and `d`, and then add `e`, `f`, and `g`.

In cases where widgets need to be removed and widgets need to be added in a single update (such as the example above), I think you can still do the queued removal of popped widgets and transitions will work correctly. After the update you'll end up with a GtkStack of `[a, b, _c, _d, e, f, g]` (with `_c` and `_d` being queued for removal), and then at the start of the next update you'll remove `_c` and `_d` and it'll all work fine.

Sources/SwiftCrossUI/Views/NavigationStack.swift
stackotter: Actually, you could just remove them at the start of `update`, I don't there's any situation where it'd be incorrect to remove all queued widgets.

Sources/SwiftCrossUI/Views/NavigationStack.swift
stackotter: To find the common prefix you'll need to compare corresponding elements of the new and old paths. You actually only need to compare the types of the elements I think (which can be done with `type(of: previousPath[i]) == type(of: newPath[i])`).

Sources/SwiftCrossUI/Views/NavigationPath.swift
```diff
@@ -0,0 +1,36 @@
+/// A type-erased list of data representing the content of a navigation stack.
+public struct NavigationPath {
+    internal private(set) var previousPath: [any Hashable]?
```
stackotter: I was wondering, do we actually need the path elements to be hashable? I know that's what SwiftUI does, but I guess we might as well make the requirement looser unless you see a future reason not to

stackotter (changes requested):
Awesome work on the navigation implementation so far. I've just got a few requested changes, most of them are pretty small.

The biggest one is to do with handling arbitrary path changes in `update`, but I've got a way of handling those without changing your implementation too much I think.

Sources/SwiftCrossUI/Views/NavigationStack.swift
Amzd: Technically probably yes but this would deviate the implementation quite far from ForEach and I spent some time figuring out how to make that work but I can't crack it without making the logic hard to maintain.

Sources/SwiftCrossUI/Views/NavigationPath.swift
Amzd: Hashables are easily comparable between types, which is why I used them but we could probably use Any if you prefer.

We should think about making this maybe Codable though to support persistence: #29 

If you really feel like we should change what do you think of creating a separate issue for it and merge this PR first?

Sources/SwiftCrossUI/Views/NavigationStack.swift
Amzd: This was a bug fixed by switching from EiterViewContent to EitherView, mistake on my part. We don't need complicated path comparison as the EitherViews already take care of all this.

While debugging this I found a crash when switching from two items to 0 items. This was a bug in code copied from ForEach so I also fixed it in there.

Edit to address your concerns:

> I think it'd probably be best to implement update under the assumption that the new path is not related to the old one (for ultimate flexibility).

That is currently how it works. Every page in the stack is an EitherView, when the page changes from `[a, b]` to `[a, c]` the second page EitherView swaps from b to c by NavigationStack calling the `ViewGraphNode<EitherView<...>>.update(with:)` which calls the `EitherView<...>.update(_:children:)` which swaps out the b for the c view

Sources/SwiftCrossUI/Views/NavigationStack.swift
Amzd: fixed

Sources/SwiftCrossUI/Views/NavigationStack.swift
Amzd: fixed

Sources/SwiftCrossUI/Views/NavigationStack.swift
Amzd: yep, fixed

Sources/SwiftCrossUI/Views/NavigationStack.swift
Amzd: VS Code worked like this, I fixed it though.

Sources/SwiftCrossUI/Views/NavigationLink.swift
Amzd: Was copied from Apple, your description sounds good. Will change.

stackotter: Lol, sounds Appley

Sources/SwiftCrossUI/Views/NavigationStack.swift
stackotter: Interesting, maybe DocC doesn't care then, cause from memory Xcode generates it the other way around. Best to stay consistent though 👍

Sources/SwiftCrossUI/Views/NavigationPath.swift
stackotter: Yeah sure I'll merge this first.

stackotter:
> Hashables are easily comparable between types

What do you mean comparable between types? Are you referring to the fact that `"red".hashValue == MyEnum.red.hashValue`?

stackotter (dismissed):
Looks good to me!

stackotter (approved):

PR #30, opened 2023-05-17
ForEach rework

Amzd:
Reworked Box usage across modifiers so ForEach can retrieve the correct orientation for its context.

Previously the following would create vertical items in swift-cross-ui while in SwiftUI it generates the views inside ForEach horizontally. 

```swift
HStack {
  ForEach(...) {
    ...
  }
}
```

This PR aligns our ForEach with Apples SwiftUI behaviour. From here on out if you want to wrap content (eg ModifierView) without caring about orientation, use `GtkSingleChildBox`.

### DIFF
--- a/Sources/SwiftCrossUI/Views/PaddingView.swift
+++ b/Sources/SwiftCrossUI/Views/PaddingView.swift
@@ -1,17 +1,15 @@
-public struct PaddingView<Content: ViewContent>: View {
-    public var body: Content
+public struct PaddingView<Child: View>: View {
+    public var body: ViewContent1<Child>
     public var padding: Padding
 
-    init(_ content: Content, _ padding: Padding) {
-        body = content
+    init(_ content: Child, _ padding: Padding) {
+        body = ViewContent1(content)
         self.padding = padding
     }
 
-    public func asWidget(_ children: Content.Children) -> GtkBox {
-        let box = GtkBox(orientation: .vertical, spacing: 0)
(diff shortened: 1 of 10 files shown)


Sources/SwiftCrossUI/Views/PaddingView.swift
```diff
@@ -20,4 +18,11 @@ public struct PaddingView<Content: ViewContent>: View {
 
         return box
     }
+
+    public func update(_ box: GtkSingleChildBox, children: ViewGraphNodeChildren1<Child>) {
```
stackotter, line 22: Good catch noticing that it was missing an `update`

Sources/SwiftCrossUI/Views/PaddingView.swift
```diff
@@ -1,17 +1,15 @@
...
+        body = ViewContent1(content)
         self.padding = padding
     }
 
-    public func asWidget(_ children: Content.Children) -> GtkBox {
-        let box = GtkBox(orientation: .vertical, spacing: 0)
-        for widget in children.widgets {
-            box.add(widget)
-        }
+    public func asWidget(_ children: ViewGraphNodeChildren1<Child>) -> GtkSingleChildBox {
```
stackotter, line 10: I like that it's now explicitly ViewGraphNodeChildren1 👍

stackotter (approved):
Looks great! I'll just locally test it quickly first, but I don't see any problems

PR #32, opened 2023-05-17
Codable NavigationPath without _mangledTypeName

Amzd:
Proof that it is possible.

Negatives:

- you can see it is decoding when you set a NavigationPath thats not decoded yet, (because NavigationPath returns an empty path instead of returning not yet decoded entries) 
  - could possibly be fixed by showing an empty view for LazyDecodableEntry but it is not guaranteed that the entry will ever decode so idk
- if you get an item that didnt decode (due to dev changing types) in persistence, it will never remove it. We just encode them back into the data again if an entry was not decoded.
- entries.count and NavigationPath.count are not the same, idk if this could be confusing
- changes state during view loop (because it decodes based on the NavigationStack view) 
  - could probably be fixed

TODO:
- [x] document


Examples/Navigation/NavigationApp.swift
```diff
@@ -1,16 +1,19 @@
...
-enum SubjectArea: Hashable {
+enum SubjectArea: Codable {
     case science
     case humanities
 }
 
-enum ScienceSubject: Hashable {
-    case physics
-    case chemistry
+struct Science {
```
stackotter: I assume this change is just for testing decoding of non-top-level types?

Sources/SwiftCrossUI/Views/NavigationPath.swift
```diff
@@ -16,17 +27,93 @@ public struct NavigationPath {
...
+            updated.hasPendingDecodableEntries = didFindDecodableEntry
+            return updated
+        }
+
+        return nil
+    }
+
+    private func typeName<T>(for: T.Type) -> String {
+        // If this also does not work we can do String(describing:) but that wont
+        // include parent of nested class and module
```
stackotter: I haven't tested _typeName yet, but if `_typeName` isn't available `String(reflecting:)` is what we want I think, it gives you `NavigationExample.Science.Subject` for example.

stackotter (commented):
Once avoiding state updates is figured out, this should be a pretty good solution

Examples/Navigation/NavigationApp.swift
Amzd: correct

Sources/SwiftCrossUI/Views/NavigationPath.swift
Amzd: _typeName should be available

Amzd:
I've been thinking about this and I don't think it is possible without a state update, since we dont have onAppear or anything that we can use as callback

Examples/Navigation/NavigationApp.swift
```diff
@@ -39,6 +42,16 @@ struct NavigationApp: App {
         VStack {
             Text("Change transition duration")
             Slider(state.$transitionDuration, minimum: 0, maximum: 3)
+            Button("Test decoded Path", action: {
```
stackotter: I think this should butotn probably be removed from the example (and maybe put in a test later on once I setup unit testing)

Examples/Navigation/NavigationApp.swift
stackotter: This enum should probably be renamed to what it was before (the examples are meant to show how to use the library, the unit tests can test the edge cases once I set them up).

Sources/SwiftCrossUI/Views/NavigationLink.swift
```diff
@@ -3,23 +3,20 @@
...
     private let label: String
     /// The value to append to the navigation path when clicked.
-    private let value: AnyHashable
+    private let value: any Codable
     /// The navigation path to append to when clicked.
     private let path: Binding<NavigationPath>
 
     /// Creates a navigation link that presents the view corresponding to a value in the NavigationStack that uses the same path.
-    public init(_ label: String, value: AnyHashable, path: Binding<NavigationPath>) {
+    public init<C: Codable>(_ label: String, value: C, path: Binding<NavigationPath>) {
```
stackotter, line 19: I think the generic isn't necessary, you can just use `any Codable`

Sources/SwiftCrossUI/Views/NavigationPath.swift
```diff
@@ -1,33 +1,150 @@
...
     }
 
     /// Creates an empty navigation path.
     public init() {}
 
-    /// Appends a new value to the end of this path.
-    public mutating func append<H: Hashable>(_ component: H) {
-        path.append(component)
+    /// Appends a new value to the end of the path.
+    public mutating func append<C: Codable>(_ component: C) {
```
stackotter, line 49: Also can probably just be `any Codable` instead of a generic

stackotter (changes requested):
Just a few requested changes, looks good other than those

Sources/SwiftCrossUI/Views/NavigationPath.swift
Amzd: I dont mind either way but Apples SwiftUI does it this way and the WWDC video on any says to use generics instead of any where possible: 
image
https://developer.apple.com/videos/play/wwdc2022/110354/?time=2130 (35:30)

This might have been fixed since then in swift 5.8? Idk

stackotter (approved):
Looks good

PR #35, opened 2023-06-02
NavigationSplitView

Amzd:
Currently `NavigationSplitView` does not support any `.navigationDestination` api unlike the SwiftUI counterpart but it introduces the basic split view built on Gtk Paned.

This PR also adds `Spacer`

### DIFF
--- a/Examples/Split/SplitNavigationApp.swift
+++ b/Examples/Split/SplitNavigationApp.swift
@@ -1,0 +1,127 @@
+import Foundation
+import SwiftCrossUI
+
+enum SubjectArea {
+    case science
+    case humanities
+}
+
+enum ScienceSubject {
+    case physics
+    case chemistry
+}
+
+enum HumanitiesSubject {
+    case english
+    case history
+}
+
+enum Columns {
+    case two
+    case three
+}
+
+class NavigationAppState: Observable {
+    @Observed var selectedArea: SubjectArea?
+    @Observed var selectedDetail: Any?
+    @Observed var columns: Columns = .two
+}
+
+@main
+struct NavigationApp: App {
+    let identifier = "dev.stackotter.SplitNavigationApp"
+
+    let state = NavigationAppState()
+
+    let windowProperties = WindowProperties(
+        title: "SplitNavigation",
+        defaultSize: WindowProperties.Size(600, 250)
+    )
+
+    var body: some ViewContent {
+        switch state.columns {
+        case .two:
+            doubleColumn
+        case .three:
+            tripleColumn
+        }
+    }
+
+    /// Example view for a two column NavigationSplitView
+    var doubleColumn: some View {
+        NavigationSplitView {
+            VStack {
+                Button("Science") { state.selectedArea = .science }
+                Button("Humanities") { state.selectedArea = .humanities }
+                Spacer()
+                Button("Switch to 3 column example") { state.columns = .three }
+            }.padding(10)
+        } detail: {
+            VStack {
+                switch state.selectedArea {
+                case .science:
+                    Text("Science")
+                        .padding(.bottom, 10)
+                case .humanities:
+                    Text("Humanities")
+                        .padding(.bottom, 10)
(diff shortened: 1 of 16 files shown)


Sources/Gtk/Widgets/Paned.swift
```diff
@@ -0,0 +1,49 @@
...
+
+open class Paned: Widget, Orientable {
+    public init(orientation: Orientation) {
+        super.init()
+        widgetPointer = gtk_paned_new(orientation.toGtkOrientation())
+    }
+
+    public var startChild: Widget? {
+        didSet {
+            gtk_paned_set_start_child(opaquePointer, startChild?.castedPointer())
```
Amzd, line 12: hmm do we need to unset the parent of the previous child widget?

stackotter: > hmm do we need to unset the parent of the previous child widget?

I think so, to keep the Swift side of the Gtk view hierarchy up to date with the C side

Examples/Split/SplitNavigationApp.swift
```diff
@@ -0,0 +1,127 @@
...
+}
+
+class NavigationAppState: Observable {
+    @Observed var selectedArea: SubjectArea?
+    @Observed var selectedDetail: Any?
+    @Observed var columns: Columns = .two
+}
+
+@main
+struct NavigationApp: App {
```
stackotter: Should be renamed to `SplitNavigationApp`

Package.swift
```diff
@@ -77,6 +77,10 @@ let package = Package(
             name: "NavigationExample",
             targets: ["NavigationExample"]
         ),
+        .executable(
+            name: "SplitExample",
```
stackotter, line 81: Either rename this to match `SplitNavigationApp` or the other way around (I think `SplitExample` is suitable, so renaming to `SplitApp` would be fine).

Sources/SwiftCrossUI/Views/NavigationSplitView.swift
```diff
@@ -0,0 +1,44 @@
...
+    public init(@ViewContentBuilder sidebar: () -> SideBar,
+                @ViewContentBuilder detail: () -> Detail) {
+        body = ViewContent3(sidebar(), EmptyViewContent(), detail())
+    }
+}
+
+extension NavigationSplitView {
+    public init(@ViewContentBuilder sidebar: () -> SideBar,
+                @ViewContentBuilder content: () -> MiddleBar,
+                @ViewContentBuilder detail: () -> Detail) {
```
stackotter: Wrap arguments like this instead:
```swift
public init(
    @ViewContentBuilder sidebar: () -> SideBar,
    @ViewContentBuilder content: () -> MiddleBar,
    @ViewContentBuilder detail: () -> Detail
) {
```

Sources/SwiftCrossUI/Views/NavigationSplitView.swift
```diff
@@ -0,0 +1,44 @@
...
+            next.shrinkEndChild = false
+            next.position = 0
+            next.expandVertically = true
+
+            widget.endChild = next
+        }
+
+        widget.shrinkStartChild = false
+        widget.shrinkEndChild = false
+        widget.position = 0 // sets
```
stackotter: what does the comment mean?

stackotter (changes requested):
Looks good so far! This is going to be very useful. Just a few requested changes

Sources/SwiftCrossUI/Views/NavigationSplitView.swift
Amzd: updated to 
```swift
// Set the position to the farthest left possible.
// TODO: Allow setting the default offset (SwiftUI api: `navigationSplitViewColumnWidth(min:ideal:max:)`).
//       This needs frame modifier to be fledged out first
```

stackotter (approved):
Looks good to me 👍

PR #36, opened 2023-06-02
Modifiers update

Amzd:
- Made modifier views private, 
- Aligned padding edge to SwiftUI naming (Side to Edge) , 
- Allow multiple edges in padding modifier just like SwiftUI using OptionSet, 
- Fixed frame modifier now working with minimum sizes using set_size_request

TODO: For max and ideal frame size we need constraints, but there are 0 tutorials on how to use these.
ConstraintRelation has less than, equal, greater than: https://docs.gtk.org/gtk4/enum.ConstraintRelation.html
`gtk_constraint_new` even allows for `strength` so maybe we can allow Text to expand with low strength.


Sources/Gtk/Widgets/Widget.swift
```diff
@@ -308,4 +308,17 @@ open class Widget {
             gtk_widget_set_vexpand(castedPointer(), newValue ? 1 : 0)
         }
     }
+
```
stackotter: Remove the double newlines

Sources/SwiftCrossUI/Values/Edge.swift
```diff
@@ -0,0 +1,27 @@
+/// An enumeration to indicate one edge of a rectangle.
+public enum Edge: Int8, CaseIterable, Hashable {
+    case top
+    case bottom
+    case leading
+    case trailing
+
+    /// An efficient set of Edges.
+    public struct Set: OptionSet, Hashable {
+        public let rawValue : Int8
```
stackotter: Remove the leading space before the colon

Sources/SwiftCrossUI/Values/Edge.swift
```diff
@@ -0,0 +1,27 @@
...
+        public init(rawValue: Int8) {
+            self.rawValue = rawValue
+        }
+        public init(_ e: Edge) {
+            self.rawValue = e.rawValue
```
stackotter: Have you tested this with multiple edges? I have a feeling this should be `self.rawValue = 1 << e.rawValue` so that the individual edges can be distinguished when multiple raw values are combined.

Also, I prefer to use `edge` instead of `e` even when it's obvious what the single-character variable name means; just a style thing

stackotter (changes requested):
Looking pretty good so far 👍 I know it's not finished yet, just doing a partial review so I don't have to review as much later

Sources/SwiftCrossUI/Values/Edge.swift
Amzd: I noticed this too and fixed. 

I also usually use full word but I just copied the api from apple: https://developer.apple.com/documentation/swiftui/edge/set/init(_:)-91lwb

will change

stackotter: Ah makes sense 👍 yeah it's fine to change the internal names for parameters to make more sense as long as the label is the same

Sources/Gtk/Utility/GObjectRepresentable.swift
```diff
@@ -0,0 +1,25 @@
+//
+//  Copyright © 2016 Tomas Linhart. All rights reserved.
+//
```
stackotter: No need to add the copyright to our code

Sources/Gtk/Widgets/Widget.swift
```diff
@@ -237,41 +241,10 @@ open class Widget {
...
-        set {
-            gtk_widget_set_margin_end(widgetPointer, gint(newValue))
-        }
-    }
+    @GObjectProperty(named: "margin-top") public var topMargin: Int
+    @GObjectProperty(named: "margin-bottom") public var bottomMargin: Int
+    @GObjectProperty(named: "margin-start") public var leadingMargin: Int
+    @GObjectProperty(named: "margin-end") public var trailingMargin: Int
 
     public var horizontalAlignment: Align {
```
stackotter, line 242: Is it possible to update these properties too with GValueRepresentableEnum?

Sources/Gtk/Utility/GObjectProperty.swift
```diff
@@ -0,0 +1,28 @@
...
+
+    public init(named: String) {
+        self.name = named
+    }
+
+    public static subscript<EnclosingSelf: GObjectRepresentable>(
+        _enclosingInstance instance: EnclosingSelf,
+        wrapped _: ReferenceWritableKeyPath<EnclosingSelf, Value>,
+        storage storageKeyPath: ReferenceWritableKeyPath<EnclosingSelf, Self>
+    ) -> Value {
```
stackotter, line 20: Woah super cool! I didn't know you could access the enclosing instance like this

stackotter (changes requested):
Your GObjectProperty property wrapper looks super cool! It's going to make autogenerating class wrappers easier too, at the moment I need to generate a bunch of conversion code when generating property wrappers, which is pretty fiddly.

Sources/Gtk/Utility/GObjectRepresentable.swift
Amzd: Ah that got copied over I guess, will remove.

Sources/Gtk/Widgets/Widget.swift
Amzd: Yep, I've already used it in the constraints branch so I know it works. Didn't know if you would be okay with me just changing all the properties.

Sources/Gtk/Utility/GObjectProperty.swift
Amzd: Yep! Is how Published property wrapper work without mirror, we could use it for our Observed

Sources/Gtk/Utility/GObjectRepresentable.swift
```diff
@@ -0,0 +1,27 @@
...
+        let value = g_value_init(&v, V.type)!
+        newValue.apply(to: value)
+        g_object_set_property(gobjectPointer, named, value)
+    }
+
+    public func getProperty<V: GValueRepresentable>(named: String) -> V {
+        var v = GValue()
+        let value = g_value_init(&v, V.type)!
+        g_object_get_property(gobjectPointer, named, value)
+        // If this returned an optional at runtime and crashed it is because the
```
stackotter: I love this detailed comment. If you move it into a documentation comment for getProperty that would be amazing (more discoverable)

Sources/Gtk/Widgets/HeaderBar.swift
```diff
@@ -24,15 +24,7 @@ public class HeaderBar: Widget {
         }
     }
 
-    public var decorationLayout: String? {
-        get { return gtk_header_bar_get_decoration_layout(opaquePointer)?.toString() }
-        set { gtk_header_bar_set_decoration_layout(opaquePointer, newValue) }
-    }
-
-    public var spacing: Int {
```
stackotter, line 32: where did spacing go? was it invalid?

Sources/Gtk/Widgets/Scale.swift
```diff
@@ -15,7 +15,7 @@ public class Scale: Widget {
         }
     }
 
-    public var value: Double {
+    public var value: Double { // has no property
```
stackotter, line 18: huh weird

stackotter (changes requested):
Looks great! Just a couple of small comments/suggestions

Sources/Gtk/Widgets/HeaderBar.swift
Amzd: doesn't exist on HeaderBar :shrug: not sure how it got there https://docs.gtk.org/gtk4/class.HeaderBar.html#properties

Sources/Gtk/Widgets/Scale.swift
Amzd: I agree haha!

Sources/Gtk/Widgets/HeaderBar.swift
stackotter: lol, no clue 😅 maybe it did in gtk3?

Sources/Gtk/Utility/GObjectRepresentable.swift
Amzd: fixed

stackotter (approved):
Awesome!

PR #37, opened 2023-06-04
Generate all enum wrappers and implement enough to generate `Button` wrapper

stackotter:
This PR adds an automatic Gtk wrapper generation tool which parses the GIR formatted Gtk API specification and generates Swift wrappers for the underlying Gtk types.

Currently it is complete enough to generate all enum wrappers, but it isn't ready to generate many class wrappers. So far it can only generate the wrapper for `Button`. The plan is to have as little manual code as possible (manual code should just be one time exceptions, or helper functions for the generated code to use).


Sources/FileDialog/FileDialog.swift
```diff
@@ -3,35 +3,21 @@ import Foundation
 import Gtk
 
 /// A wrapper for Gtk's file dialog.
-public class FileDialog {
```
Amzd, line 6: Oh whoops I forgot about this one in my PR :+1:

Sources/Gtk/Generated/AccessibleInvalidState.swift
```diff
@@ -0,0 +1,50 @@
...
+/// accessible state.
+///
+/// Note that the %GTK_ACCESSIBLE_INVALID_FALSE and
+/// %GTK_ACCESSIBLE_INVALID_TRUE have the same values
+/// as %FALSE and %TRUE.
+public enum AccessibleInvalidState: GValueRepresentableEnum {
+    public typealias GtkEnum = GtkAccessibleInvalidState
+
+    /// There are no detected errors in the value
+    case false_
```
Amzd, line 13: You could also backtick escape keywords
```swift
enum Foo {
    case `true`
}

let bar: Foo = .true
```

Sources/Gtk/Utility/GValueRepresentableEnum.swift
```diff
@@ -14,7 +14,7 @@ extension GValueRepresentableEnum {
 
 extension GValueRepresentableEnum where GtkEnum.RawValue == Int32 {
     public init(_ pointer: UnsafeMutablePointer<GValue>) {
-        self.init(rawGtkValue: g_value_get_enum(pointer))
+        self.init(from: GtkEnum(rawValue: g_value_get_enum(pointer))!)
```
Amzd, line 17: :+1:

Sources/Gtk/Widgets/Orientable.swift
```diff
@@ -8,10 +8,10 @@ public protocol Orientable: AnyObject {
...
         get {
-            gtk_orientable_get_orientation(opaquePointer).toOrientation()
+            return Orientation(from: gtk_orientable_get_orientation(opaquePointer))
         }
         set {
-            gtk_orientable_set_orientation(opaquePointer, newValue.toGtkOrientation())
+            gtk_orientable_set_orientation(opaquePointer, newValue.toGtk())
         }
     }
 }
```
Amzd, line 17: I think you can do `extension Orientable where Self: GObjectRepresentable` here and then you could use the property wrapper

Amzd (dismissed):

Sources/Gtk/Generated/AccessibleInvalidState.swift
stackotter: Yeah, i wonder which is preferable. I chose to use a trailing `_` because I knew it would keep causing different issues in the code generator, but it would be doable. I still feel weird seeing `.true` even if it's valid

Sources/Gtk/Widgets/Orientable.swift
stackotter: Yeah true, missed this one (I only made those changes in the pass where I was fixing compile errors, and didn't remember it when doing a pass to replace manual enum properties with the property wrapper)

stackotter: Sad, it counts as a stored property :(

Amzd: ah yeah true :facepalm:

Sources/Gtk/Generated/AccessibleInvalidState.swift
Amzd: I don't mind .true anymore but I can see what you mean. Just making sure you were aware of the options, it was not a change request.

stackotter: 👍 Yeah thanks, I'll leave a todo in the code and decide at some point

Sources/SwiftCrossUI/Views/ForEach.swift
```diff
@@ -6,7 +6,8 @@ public struct ForEachViewChildren<
 
     class Storage {
         var nodes: [ViewGraphNode<Child>] = []
-        var container = GtkSectionBox(orientation: .vertical, spacing: 0).debugName(ForEach<Items, Child>.self)
```
Amzd, line 9: I assume this line was too long? Wouldn't it be nicer to newline before `.debugName` instead of before the param?

Sources/SwiftCrossUI/Views/Text.swift
```diff
@@ -14,14 +14,14 @@ public struct Text: View {
     }
 
     public func asWidget(_ children: EmptyViewContent.Children) -> GtkLabel {
-        let widget = GtkLabel()
+        let widget = GtkLabel(string: "")
```
Amzd: assuming it does smart comparison under the hood for checking draw size:
when you set it to `""` it will calculate its size for that then when you set it to `string` in the update func it has to calculate size again, and it costs us nothing to give it the string right away.
```suggestion
        let widget = GtkLabel(string: string)
```

Amzd (approved):

Sources/SwiftCrossUI/Views/ForEach.swift
stackotter: Oh yep, that wasn't me that was swift format, it makes some weird decisions sometimes 😅 Maybe I should try fixing that specific swift-format issue at some point

stackotter (commented):
*I have thoroughly inspected every single line changed*, looks good

PR #38, opened 2023-06-06
Orientation and spacing inside if/switch/for, ForEach and ViewContent

Amzd:
This PR fixes orientation and spacing not forwarding into if/switch/for, ForEach and ViewContent.

[Screencast from 2023-06-06 17-51-57.webm](https://github.com/stackotter/swift-cross-ui/assets/17001151/24788202-a051-4c5f-b190-a53e8f78db3f)

What I tested:

- HStack with variable spacing 
- variable window size
- Spacers in HStack and VStack
- nested if, switch and ForEach (deeper than 2 because every View.update is guaranteed to be called twice from inner to outer view)

The view I used for testing:

```swift
var body: some ViewContent {
    VStack {
        HStack {
            Text("Spacing:")
            Slider(state.$spacing, minimum: 0, maximum: 60)
        }

        HStack(spacing: state.spacing) {
            switch "test" {
            case "test":
                Text("Some text here")
                Button("Button") { }
                Spacer()
                if true {
                    Text("< spacer")
                    if true {
                        ForEach(["1", "2", "3"]) {
                            Text($0)
                            if true {
                                Text("-")
                            }
                        }
                        Text("4")
                    }
                }
            default:
                Text("not shown")
            }
            Text("bottom level")
        }

        HStack {
            Text("A")
            Spacer()
            Text("B")
        }

        VStack {
            Text("A")
            Spacer()
            Text("B")
        }
    }
    .padding(20)
}
```


Sources/Gtk/Widgets/Box.swift
```diff
@@ -35,5 +41,38 @@ open class Box: Widget, Orientable {
         widgets = []
     }
 
-    @GObjectProperty(named: "spacing") public var spacing: Int
+    @GObjectProperty(named: "spacing") open var spacing: Int
+
+    open var orientation: Orientation {
```
Amzd, line 46: This can be updated after #37 merged

Sources/Gtk/Widgets/Box.swift
```diff
@@ -35,5 +41,38 @@ open class Box: Widget, Orientable {
...
+        get { gtk_orientable_get_orientation(opaquePointer).toOrientation() }
+        set { gtk_orientable_set_orientation(opaquePointer, newValue.toGtkOrientation()) }
+    }
+}
+
+/// A Box that applies its parent Box' padding and orientation
+public class SectionBox: Box {
```
stackotter: This should probably be in a separate file

Sources/Gtk/Widgets/ModifierBox.swift
```diff
@@ -0,0 +1,58 @@
+import CGtk
+
+/// Helper Box that uses GtkBox under the hood but only allows one child.
+open class ModifierBox: Widget {
+    public private(set) var child: Widget?
+
+    public override init() {
+        super.init()
+        widgetPointer = gtk_box_new(GTK_ORIENTATION_VERTICAL, 0)
```
stackotter: Use `Orientation.vertical.toGtkOrientation()` here (it'll be best for generation)

Sources/SwiftCrossUI/Views/View.swift
```diff
@@ -18,7 +18,7 @@ public protocol View {
 
 extension View where Widget == GtkBox {
     public func asWidget(_ children: Content.Children) -> GtkBox {
-        let vStack = GtkBox(orientation: .vertical, spacing: 8)
+        let vStack = GtkBox(orientation: .vertical, spacing: 8).debugName(Self.self)
```
stackotter, line 21: This debugName thing is cool. Should it be added to the widgets of all views so that we don't need to keep adding it incrementally?

stackotter (changes requested):
Looking good! Just a few small requests

Sources/SwiftCrossUI/Views/View.swift
Amzd: Hmm wdym? It has to be called from the View to get its type

Sources/Gtk/Widgets/ModifierBox.swift
Amzd: I should probably change this after #37 merged then? Since it will change to `.toGtk()` right?

stackotter: I'll probably merge this first (I might not solve the signal stuff tonight). I'm rebasing the generation branch as I merge your PRs so merge conflicts aren't much of a concern (there aren't many and they're easy to fix with rebasing)

stackotter (approved):
Looks good!

PR #40, opened 2023-06-08
Fixed memory leaks in CSS implementation and made it type safe

Amzd:


Sources/Gtk/Utility/CSS/CSSProperty.swift
```diff
@@ -0,0 +1,47 @@
+/// A type safe CSS property
+///
+/// All CSS properties supported in Gtk: https://docs.gtk.org/gtk4/css-properties.html
+public struct CSSProperty: Equatable {
+    var stringRepresentation: String {
+        key + ":" + value + ";"
```
stackotter: Probably best to use string interpolation here for consistency

Sources/Gtk/Utility/CSS/CSSProperty.swift
```diff
@@ -0,0 +1,47 @@
...
+    var key: String
+    private var value: String
+
+    private static func rgba(_ color: Color) -> String {
+        "rgba(\(color.red*255),\(color.green*255),\(color.blue*255),\(color.alpha*255))"
+    }
+
+    public static func foregroundColor(_ color: Color) -> CSSProperty {
+        CSSProperty(key: "color", value: rgba(color))
+    }
```
stackotter, line 18: Just a formatting thing, put a blank line between each function

Sources/Gtk/Utility/CSS/CSSBlock.swift
```diff
@@ -0,0 +1,42 @@
...
+public struct CSSBlock: Equatable {
+    private var cssClass: String
+    /// property.key: property
+    private var keyedProperties: [String: CSSProperty] = [:]
+
+    var stringRepresentation: String {
+        ".\(cssClass){\(keyedProperties.values.map(\.stringRepresentation).joined())}"
+    }
+
+    init(forCssClass: String) {
```
stackotter: Replace with this to be a bit more swifty:
```swift
init(forCssClass cssClass: String) {
    self.cssClass = cssClass
}
```

Sources/Gtk/Utility/CSS/CSSBlock.swift
```diff
@@ -0,0 +1,42 @@
+/// Type safe representation of a block of css
+///
+/// ```
+/// .class {
+///   opacity: 0.3;
+/// }
+/// ```
+public struct CSSBlock: Equatable {
+    private var cssClass: String
+    /// property.key: property
```
stackotter:
```swift
/// The block's properties, keyed by property name.
```

Sources/Gtk/Utility/CSS/CSSProperty.swift
```diff
@@ -0,0 +1,47 @@
+/// A type safe CSS property
+///
+/// All CSS properties supported in Gtk: https://docs.gtk.org/gtk4/css-properties.html
```
stackotter: Something like this will show up nicer in the documentation site made with DocC:
```swift
/// For a list of all supported CSS properties, see [Gtk's CSS documentation](https://docs.gtk.org/gtk4/css-properties.html).
```

Sources/Gtk/Utility/CSS/CSSProperty.swift
```diff
@@ -0,0 +1,47 @@
...
+public struct CSSProperty: Equatable {
+    var stringRepresentation: String {
+        key + ":" + value + ";"
+    }
+
+    var key: String
+    private var value: String
```
stackotter: Value should probably have the same access level as key for consistency

Sources/Gtk/Widgets/Widget.swift
```diff
@@ -200,29 +200,26 @@ open class Widget: GObjectRepresentable {
...
 
-    public func setForegroundColor(color: Color) {
-        let className = String("class-\(UUID().uuidString)").replacingOccurrences(
-            of: "-",
-            with: "_"
-        )
-        className.withCString { string in
-            gtk_widget_add_css_class(widgetPointer, string)
+    /// A block of CSS for this widget
+    public lazy var css: CSSBlock = CSSBlock(forCssClass: customCssClass) {
```
stackotter, line 162: Does this need to be lazy? As far as I understand that would only be avoiding initialising a very simple structure (as opposed to the lazy on cssProvider which is necessary to avoid making and attaching unused providers)

Sources/SwiftCrossUI/Modifiers/Frame.swift
```diff
@@ -1,33 +1,16 @@
...
-    //         maximumWidth: Int?,
-    //         minimumHeight: minimumHeight,
-    //         maximumHeight: Int?
-    //     )
-    // }
 }
 
 /// A view used to manage a child view's size.
-private struct SimpleFrameModifierView<Child: View>: View {
+private struct CSSFrameModifierView<Child: View>: View {
```
stackotter: Could this possibly even be a generic `CSSModifierView` that can be used to implement all css-based modifiers?

stackotter (changes requested):
Looking like a massive improvement over the previous system! Just some requested changes and suggestions/questions

Sources/Gtk/Widgets/Widget.swift
Amzd: It also avoids initializing the customCssClass String which does some string manipulation, nothing slow but I thought if its not needed then why do it? I can change both to stored if you'd like

stackotter: Usually I only use `lazy` if I feel it really has a purpose (otherwise I'd be adding it everywhere for micro-optimisations that probably won't do much), so I'd so best to only have `lazy` for the provider

Amzd: actually -- the ObjectIdentifier init references self, which is only available in lazy so I will keep this lazy

Sources/SwiftCrossUI/Modifiers/CSSModifierView.swift
```diff
@@ -0,0 +1,31 @@
...
+    ///                    are additive and it does not remove css properties previously set.
+    internal init(_ child: Child, properties: [CSSProperty], clear: Bool = false, modifierName: String = #function) {
+        self.body = ViewContent1(child)
+        self.properties = properties
+        self.clear = clear
+        self.modifierName = modifierName
+    }
+
+    func asWidget(_ children: ViewGraphNodeChildren1<Child>) -> GtkModifierBox {
+        let widget = GtkModifierBox().debugName(Self.self)
```
stackotter: Should we pass `modifierName` to `debugName` (potentially as some optional 3rd context parameter) to make it clear which modifier was used to create the CSSModifierView?

stackotter (changes requested):
Looks awesome (I love the new CSSModifierView). Just one request related to `debugName`

Sources/SwiftCrossUI/Modifiers/CSSModifierView.swift
Amzd: Yeah I was looking into if we can't pass more variables to the inspector side but I can't find an alternative to `Widget.name`. We should add an "id" param to debugName function. I will not have time to update this PR until next week though.

Sources/Gtk/Utility/CSS/CSSProvider.swift
```diff
@@ -0,0 +1,29 @@
...
+        gtk_style_context_remove_provider_for_display(display, OpaquePointer(pointer))
+    }
+
+    /// Loads data into css_provider.
+    ///
+    /// This clears any previously loaded information.
+    ///
+    /// Deprecated since: 4.12
+    public func loadCss(from data: String) {
+        gtk_css_provider_load_from_data(pointer, data, data.count)
```
Amzd: Apply fix from #41

Amzd (commented):
Just reminding myself to fix this when I have time

Sources/SwiftCrossUI/Modifiers/CSSModifierView.swift
stackotter: Yeah all good 👍 No rush

stackotter (approved):
Looks good to me! CSSModifierView will definitely allow us to create new modifiers a lot more succinctly now

PR #41, opened 2023-06-09
Initial Windows support with vcpkg

stevapple:
This pull request adds initial Windows support by pulling in Gtk+ 4 dependency with vcpkg. It also adds a CI workflow for building on Windows, and updates all workflow dependencies.

Documentation is working in progress. We need extra guidelines on how to set environment variables correctly.


stackotter (commented):
Looking good so far! If you want me to do anything (such as attempt to follow the instructions to see if they are clear enough), just let me know. Btw thanks for consolidating the CGtk source directories for macos and windows+Linux, they used to have some differences under gtk3, but now they’re identical 😅

I’ll try getting a Windows machine up and running to try out the installation instructions tonight (unless you think the instructions need changes before they’re ready to be tested)

stackotter:
I now see why you added a dependency cache to the Windows action haha, it’s pretty slow installing all those dependencies

stevapple:
It worked on my machine, but the CI is failing and I filed https://github.com/apple/swift-package-manager/issues/6644.

If we cannot fix or work around the bug, we can simply build the `SwiftCrossUI` target.

stackotter:
Yep sounds like a suitable workaround until that can be fixed 👍

stackotter:
I’d be happy to merge the PR even if it just builds the library target for now (once I’ve tried it out)

stevapple:
I can’t reproduce it locally… On my machine plain `swift build` just works fine. And it should. We can basically assume everything should work.

stackotter (approved):
Looks good to me! 👍

stackotter:
I'll try it out for myself on Windows tonight, hopefully I can get it working 🤞 I'll update the documentation if I run into any issues, and I might even move the finer details into a DocC article at some point so that we can add more details than we otherwise might in a Readme

stevapple:
@stackotter Looking forward to your experience. I planned to post on Swift Forums next week to showcase the use of vcpkg with SwiftPM.

It's worth noting that vcpkg does work with macOS and Linux, and it has the benefit of not requiring any extra permission. So I guess there could be an article about how to use it. We may extend vcpkg-based Gtk version detection to other platforms too.

Package.swift
```diff
@@ -172,31 +167,40 @@ let package = Package(
...
 func getGtk4MinorVersion() -> Int? {
+#if os(Windows)
+    guard let pkgConfigPath = ProcessInfo.processInfo.environment["PKG_CONFIG_PATH"],
+          case let tripletRoot = URL(fileURLWithPath: pkgConfigPath, isDirectory: true)
+              .deletingLastPathComponent().deletingLastPathComponent(),
+          case let vcpkgInfoDirectory = tripletRoot.deletingLastPathComponent()
+              .appendingPathComponent("vcpkg").appendingPathComponent("info"),
+          let installedList = try? FileManager.default.contentsOfDirectory(at: vcpkgInfoDirectory, includingPropertiesForKeys: nil)
+            .map({ $0.deletingPathExtension().lastPathComponent }),
+          let packageName = installedList.first(where: { $0.hasPrefix("gtk_") && $0.hasSuffix("_\(tripletRoot.lastPathComponent)") })
```
Amzd, line 178: Windows implementation never checks if installed gtk major version is 4 or am I mistaken? 

The Mac/Linux version does gtk4-launch so doesn't need to confirm major version. But as far as I can tell Windows implementation doesn't check and could return minor version of Gtk3 if that is installed instead of 4?

README.md
```diff
@@ -8,7 +8,7 @@
     <img alt="GitHub" src="https://img.shields.io/github/license/stackotter/swift-cross-ui">
 </p>
 
-A SwiftUI-like framework for creating cross-platform apps in Swift. It uses [SwiftGtk](https://github.com/stackotter/SwiftGtk) as its backend (GTK 4.10+).
+A SwiftUI-like framework for creating cross-platform apps in Swift. It uses Gtk 4.10+ as its backend.
```
Amzd, line 11: Good catch! But we support any Gtk4 not just 4.10+

README.md
```diff
@@ -74,21 +74,86 @@ Here's the [documentation site](https://stackotter.github.io/swift-cross-ui/docu
...
 ```sh
 brew install gtk4
 ```
 
-### Linux: Installing Gtk 4.10+ and clang
+### Linux: Installing Gtk 4.10+ and Clang
 
-Install Gtk+3 and Clang using apt or the package manager of your choice. Ensure that the installed
+Install Gtk+ and Clang using apt or the package manager of your choice. Ensure that the installed
 version of Gtk is Gtk 4.10+
```
Amzd, line 86: Just noticed this one is also wrong. We should maybe also add that gtk is probably already installed for Linux users with GNOME environment? @stackotter

Sources/Gtk/CssProvider.swift
```diff
@@ -9,7 +9,7 @@ public class CssProvider {
 
     public func loadFromData(_ data: String) {
         data.withCString { cString in
-            gtk_css_provider_load_from_data(pointer, cString, data.count)
+            gtk_css_provider_load_from_data(pointer, cString, gssize(data.count))
```
Amzd, line 12: I need to check if this conflicts with #40

Amzd (commented):
Great news to get Windows working 🥳

Package.swift
stevapple: In vcpkg "gtk" always refers to Gtk+ 4. We currently rely on vcpkg installation structure for version detection.

README.md
stackotter: Yeah sure 👍

README.md
stackotter: Fixed 👍

PR #55, opened 2023-10-19
Add slider for QtBackend

sirlan-ff00ff:
- Add Qt back-end implementation of slider

The `QSlider` widget doesn't seem to have an equivalent of 'decimal places'


stackotter (approved):
Looks good!

PR #56, opened 2023-10-19
Update back-ends to not run actions on main thread on non Apple platfo…

sirlan-ff00ff:
…rms; Fix typo in QtBackend

- Changes apply to current Qt, LVGL, and Curses back-ends
- Fix missing letter typo in Qt back-end internal state


stackotter (approved):
Thanks for fixing this bug! 🎉

PR #59, opened 2023-10-22
ViewGraph rewrite

stackotter:
Still a work in progress, but essentially I'm rewriting the view graph and all related code to get rid of the `ViewContent` protocol, and make the system of types and protocols easier to reason about. Unfortunately this requires using more existential generics (e.g. `any ViewGraphNodeChildren`), but I believe an ideal solution with no existential generics isn't possible in current Swift within the restrictions of the SwiftUI API. Being fully typesafe would require leaking a lot of internal implementation details into user UI code, and these changes show exactly why that's important; anyone already using SwiftCrossUI only has to find and replace `ViewContent` with `View` to update to the latest version (which wouldn't be the case with changes as aggressive as this if internal details were leaked).

The rationale behind this change is that without using proper variadic generics (which aren't backwards compatible or stable enough yet), the current implementation will explode in complexity exponentially when we try to introduce the `SceneGraph` between `App` and the `ViewGraph` (to allow multi-windowing in a similar way to SwiftUI) — trust me, I've already tried, and that's what led me to do this first 😅 Additionally, this matches `SwiftUI` a bit better but that's just a bonus side effect.

## Overview of changes

- Removed `ViewContent` (`View` now fills its role with the `VariadicViewN` series of types).
- Introduced more type-erasure to ensure that the backend and internal implementation details (such as `ViewGraphNodeChildren`) don't leak into the `View` protocol. Used sparingly as always, but found it a necessary compromise.
- Introduced subprotocols of `View` to make implementing type-safe views simpler.
- Changed `ViewGraphNodeChildren` implementations to classes where necessary to avoid nested `Storage` classes which were just an unnecessary complication.
- Documented everything including internal implementations (to aid both users and contributors)
- Removed `LinkedList` (I hate linked lists and it wasn't necessary)
- Added a `StressTest` example for performance testing (will be expanded on in the future)


Amzd:
> anyone already using SwiftCrossUI only has to find and replace `ViewContent` with `View`

Not sure if this is common but if so maybe a typealias with an `@available` tag to instruct switching to View might be nice for developer experience

stackotter:
> > anyone already using SwiftCrossUI only has to find and replace `ViewContent` with `View`
> 
> Not sure if this is common but if so maybe a typealias with an `@available` tag to instruct switching to View might be nice for developer experience

Love that idea, implemented it now 👌

stackotter (commented):
Looks good to me! (surprise)

PR #61, opened 2023-10-23
Implement `Scene` and an associated internal scene graph system

stackotter:
This change was made to be able to introduce multi-windowing and get also get rid of the windowProperties property currently required to configure the app's window. I've basically followed SwiftUI's API, but probably not their implementation at all. This was so much easier to figure out after having grappled with the ViewGraph for so long yesterday.

To update an existing app, the changes required are pretty minimal. Simply wrap your app's body in a `WindowGroup` and change it from `some View` to `some Scene` (just like you would in SwiftUI).

```swift
struct MyApp: App {
    // ...
    
    // Before
    var body: some View {
        Text("Hello, world!")
    }

    // After
    var body: some Scene {
        WindowGroup {
            Text("Hello, world!")
        }
    }
}
```

This more closely aligns our `App` protocol with that of SwiftUI easing the process of transitioning from SwiftUI to SwiftCrossUI.


stackotter (commented):
Lgtm (surprise)

PR #67, opened 2023-11-02
Added ToggleButton View (via GtkBackend)

NinjaCheetah:
Added support for GTK's `ToggleButton` to `AppBackend` and `GtkBackend` so it can be used from swift-cross-ui. I also created `ButtonTypesExample` to demonstrate the types of buttons, which are currently just the default `Button` and now `ToggleButton`. Everything appears to be working as intended as is documented to the best of my ability, but let me know if any changes are required.

### DIFF
--- a/Sources/SwiftCrossUI/Backend/AppBackend.swift
+++ b/Sources/SwiftCrossUI/Backend/AppBackend.swift
@@ -95,6 +95,17 @@ public protocol AppBackend {
     /// Sets a button's action (triggered on click). Replaces any existing actions.
     func setAction(ofButton button: Widget, to action: @escaping () -> Void)
 
+    /// Created a labelled toggle button that toggles a Bool on click. Predominantly
+    /// used by ``ToggleButton``.
+    func createToggleButton(
+        label: String, toggled: Bool, onChange: @escaping (Bool) -> Void
+    ) -> Widget
+    /// Update the state of the toggleButton as soon as it is toggled for live access
+    /// to the state of the button.
+    func setOnToggled(ofToggleButton toggleButton: Widget, to onChange: Binding<Bool>)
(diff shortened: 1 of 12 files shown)


Sources/SwiftCrossUI/Backend/AppBackend.swift
```diff
@@ -95,6 +95,17 @@ public protocol AppBackend {
     /// Sets a button's action (triggered on click). Replaces any existing actions.
     func setAction(ofButton button: Widget, to action: @escaping () -> Void)
 
+    /// Created a labelled toggle button that toggles a Bool on click. Predominantly
```
stackotter: `Creates a labelled toggle button that is either on or off` would probably a bit more clear (the `Bool` part is already implied by the function signature)

Sources/SwiftCrossUI/Views/ToggleButton.swift
```diff
@@ -0,0 +1,36 @@
+/// A control that toggles an option.
```
stackotter: `A control that is either on or off.` is a bit more clear in my opinion, as `an option` is a bit vague

Sources/SwiftCrossUI/Views/ToggleButton.swift
```diff
@@ -0,0 +1,36 @@
...
+public struct ToggleButton: ElementaryView, View {
```
stackotter: SwiftUI has a view called `Toggle`, the `.button` variant of which is equivalent to this. I'm open to discussion on this, but it might make sense to rename this to `Toggle` for now to match SwiftUI, and then in the future if we ever decide that we're going to leave SwiftUI semi-compatibility behind, we can rename everything to make more sense.

PR #68, opened 2023-11-04
Simplify the AppBackend API

stackotter:
There are currently a lot of atomic methods for updating each parameter of each view. In reality, SwiftCrossUI always uses these atomic methods one after another, meaning that they can just be combined into a single `update` method for each widget type. Although the methods will be bigger, I believe that it does make creating and maintaining a backend simpler. For example, adding a new configurable parameter to a widget used to mean adding a whole new method to the backend api, possibly modifying the creation method, adding a default implementation, and then adding a method to every compatible backend, it now means adding a single parameter to each place that the update method is referenced, which is a lot easier.


stackotter (commented):
Lgtm! (no way 🙀)

PR #69, opened 2023-11-04
Added toggleStyle modifier with .button and .switch

NinjaCheetah:
Added support for `Gtk.Switch` to `AppBackend`, `GtkBackend`, and gave it a view. I also added it to `GtkCodeGen` since it was not previously there. `ControlsExample` has also been updated to showcase the switch.


stackotter:
I find it very amusing that it fails to build on Linux and macOS, but succeeds on Windows 😅 That must be very uncommon among Swift packages

Examples/Controls/Controls.swift
```diff
@@ -14,21 +21,62 @@ struct ControlsApp: App {
...
     var body: some View {
-        VStack(spacing: 5) {
+        NavigationSplitView {
+            VStack() {
+                Button("Buttons") { state.selectedCategory = .buttons }
+                Button("Toggles") { state.selectedCategory = .toggles }
+                Spacer()
+            }.padding(10)
+        } detail: {
+            VStack() {
```
stackotter: You don't need the `()` here as the trailing closure implies them

Examples/Controls/Controls.swift
```diff
@@ -14,21 +21,62 @@ struct ControlsApp: App {
...
+                        TogglesView
+                            .padding(.bottom, 10)
+                    case nil:
+                        Text("Select a category.")
+                }
+            }.padding(10)
+        }
+    }
+
+    var ButtonsView: some View {
```
stackotter: Name this `buttonsView` instead to follow the Swift conventions for property names

Examples/Controls/Controls.swift
```diff
@@ -14,21 +21,62 @@ struct ControlsApp: App {
...
-    let windowProperties = WindowProperties(title: "ControlsApp", resizable: true)
+    let windowProperties = WindowProperties(
+        title: "ControlsApp",
+        defaultSize: WindowProperties.Size(400, 350),
+        resizable: true
+    )
 
     var body: some View {
-        VStack(spacing: 5) {
+        NavigationSplitView {
```
stackotter: It would probably be best not to use a `NavigationSplitView` and just have all of the controls visible at once. I think it's important to have the examples stick to the specific things they're trying to showcase so that for example if someone's implementing/testing a new backend, it doesn't have to have too much implemented to try out individual examples.

Sources/GtkBackend/GtkBackend.swift
```diff
@@ -188,6 +188,22 @@ public struct GtkBackend: AppBackend {
         }
     }
 
+    public func createSwitch(active: Bool, onChange: @escaping (Bool) -> Void) -> Widget {
+        let switchWidget = Switch()
+        switchWidget.active = active
```
stackotter, line 193: Set the `onChange` here too (it's not super necessary since SwiftCrossUI will call `setIsActive` and `setOnChange` before rendering it anyway, but it's best to have the method do what it says it will for now)

Sources/SwiftCrossUI/Views/Toggle.swift
```diff
@@ -1,5 +1,82 @@
...
+        switch selectedToggleStyle {
+            case .switch:
+                HStack() {
+                    Text(label)
+                    ToggleSwitch(active: active)
+                }
+            case .button:
+                ToggleButton(label, active: active)
+    }
+  }
```
stackotter: The indentation of these two closing braces is a bit off

Sources/SwiftCrossUI/Views/Toggle.swift
```diff
@@ -1,5 +1,82 @@
-/// A control that is either on or off.
-public struct Toggle: ElementaryView, View {
+
```
stackotter: Delete the trailing newline

Sources/SwiftCrossUI/Views/Toggle.swift
```diff
@@ -1,5 +1,82 @@
...
+public struct Toggle: View {
+    /// The style of toggle shown.
+    public let selectedToggleStyle: ToggleStyle
```
stackotter: If you set this as a non-public `var` then the modifier can be implemented a bit nicer (doesn't need to be public because people shouldn't be getting properties of views really)

Sources/SwiftCrossUI/Views/Toggle.swift
```diff
@@ -1,5 +1,82 @@
...
+    /// The label to be shown on or beside the toggle.
+    var label: String
+    /// Whether the toggle is active or not.
+    var active: Binding<Bool>
+
+    /// Creates a toggle that displays a custom label.
+    public init(selectedToggleStyle: ToggleStyle = .button, _ label: String, active: Binding<Bool>) {
```
stackotter: To match SwiftUI I think this first parameter should be removed (so that people don't have two different ways of setting the toggle style). The initializer should just set the toggle style to `button` for now.

Sources/SwiftCrossUI/Views/Toggle.swift
```diff
@@ -1,5 +1,82 @@
...
+}
+
+extension Toggle {
+    /// Sets the style of the toggle.
+    public func toggleStyle(_ selectedToggleStyle: ToggleStyle) -> Toggle {
+        return Toggle(
+            selectedToggleStyle: selectedToggleStyle, 
+            self.label, 
+            active: self.active
+        )
```
stackotter: If `selectedToggleStyle` is changed to a `var` then you can do the following instead. The main benefit being that if more properties are added to `Toggle` you don't need to update this modifier's implementation (that's why I generally follow this pattern for modifiers):

```swift
var toggle = self
toggle.selectedToggleStyle = selectedToggleStyle
return toggle
```

Sources/SwiftCrossUI/Views/Toggle.swift
```diff
@@ -1,5 +1,82 @@
...
+    }
+}
+
+public enum ToggleStyle {
+    case `switch`
+    case button
+}
+
+/// A light switch style control that is either on or off.
+private struct ToggleSwitch: ElementaryView, View {
```
stackotter: I'm surprised it lets you make this `private`, I don't really know what that means for declarations like this. If you instead omit any access level modifiers it defaults to `internal` (so that users of the package can't access it, but it can still be used within the package as an implementation detail).

Sources/SwiftCrossUI/Views/Toggle.swift
```diff
@@ -1,5 +1,82 @@
...
+    public init(selectedToggleStyle: ToggleStyle = .button, _ label: String, active: Binding<Bool>) {
+        self.selectedToggleStyle = selectedToggleStyle
+        self.label = label
+        self.active = active
+    }
+
+    public var body: some View {
+        switch selectedToggleStyle {
+            case .switch:
+                HStack() {
```
stackotter: Remove the `()` here

Sources/SwiftCrossUI/Views/Toggle.swift
```diff
@@ -1,5 +1,82 @@
...
+
+public enum ToggleStyle {
+    case `switch`
+    case button
+}
+
+/// A light switch style control that is either on or off.
+private struct ToggleSwitch: ElementaryView, View {
+    /// Whether the switch is active or not.
+    private var active: Binding<Bool>
```
stackotter: Similarly, since it's an internal implementation detail you can get rid of all the access level modifiers (`private`, `public`, etc) on its properties, inits, and methods.

Sources/SwiftCrossUI/Views/Toggle.swift
```diff
@@ -1,5 +1,82 @@
...
+    ) {
+        backend.setIsActive(ofSwitch: widget, to: active.wrappedValue)
+        backend.setOnChange(ofSwitch: widget) { newActiveState in
+            active.wrappedValue = newActiveState
+        }
+    }
+}
+
+/// A button style control that is either on or off.
+private struct ToggleButton: ElementaryView, View {
```
stackotter: `ToggleButton` and `ToggleSwitch` should be moved to separate files just to keep things a bit easier to find.

stackotter (changes requested):
Looking great so far! I've got some requested changes but for the most part they're pretty small things. Loving the documentation comments, they mostly just sound like I wrote them now 😅

stackotter (approved):
Looks great, I'll just test out the example quickly before merging 👍

PR #70, opened 2023-11-10
Add new examples to documentation and categorize pages

NinjaCheetah:
Added the following examples to `Examples.md` and wrote a short description for each:
- SpitExample
- StressTestExample
- SpreadsheetExample
- ControlsExample

Categorized documentation pages. Currently using the categories `App Structure` (Apple's name for that category), `Views`, `States`, and `Implementation Details`. Descriptions for most of those categories are missing at the moment.

Some small changes to the overview of SwiftCrossUI and the README were also made, namely changing the formatting of the package import because its changed slightly since they were written, and touching up some wording here and there.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -2,42 +2,6 @@
   "object": {
     "pins": [
       {
-        "package": "AsyncExtensions",
-        "repositoryURL": "https://github.com/lhoward/AsyncExtensions",
-        "state": {
-          "branch": "linux",
-          "revision": "2218eaa30dbdb39b063e92644fc28ed22e2cb942",
-          "version": null
-        }
-      },
-      {
-        "package": "LVGL",
-        "repositoryURL": "https://github.com/PADL/LVGLSwift",
-        "state": {
-          "branch": null,
-          "revision": "19c19a942153b50d61486faf1d0d45daf79e7be5",
-          "version": null
-        }
-      },
-      {
-        "package": "Qlift",
-        "repositoryURL": "https://github.com/Longhanks/qlift",
-        "state": {
-          "branch": null,
-          "revision": "ddab1f1ecc113ad4f8e05d2999c2734cdf706210",
-          "version": null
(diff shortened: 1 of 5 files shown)


Sources/SwiftCrossUI/SwiftCrossUI.docc/Basic Usage.md
```diff
@@ -27,14 +27,18 @@ If you're on a Debian or Ubuntu-based Linux distro, you can use the following co
 ```
 sudo apt install libgtk-4-dev clang
 ```
+For Fedora:
```
stackotter, line 31: Add a newline before this

stackotter (changes requested):
Just a single missing newline, other than that it looks great 👍

Package.resolved
Amzd, line 1: the Package.swift didn't change, was the Package.resolved change intentional?

stackotter: I reckon that's because they don't have lvgl or qt installed, but yeah I guess the package.resolved should probably be reset to the one from the previous commit so that it's still usefully locking deps. Once the backends are split out into separate packages this shouldn't be an issue anymore

NinjaCheetah: Should the Package.resolved be under version control at all? Most people probably don't have all three backends anyway meaning that the file is going to be regenerated when they go to use it. It just doesn't seem like having it has any benefits here, but I could be wrong.

Amzd: I guess Package.swift could enforce exact versions to make sure maintainers are using the same versions but usually if you have Package.resolved changes it is a good warning a dependency was updated. It's a trade off with having to be careful when committing.

stackotter (approved):
lgtm, I'll be addressing the Package.resolved properly soon, so it doesn't really bother me that the Package.resolved was updated by this PR (I've been pretty relaxed about the Package.resolved in the past too)

PR #76, opened 2024-03-19
Make all products/targets visible to consumers.

furby-tm:
**The Problem**

> error: target `GtkBackend` does not exist in package `swift-cross-ui`.

The way to prevent this error from happening is for _consuming users_ to implement something like the following to reference **swift-cross-ui** in their own **Package.swift** manifests in a way that doesn't break a user's existing **SwiftPM** package:

```swift
public static func pkgDeps() -> [Package.Dependency]
{
  #if os(Linux) || os(Android) || os(OpenBSD) || os(FreeBSD) || os(Windows) || os(Cygwin)
    [
      .package(url: "https://github.com/stackotter/swift-cross-ui", revision: "dd09e61")
    ]
  #else
    []
  #endif
}

public static func backend() -> [Target.Dependency]
{
  #if os(Linux) || os(Android) || os(OpenBSD) || os(FreeBSD) || os(Windows) || os(Cygwin)
    [
      .product(name: "SwiftCrossUI", package: "swift-cross-ui", condition: .when(platforms: [.linux, .windows])),
      .product(name: "GtkBackend", package: "swift-cross-ui", condition: .when(platforms: [.linux, .windows])),
    ]
  #else
    []
  #endif
}
```

**The Solution**

* This revision attempts to make it easier for consuming packages which rely on **swift-cross-ui** as a package dependency to conditionally apply targets based on the OS platform, without having to preprocess the various backends in their own Package.swift manifest files, the preprocessing workaround is exampled above.

**Use Case**

Now in my own **SwiftPM** project I can do the following without inducing the error noted above:

```swift
  // maintaining the "standard" method of adding SwiftPM dependencies
  // to an existing SwiftPM project is what is demonstrated here.

  ...
  dependencies: [
    .package(url: "https://github.com/furby-tm/swift-cross-ui", revision: "a26353e")
  ],
  targets: [
    .executableTarget(
      name: "MyReallyEpicCrossPlatformApp",
      dependencies: [
        .product(name: "SwiftCrossUI", package: "swift-cross-ui", condition: .when(platforms: [.linux, .windows])),
        .product(name: "GtkBackend", package: "swift-cross-ui", condition: .when(platforms: [.linux, .windows])),
      ]
    ),
  ]
```

**Potential Drawback**

You will now have to preprocess all backend-dependent Swift code in **swift-cross-ui** if you wish to be able to build the _whole package_ (`swift build`) (now that specifying a _specific target_ is a prerequisite (`swift build --target GtkBackend`)) like the following:
```swift
// for any file that uses anything from 'Gtk'.
#if canImport(Gtk)
  import Gtk

  // many lines of swift code.
  // many lines of swift code.
  // many lines of swift code.
  // many lines of swift code.
#endif /* canImport(Gtk) */
```

However, it appears that supporting `swift build` to "target" the whole package was not something that was working anyway from the latest **SPI** [build results](https://swiftpackageindex.com/builds/00C4697B-3832-4803-BC4C-BDB953B7327A), so this drawback seems unsubstantial.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -29,21 +29,12 @@
         }
       },
       {
-        "package": "Qlift",
-        "repositoryURL": "https://github.com/Longhanks/qlift",
-        "state": {
-          "branch": null,
-          "revision": "ddab1f1ecc113ad4f8e05d2999c2734cdf706210",
-          "version": null
-        }
-      },
-      {
         "package": "swift-collections",
         "repositoryURL": "https://github.com/apple/swift-collections.git",
         "state": {
           "branch": null,
-          "revision": "937e904258d22af6e447a0b72c0bc67583ef64a2",
-          "version": "1.0.4"
+          "revision": "94cf62b3ba8d4bed62680a282d4c25f9c63c2efb",
+          "version": "1.1.0"
         }
       },
       {
--- a/Package.swift
+++ b/Package.swift
@@ -37,7 +37,6 @@ var dependencies: [Package.Dependency] = [
     fatalError("Unsupported platform.")
 #endif
 
-var conditionalProducts: [Product] = []
 var conditionalTargets: [Target] = []
 var exampleDependencies: [Target.Dependency] = [
     "SwiftCrossUI", "GtkBackend",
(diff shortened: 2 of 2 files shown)


stackotter:
I've just pushed some changes that I've had locally for a little while and I believe that they fix the same issue (although possibly slightly differently). I ended up implementing a SwiftPM plugin that checks the system dependencies so that instead of the disgusting `couldn't build Objective C module CGtk` error or whatever SwiftPM spits out, you get something saying that you need to install gtk to use the `GtkBackend` (and similarly for other backends). Feel free to open an issue if the changes don't fix the issue that you were trying to fix!

stackotter:
Hmm, annoyingly the plugin seems to run for QtBackend even if you're only using the GtkBackend and vice versa 😭 I'll remove it for now even though the errors you get are terrible:

```
warning: couldn't find pc file for Qt5Widgets
Building for debugging...
/Users/stackotter/Desktop/Projects/Swift/SwiftCrossUI/.build/checkouts/qlift/Sources/CQlift/qlift-dispatchToQThread.cpp:8:10: fatal error: 'QTimer' file not found
#include <QTimer>
         ^~~~~~~~
1 error generated.
```

PR #91, opened 2024-05-19
Fix build errors when running examples using WinUIBackend

ducaale:
Example apps that didn't specify the `Backend` typealias were failing under `WinUIBackend` because it didn't have a default value.

See https://github.com/stackotter/swift-cross-ui/commit/26627bdd148ea53e4dab70af823fff577836c409 which removed need for explicit backend typealias in all backends except for `WinUIBackend`


stackotter (approved):
Awesome, thanks for fixing these oversights! 😅 Will merge one the workflows finish running

PR #96, opened 2025-01-02
Support WinUIBackend on both arm64 and x86_64.

furby-tm:
* this is so that the **WindowsAppSDK** bootstrap works on both arm64 and x86_64 windows, as it provides the **`Microsoft.WindowsAppRuntime.Bootstrap.dll`** for each architecture, and is conditionally compiled based on the architecture of the host machine.

### DIFF
--- a/Examples/Package.resolved
+++ b/Examples/Package.resolved
@@ -143,6 +143,15 @@
       }
     },
     {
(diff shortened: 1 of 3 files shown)


stackotter:
I've merged these into `layout_system` locally with you credited in the `Co-authored-by` field.

PR #97, opened 2025-01-04
Add a default first menu item implementation.

furby-tm:
* Created a default first menu item implementation, which is better than a dummy because this presents a standard "**About MyApp**" and ability to quit the app, similar to the default first menu item that `SwiftUI` creates by default on **macOS**.

It looks like the following:
Screenshot 2025-01-04 at 2 31 44 PM


stackotter (approved):
Lgtm, just trying out locally before merging

stackotter:
Thanks for PRing this! Great quality of life improvement

PR #98, opened 2025-01-11
UIKit backend

bbrk24:
Incomplete features:
- Menus (as discussed in Discord, `UIMenu` is incompatible with the current API)
- `Picker` (only supported on iOS) and `Slider` (only supported on iOS & Mac Catalyst)
- SplitView (how to handle this on mobile is an open question)
- Table (need column widths in order to render well)
- Open/Save file dialogs (probably not even possible on all UIKit platforms)


stackotter (approved):
Looks good to me, super excited! 🎉 Thanks for implementing everything so neatly too 🙏

PR #99, opened 2025-01-14
Implement Picker for tvOS & Mac Catalyst

bbrk24:
This implements two variants of `Picker`, one based on `UIPickerView` and one based on `UITableView`. This is the logic I use for deciding which version to display:

https://github.com/stackotter/swift-cross-ui/blob/98717e895aa791b7bf88808534ed2167cb582f9b/Sources/UIKitBackend/UIKitBackend%2BPicker.swift#L135-L145

Basically, I use `UIPickerView` wherever it works, and `UITableView` everywhere else.


stackotter (approved):
Looks good 👌

PR #100, opened 2025-01-15
More native UIKit features

bbrk24:
Including custom application delegates and `UIViewRepresentable`. The context type is different from SwiftUI but the `UIViewRepresentable` protocol is otherwise the same.

I'm really unsure about the size calculation code so I would appreciate guidance on that.

### DIFF
--- a/Sources/UIKitBackend/UIViewRepresentable.swift
+++ b/Sources/UIKitBackend/UIViewRepresentable.swift
@@ -1,0 +1,195 @@
+import SwiftCrossUI
+import UIKit
+
+public struct UIViewRepresentableContext<Coordinator> {
+    public let coordinator: Coordinator
+    public internal(set) var environment: EnvironmentValues
+}
+
+public protocol UIViewRepresentable: View
+where Content == Never {
+    associatedtype UIViewType: UIView
+    associatedtype Coordinator = Void
+
+    /// Create the initial UIView instance.
+    func makeUIView(context: UIViewRepresentableContext<Coordinator>) -> UIViewType
+
+    /// Update the view with new values.
+    /// - Parameters:
+    ///   - uiView: The view to update.
+    ///   - context: The context, including the coordinator and potentially new environment
+    ///   values.
+    /// - Note: This may be called even when `context` has not changed.
+    func updateUIView(_ uiView: UIViewType, context: UIViewRepresentableContext<Coordinator>)
+
+    /// Make the coordinator for this view.
+    ///
+    /// The coordinator is used when the view needs to communicate changes to the rest of
+    /// the view hierarchy (i.e. through bindings), and is often the view's delegate.
+    func makeCoordinator() -> Coordinator
+
+    /// Compute the view's preferred size.
+    /// - Parameters:
+    ///   - proposal: The proposed frame for the view to render in.
+    ///   - uiVIew: The view being queried for its preferred size.
+    ///   - context: The context, including the coordinator and environment values.
+    /// - Returns: The preferred size for the view, ideally one that fits within `proposal`,
+    /// or `nil` to use a default size for this view.
+    func sizeThatFits(
+        _ proposal: CGSize, uiView: UIViewType, context: UIViewRepresentableContext<Coordinator>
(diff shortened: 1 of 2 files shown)


Sources/UIKitBackend/UIViewRepresentable.swift
```diff
@@ -0,0 +1,195 @@
...
+            ),
+            idealSize: SIMD2(
+                roundedSize.x < 0 ? proposedSize.x : roundedSize.x,
+                roundedSize.y < 0 ? proposedSize.y : roundedSize.y
+            ),
+            minimumWidth: roundedSize.x > proposedSize.x ? roundedSize.x : 0,
+            minimumHeight: roundedSize.y > proposedSize.y ? roundedSize.y : 0,
+            maximumWidth: nil,
+            maximumHeight: nil
+        )
```
stackotter, line 138: What do you reckon about the idea of updating the `UIViewRepresentable.sizeThatFits` method to return a `ViewSize` directly? (this would probably require naming the method something else too)

The issue with trying to discern the minimum and maximum sizing bounds from just the view's current size is that the `sizeThatFits` implementation is completely opaque to SwiftCrossUI meaning that any sizing implementation will need to make some assumptions, and things can go pretty wrong in SwiftCrossUI's layout system if the minimum and maximum sizes are incorrect. We need to at least give users some way to override the minimum and maximum size calculations in case we're incorrect, and that's why I believe that my suggested approach could be a worthwhile deviation from the corresponding SwiftUI API.

Sources/UIKitBackend/UIViewRepresentable.swift
```diff
@@ -0,0 +1,195 @@
...
+            view.topAnchor.constraint(equalTo: self.topAnchor),
+            view.leadingAnchor.constraint(equalTo: self.leadingAnchor),
+            view.trailingAnchor.constraint(equalTo: self.trailingAnchor),
+            view.bottomAnchor.constraint(equalTo: self.bottomAnchor),
+        ])
+
+        return view
+    }()
+
+    func updateContext(environment: EnvironmentValues) {
```
stackotter, line 176: Renaming this to `update(with environment: EnvironmentValues)` would make a bit more sense to me. Especially if you implement the comment I've left a few lines down.

Sources/UIKitBackend/UIViewRepresentable.swift
```diff
@@ -0,0 +1,195 @@
...
+        if context == nil {
+            context = .init(coordinator: representable.makeCoordinator(), environment: environment)
```
stackotter, line 178: Let me know if there's a reason this method has to be implemented the current way, but I reckon this line should be moved to the `RepresentingWidget` initialiser so that it runs when the parent UIViewRepresentable's `asWidget` gets called. Then `context` doesn't have to be optional, and this update method can be reduced to just the else case.

stackotter (changes requested):
Looking good so far, let me know what you think about the `UIViewRepresentable` suggestions I made (other than the comments I left, everything else looks great).

Sources/UIKitBackend/UIViewRepresentable.swift
bbrk24: That sounds reasonable to me. I'll update this tonight

Sources/UIKitBackend/UIViewRepresentable.swift
bbrk24: `asWidget` doesn't take the environment values as a parameter, so there's no way to get the environment in the initializer, is there? Otherwise I totally would have done it that way (and I actually tried to at first before realizing why I can't).

stackotter: Ah ok yeah, makes sense. This way works well enough then.

Renaming the function probably still makes sense cause it doesn't just update the context it also updates/creates the underlying UIView using the new updated context at the same time.

stackotter:
If you retarget this to the `main` branch you should be able to reopen it (GitHub closed it automatically on my behalf). The `main` branch is now identical to what the `layout_system` branch was.

PR #103, opened 2025-01-16
[UIKitBackend] Custom Application Delegates & `UIViewRepresentable`

bbrk24:
Continuing #100

### DIFF
--- a/Sources/UIKitBackend/UIViewRepresentable.swift
+++ b/Sources/UIKitBackend/UIViewRepresentable.swift
@@ -1,0 +1,195 @@
+import SwiftCrossUI
+import UIKit
+
+public struct UIViewRepresentableContext<Coordinator> {
+    public let coordinator: Coordinator
+    public internal(set) var environment: EnvironmentValues
+}
+
+public protocol UIViewRepresentable: View
+where Content == Never {
+    associatedtype UIViewType: UIView
+    associatedtype Coordinator = Void
+
+    /// Create the initial UIView instance.
+    func makeUIView(context: UIViewRepresentableContext<Coordinator>) -> UIViewType
+
+    /// Update the view with new values.
+    /// - Parameters:
+    ///   - uiView: The view to update.
+    ///   - context: The context, including the coordinator and potentially new environment
+    ///   values.
+    /// - Note: This may be called even when `context` has not changed.
+    func updateUIView(_ uiView: UIViewType, context: UIViewRepresentableContext<Coordinator>)
+
+    /// Make the coordinator for this view.
+    ///
+    /// The coordinator is used when the view needs to communicate changes to the rest of
+    /// the view hierarchy (i.e. through bindings), and is often the view's delegate.
+    func makeCoordinator() -> Coordinator
+
+    /// Compute the view's preferred size.
+    /// - Parameters:
+    ///   - proposal: The proposed frame for the view to render in.
+    ///   - uiVIew: The view being queried for its preferred size.
+    ///   - context: The context, including the coordinator and environment values.
+    /// - Returns: The preferred size for the view, ideally one that fits within `proposal`,
+    /// or `nil` to use a default size for this view.
+    func sizeThatFits(
+        _ proposal: CGSize, uiView: UIViewType, context: UIViewRepresentableContext<Coordinator>
+    ) -> CGSize?
+
+    /// Called to clean up the view when it's removed.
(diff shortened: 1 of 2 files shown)


bbrk24:
Looks like the Windows build is failing because the CI system has Swift 5.8, which is less than the minimum version required to build SwiftCrossUI (5.10).

Sources/UIKitBackend/UIViewRepresentable.swift
```diff
@@ -0,0 +1,198 @@
...
+        let intrinsicSize = uiView.intrinsicContentSize
+        let sizeThatFits = uiView.sizeThatFits(
+            CGSize(width: CGFloat(proposal.x), height: CGFloat(proposal.y)))
+        let roundedSize = SIMD2(
+            Int(sizeThatFits.width.rounded(.up)),
+            Int(sizeThatFits.height.rounded(.up)))
+        return ViewSize(
+            size: SIMD2(
+                intrinsicSize.width < 0.0 ? proposal.x : min(proposal.x, roundedSize.x),
+                intrinsicSize.height < 0.0 ? proposal.y : min(proposal.y, roundedSize.y)
```
stackotter: I believe `min(proposal._, roundedSize._)` should just be `roundedSize._`. If a view wants to be bigger than proposed, it always gets the final say and the parent view just has to deal with it.

In-fact, you can probably replace all this `size: ` logic with `size: roundedSize`. UIKit should be behaving such that this code generally returns the same value as UIKit, and if it doesn't return the same value as UIKit then we should go with what UIKit says anyway. The main way I see the current approach going wrong is if you've for example got a resizable image view with a constant aspect ratio; it'd likely have negative intrinsic size so SwiftCrossUI would keep giving it all proposed space rather than the space the view actually wants to take up (which may be shorter or thinner to maintain aspect ratio). The intrinsicSize should instead only be used for `idealSize` and `minimumSize` as you've already done below.

Sources/UIKitBackend/UIViewRepresentable.swift
```diff
@@ -0,0 +1,198 @@
...
+            ),
+            idealSize: SIMD2(
+                intrinsicSize.width < 0.0 ? proposal.x : roundedSize.x,
+                intrinsicSize.height < 0.0 ? proposal.y : roundedSize.y
```
stackotter: Ideal size should be 10 when intrinsic size is negative and equal to the intrinsic size otherwise.

By convention, SwiftUI and SwiftCrossUI use 10 for the default ideal size along each dimension when a view can take on any size (e.g. `Color`). Some views such as a resizable `Image` can take on any size but instead use the dimensions of the underlying image as their ideal size, but in this default implementation we can't really know what the underlying sizing behaviour of the `UIView` is, so 10 is probably the best we can do.

The idea behind `idealSize` is that it should always have the same value for the same content, regardless of proposed size. It's required for things like the `fixedSize()` modifier (which SwiftCrossUI doesn't have yet).

stackotter (changes requested):
Just a few changes to make to the sizing code. Ready to merge other than that.

Sources/UIKitBackend/UIViewRepresentable.swift
bbrk24: The reason I'm doing it this way is that the default implementation of `sizeThatFits(_:)` -- which isn't overridden by e.g. `WKWebView` -- ignores the argument and returns `frame.size`. This means that if the view was initialized with `init()` (or equivalently, `init(frame: .zero)`), `roundedSize` will *always* be 0x0. I can definitely get rid of the `min` call, but I don't want to blindly trust `sizeThatFits` either.

For another data point, `UITextField.sizeThatFits(_:)` appears to ignore its argument and return its `intrinsicContentSize`.

stackotter: Hm ok, then with `min` won't the size still be zero?

bbrk24: No, because these views also report their `intrinsicContentSize` as -1 x -1, so they get caught by the ternary.

stackotter: Ok sounds like you've explored most options. If you remove the `min` I think the implementation is probably as good as it'll get without view-specific assumptions or somehow parsing out and solving the view's constraints.

stackotter (approved):
Looks good!

PR #105, opened 2025-02-16
Add `NSViewRepresentable` for AppKit & support Swift 6 concurrency.

furby-tm:
**Note**: this revision bumps the _minimum **macOS** version from **v10.15** to **v11**_ though we should probably do as the **CI** compiler error for the examples indicate and maintain backwards compatibility via `#available` macros.

Leaving the error in for now, to make obvious there was a minimum version bump and discuss how we should best move forward before merging this in.


Sources/AppKitBackend/BaseWidget.swift
```diff
@@ -0,0 +1,113 @@
+import AppKit
+
+public class BaseWidget: NSView {
```
stackotter, line 3: I believe that you're not actually using `BaseWidget` anywhere? If so this should be removed.

Sources/AppKitBackend/NSViewRepresentable.swift
```diff
@@ -0,0 +1,209 @@
...
+    /// property is what frame the view will actually be rendered with if the current layout
+    /// pass is not a dry run, while the other properties are used to inform the layout engine
+    /// how big or small the view can be. The ``SwiftCrossUI/ViewSize/idealSize`` property
+    /// should not vary with the `proposal`, and should only depend on the view's contents.
+    /// Pass `nil` for the maximum width/height if the view has no maximum size (and therefore
+    /// may occupy the entire screen).
+    ///
+    /// The default implementation uses `nsView.intrinsicContentSize` and `nsView.sizeThatFits(_:)`
+    /// to determine the return value.
+    func determineViewSize(
```
stackotter, line 47: I feel like this should be `@MainActor` as well if the other uiview methods have to be

Sources/AppKitBackend/NSViewRepresentable.swift
```diff
@@ -0,0 +1,209 @@
...
+            representingWidget.width = size.size.x
+            representingWidget.height = size.size.y
+        }
+
+        return ViewUpdateResult.leafView(size: size)
+    }
+}
+
+extension NSViewRepresentable
+where Coordinator == Void {
```
stackotter: Put the where clause on the same line as the extension

Sources/AppKitBackend/NSViewRepresentable.swift
```diff
@@ -0,0 +1,209 @@
...
+            minimumWidth: max(0, roundedIntrinsicSize.x),
+            minimumHeight: max(0, roundedIntrinsicSize.x),
+            maximumWidth: nil,
+            maximumHeight: nil
+        )
+    }
+}
+
+extension View
+where Self: NSViewRepresentable {
```
stackotter: Put the where clause on the same line as the extension

Sources/AppKitBackend/NSViewRepresentable.swift
```diff
@@ -0,0 +1,209 @@
+import SwiftCrossUI
+import AppKit
+
+public struct NSViewRepresentableContext<Coordinator> {
+    public let coordinator: Coordinator
+    public internal(set) var environment: EnvironmentValues
+}
+
+public protocol NSViewRepresentable: View
+where Content == Never {
```
stackotter: Put the where clause on the same line as the protocol

Package.swift
```diff
@@ -61,7 +61,7 @@ switch ProcessInfo.processInfo.environment["SCUI_LIBRARY_TYPE"] {
 
 let package = Package(
     name: "swift-cross-ui",
-    platforms: [.macOS(.v10_15), .iOS(.v13), .tvOS(.v13), .macCatalyst(.v13)],
+    platforms: [.macOS(.v11), .iOS(.v13), .tvOS(.v13), .macCatalyst(.v13)],
```
stackotter: I feel like there was a reason for bumping to macOS 11 but I can't quite remember what it was. If it's at all possible we should use `if #available(...)` & co to maintain 10.15 compatibility.

Sources/UIKitBackend/UIViewRepresentable.swift
```diff
@@ -20,12 +21,14 @@ where Content == Never {
...
     /// - Note: This may be called even when `context` has not changed.
+    @MainActor
     func updateUIView(_ uiView: UIViewType, context: UIViewRepresentableContext<Coordinator>)
 
     /// Make the coordinator for this view.
     ///
     /// The coordinator is used when the view needs to communicate changes to the rest of
     /// the view hierarchy (i.e. through bindings), and is often the view's delegate.
+    @MainActor
     func makeCoordinator() -> Coordinator
```
stackotter, line 32: `determineViewSize` should probably be `@MainActor` too (it's not close enough to your changed lines for me to comment on it directly)

stackotter (changes requested):
Thanks for the PR 🙏 just requested a few changes

stackotter (dismissed):
Looks good now 🙏

stackotter:
Actually, where did the `asWidget` and `update` methods for `NSViewRepresentable`'s `View` implementation go? I don't think the code works in its current form (haven't tested yet though).

furby-tm:
> Actually, where did the `asWidget` and `update` methods for `NSViewRepresentable`'s `View` implementation go? I don't think the code works in its current form (haven't tested yet though).

Lol, oh really? Yeah I need to actually test it -- I simply compiled it is all... what would the implementation look like for those methods? I simply removed them because they were calling into that `BaseWidget` class that you had me remove, so I assumed those methods were not necessary.

furby-tm:
cc. @stackotter ^

furby-tm:
@stackotter upon actual testing I can confirm the latest changes in this PR allow it to work (without crashing & with expected behavior at least with my USD hydra powered viewport app).

### Notable changes
* I had found `BaseWidget` to be quite an important implementation detail as I could not create an `asWidget` method implementation that did not crash without it, it's totally possible I was just missing something obvious -- but will need your eyes to look at it closer to see if it's actually necessary -- my testing told me it was though.
* The added `@MainActor` specification on the `func makeCoordinator() -> Coordinator` created a bit of a domino effect that caused me to litter many of scui's methods with `@MainActor` declarations - following the compiler errors all the way to a successful compilation ended in adding those declarations all the way up to `App`'s `main()` function, but I suppose in hindsight at least that bit makes sense.

furby-tm:
also this is macOS 10.15+ compliant with the `#available` macros being utilized [here](https://github.com/stackotter/swift-cross-ui/pull/105/commits/2208de31cca30f9b2cbc35d1973050fb9b85f80f#diff-bef27dced9a23e49cd7caad50fa3e18e41bb85b9a263ba030fc9154a67629948R68-R74), using `safeAreaLayoutGuide` on more modern versions of macOS.

furby-tm:
After coming across a thread on the swift forums, it looks like there are some issues regarding swift 6 strict concurrency and non-Apple GUI frameworks like GTK, as such I've put in place a "lopsided" approach, which minimally allows things to work with swift 6 strict concurrency but removes the `@MainActor` from `makeCoordinator()`.

So things should be forwards compatible, and we can open up proper support for swift 6 cross-platform once this issue is addressed:
https://forums.swift.org/t/pitch-custom-main-and-global-executors/77247/39

stackotter (approved):
Gonna merge this into a temp branch and test/fix things there

PR #108, opened 2025-03-02
Ensure proper method(s) are supportive of cross-platform swift 6 support

furby-tm:


stackotter:
For future readers, I've just done the final change manually (this is #105 + one more change).

PR #114, opened 2025-03-13
Add keyboardToolbar modifier and extensible environment values

bbrk24:


Sources/UIKitBackend/UIKitBackend+Control.swift
```diff
@@ -202,6 +202,14 @@ extension UIKitBackend {
         textFieldWidget.child.textColor = UIColor(color: environment.suggestedForegroundColor)
         textFieldWidget.onChange = onChange
         textFieldWidget.onSubmit = onSubmit
+
+        if let updateToolbar = environment.updateToolbar {
+            let toolbar = (textFieldWidget.child.inputAccessoryView as? UIToolbar) ?? UIToolbar()
+            updateToolbar(toolbar)
```
stackotter, line 208: I reckon this would be better named `setupKeyboardToolbar`

Sources/SwiftCrossUI/Environment/EnvironmentValues.swift
```diff
@@ -131,3 +144,11 @@ public struct EnvironmentValues {
         return environment
     }
 }
+
+/// A key that can be used to extend the environment over the built-in values.
```
stackotter:
```
A key that can be used to extend the environment with new properties.
```

Sources/SwiftCrossUI/Environment/EnvironmentValues.swift
```diff
@@ -121,6 +133,7 @@ public struct EnvironmentValues {
         colorScheme = .light
         windowScaleFactor = 1
         window = nil
+        extraKeys = [:]
```
stackotter: `additionalValues` could make more sense cause it seems more intuitive to talk about it as storing values rather than storing keys.

Sources/UIKitBackend/KeyboardToolbar.swift
```diff
@@ -0,0 +1,180 @@
...
+
+    public static func buildEither(first component: Component) -> Component {
+        component
+    }
+
+    public static func buildEither(second component: Component) -> Component {
+        component
+    }
+
+    public static func buildFinalResult(_ component: Component) -> [UIBarButtonItem] {
```
stackotter: The final result should be `[Component]` if possible so that we can hide the `UIBarButtonItem` implementation detail from users that aren't implementing custom items (reducing the risk of breaking changes if we generalise this API for android at some point).

Sources/UIKitBackend/KeyboardToolbar.swift
```diff
@@ -0,0 +1,180 @@
+import SwiftCrossUI
+import UIKit
+
+/// An item which can be displayed in a keyboard toolbar. Implementers of this do not have
+/// to implement ``SwiftCrossUI/View``.
```
stackotter, line 5: This doc comment should make it clear that toolbar items are stateless and `@State` properties within conforming types will not act as expected when used in a keyboard toolbar body.

Sources/UIKitBackend/KeyboardToolbar.swift
```diff
@@ -0,0 +1,180 @@
...
+
+@available(iOS 14, macCatalyst 14, tvOS 14, *)
+extension Spacer: ToolbarItem {
+    public func asBarButtonItem() -> UIBarButtonItem {
+        if let minLength, minLength > 0 {
+            print(
+                """
+                Warning: Spacer's minLength is ignored inside keyboard toolbars, as \
+                flexible-length spacers cannot specify a minimum width. Use `.frame(width:)` \
+                for a fixed-length spacer.
```
stackotter:
```
Warning: Spacer's minLength property is ignored within keyboard toolbars \
due to UIKit limitations. Use `Spacer()` for unconstrained spacers \
and `Spacer().frame(width: _)` for fixed-length spacers.
```

Sources/UIKitBackend/KeyboardToolbar.swift
```diff
@@ -0,0 +1,180 @@
...
+        item.tintColor = color.uiColor
+        return item
+    }
+}
+
+extension ToolbarItem {
+    /// A toolbar item with the specified width.
+    ///
+    /// If `width` is positive, the item will have that exact width. If `width` is zero or
+    /// nil, the item will have its natural size.
```
stackotter, line 136: Update this to clarify that using this on spacers is a no-op on iOS 13.

stackotter (changes requested):
Thanks for this 🙏 Just suggested a few documentation updates and variable name replacements to improve clarity, but for the most part it all looks great. I'll test it on my own device shortly

Sources/UIKitBackend/KeyboardToolbar.swift
bbrk24: Spacers can't be used as toolbar items at all on iOS 13.

Sources/UIKitBackend/UIKitBackend+Control.swift
bbrk24: I named it along the lines of the various `update*` methods of the backends, which are also used for initial setup. This is also used to change the toolbar live if needed, e.g. tapping one button changes a piece of state that shows/hides another:

```swift
struct Example: View {
  @State var showSecondButton = false
  @State var text = ""

  var body: some View {
    TextField($text).keyboardToolbar {
      Button("first") {
        showSecondButton.toggle()
      }
 
      Spacer()

      if showSecondButton {
        Button("second") {}
      }
    }
  }
}
```

Sources/UIKitBackend/KeyboardToolbar.swift
bbrk24: Given that `Component` is already an array it would be easiest to just remove `buildFinalResult` entirely.

Sources/UIKitBackend/KeyboardToolbar.swift
bbrk24: Added a paragraph explaining this

Sources/SwiftCrossUI/Environment/EnvironmentValues.swift
bbrk24: Renamed to `extraValues`

Sources/UIKitBackend/KeyboardToolbar.swift
stackotter: Ah yep my bad, I misremembered the limitation

Sources/UIKitBackend/UIKitBackend+Control.swift
stackotter: That explanation makes sense 👍

stackotter (approved):
Thanks for the changes, looks great now. I'll test locally then merge.

PR #115, opened 2025-03-15
Cache toolbar items for update

bbrk24:
prevents buttons from flashing on view updates


stackotter (approved):
Cool approach, thanks for these changes 🙏

PR #116, opened 2025-03-15
Added AlertAction public initializer

DrAma999:
Added `public` `AlertAction` initializer.
I don't know if there was a specific reason behind it, but the `AlertAction` initializer was missing there was only a default `.ok`.

### DIFF
--- a/Sources/SwiftCrossUI/Modifiers/AlertAction.swift
+++ b/Sources/SwiftCrossUI/Modifiers/AlertAction.swift
@@ -9,4 +9,11 @@ public struct AlertAction {
 
     public var label: String
     public var action: () -> Void
+
+    public init(
+        label: String, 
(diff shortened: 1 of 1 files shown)


stackotter:
Hi Andrea, apologies that this isn't documented anywhere, but the intended way to use the alert modifier is like so; (the `@AlertActionsBuilder` result builder turns buttons into alert actions)

```swift
myView.alert("Delete file?", isPresented: $isPresented) {
    Button("Ok") {
        print("Ok")
    }
    Button("Cancel") {
        print("Cancel")
    }
}
```

The reason that `AlertAction` doesn't have a public init is that I want it to be kept as a bit of an internal implementation detail for now so that I have the freedom to swap it out once I have a better solution.

Hope that helps!

DrAma999:
Thank you! I see.

PR #117, opened 2025-03-16
More tap gestures

bbrk24:
Implements right-click and long-press. Currently not implemented in Gtk3 or WinUI. I'm also not sold on the argument label because `onTapGesture(gesture:) {}` looks a little odd.

### DIFF
--- a/Sources/GtkBackend/GtkBackend.swift
+++ b/Sources/GtkBackend/GtkBackend.swift
@@ -961,7 +961,7 @@ public final class GtkBackend: AppBackend {
                 gtkGesture = GestureClick()
             case .secondary:
                 gtkGesture = GestureClick()
-                gtk_gesture_single_set_button(gtkGesture.opaquePointer, 3)
+                gtk_gesture_single_set_button(gtkGesture.opaquePointer, guint(GDK_BUTTON_SECONDARY))
             case .longPress:
                 gtkGesture = GestureLongPress()
         }
@@ -978,7 +978,8 @@ public final class GtkBackend: AppBackend {
             case .primary:
                 let gesture =
                     tapGestureTarget.eventControllers.first {
-                        $0 is GestureClick && gtk_gesture_single_get_button($0.opaquePointer) == 1
+                        $0 is GestureClick
+                            && gtk_gesture_single_get_button($0.opaquePointer) == GDK_BUTTON_PRIMARY
                     } as! GestureClick
                 gesture.pressed = { _, nPress, _, _ in
                     guard nPress == 1 else {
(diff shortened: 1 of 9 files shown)


bbrk24:
Mac & Linux are failing because GTK3. But why is Windows *succeeding*

stackotter:
The Windows CI doesn't build WinUIBackend at the moment due to some annoying Swift on Windows issues. We can't use anything newer than Swift 5.10 cause of some weird issue where SwiftPM fails to compile the manifest due to missing CPP headers, but swift-winui requires Swift 6.0

Sources/SwiftCrossUI/Backend/AppBackend.swift
```diff
@@ -508,11 +508,12 @@ public protocol AppBackend {
     /// Wraps a view in a container that can receive tap gestures. Some
     /// backends may not have to wrap the child, in which case they may
     /// just return the child as is.
-    func createTapGestureTarget(wrapping child: Widget) -> Widget
+    func createTapGestureTarget(wrapping child: Widget, gesture: TapGesture) -> Widget
```
stackotter, line 511: Update the placeholder implementations at the bottom of this file to match.

Sources/GtkBackend/GtkBackend.swift
```diff
@@ -954,22 +954,56 @@ public final class GtkBackend: AppBackend {
...
-        gesture.pressed = { _, nPress, _, _ in
-            guard nPress == 1 else {
-                return
-            }
-            action()
+        switch gesture.kind {
+            case .primary:
+                let gesture =
+                    tapGestureTarget.eventControllers.first {
+                        $0 is GestureClick && gtk_gesture_single_get_button($0.opaquePointer) == 1
```
stackotter: Use `GDK_BUTTON_PRIMARY` and `GDK_BUTTON_SECONDARY` instead of `1` and `3` for primary and secondary clicks.

stackotter (approved):
Thanks for these improvements! 🙏

stackotter (changes requested):
Whoops I meant to request changes instead of approve hahah, just two minor things to change.

Sources/GtkBackend/GtkBackend.swift
bbrk24: I tried this, and I'm getting errors like

> Cannot find 'GTK_BUTTON_SECONDARY' in scope

bbrk24: Oh, wait, G**D**K. Those constants are found but they're the wrong type (`Int32` vs `UInt32`).

stackotter: Should be able to just convert the type, I've confirmed that they're the right values. Use `guint(...)` to convert it in case they ever change the underlying type or it's different on some strange system etc.

stackotter (approved):
Lgtm

PR #118, opened 2025-03-19
Implement Menu in UIKitBackend

### LINKED ISSUE #106: UIKitBackend support for Menu
The current methods on `AppBackend` for creating and updating menus are incompatible with how `UIMenu` works.

There are two ways to display a context menu in UIKit:
 
1. Create the `UIMenu` object in advance and attach it to a `UIButton`.
2. Use a [`UIContextMenuInteraction`](https://developer.apple.com/documentation/uikit/uicontextmenuinteraction), which listens for a 3D Touch, long press, right-click, etc depending on the device. When it detects the appropriate form of interaction, it asks its delegate for a [`UIContextMenuConfiguration`](https://developer.apple.com/documentation/uikit/uicontextmenuconfiguration) object, whose initializer takes a callback that returns `UIMenu`. It sounds convoluted, but it lets you defer creation of the `UIMenu` object until it's actually displayed. The problem lies in the particular form of interaction it requires.

The former is closer to how `Menu` actually works:
https://github.com/stackotter/swift-cross-ui/blob/4de77096a0f767ab3dfdccc462791c6529f8eba9/Sources/SwiftCrossUI/Views/Menu.swift#L74-L90
However, UIKitBackend can't implement `showPopoverMenu` to accommodate this exact code. It would have to know that the button is being used for a context menu rather than an arbitrary action, and associate the menu with the button.

bbrk24:
Fixes #106

### DIFF
--- a/Sources/SwiftCrossUI/Backend/AppBackend.swift
+++ b/Sources/SwiftCrossUI/Backend/AppBackend.swift
@@ -84,7 +84,8 @@ public protocol AppBackend {
     /// manually rescale the image meaning that it must get rescaled when the
     /// scale factor changes.
     var requiresImageUpdateOnScaleFactorChange: Bool { get }
-    /// How the backend handles menu rendering. Affects which menu-related methods are called.
+    /// How the backend handles rendering of menu buttons. Affects which menu-related methods
+    /// are called.
     var menuImplementationStyle: MenuImplementationStyle { get }
 
     /// Often in UI frameworks (such as Gtk), code is run in a callback
@@ -345,8 +346,8 @@ public protocol AppBackend {
         action: @escaping () -> Void,
         environment: EnvironmentValues
     )
-    /// Sets a button's label and menu. May not be called depending on the value of
-    /// ``menuImplementationStyle``.
+    /// Sets a button's label and menu. Only used when ``menuImplementationStyle`` is
+    /// ``MenuImplementationStyle/menuButton``.
     func updateButton(
         _ button: Widget,
         label: String,
@@ -442,8 +443,7 @@ public protocol AppBackend {
     )
 
     /// Creates a popover menu (the sort you often see when right clicking on
-    /// apps). The menu won't be visible until you call
-    /// ``AppBackend/showPopoverMenu(_:at:relativeTo:closeHandler:)``.
+    /// apps). The menu won't be visible when first created.
     func createPopoverMenu() -> Menu
     /// Updates a popover menu's content and appearance.
     func updatePopoverMenu(
@@ -451,8 +451,8 @@ public protocol AppBackend {
         content: ResolvedMenu,
         environment: EnvironmentValues
     )
-    /// Shows the popover menu at a position relative to the given widget. May not be called
-    /// depending on the value of ``menuImplementationStyle``.
+    /// Shows the popover menu at a position relative to the given widget. Only used when
+    /// ``menuImplementationStyle`` is ``MenuImplementationStyle/dynamicPopover``.
     func showPopoverMenu(
         _ menu: Menu,
         at position: SIMD2<Int>,
--- a/Sources/SwiftCrossUI/Backend/AppBackend.swift
+++ b/Sources/SwiftCrossUI/Backend/AppBackend.swift
@@ -84,6 +84,8 @@ public protocol AppBackend {
     /// manually rescale the image meaning that it must get rescaled when the
     /// scale factor changes.
     var requiresImageUpdateOnScaleFactorChange: Bool { get }
+    /// How the backend handles menu rendering. Affects which menu-related methods are called.
+    var menuImplementationStyle: MenuImplementationStyle { get }
 
     /// Often in UI frameworks (such as Gtk), code is run in a callback
     /// after starting the app, and hence this generic root window creation
@@ -343,6 +345,14 @@ public protocol AppBackend {
         action: @escaping () -> Void,
         environment: EnvironmentValues
     )
+    /// Sets a button's label and menu. May not be called depending on the value of
+    /// ``menuImplementationStyle``.
+    func updateButton(
+        _ button: Widget,
+        label: String,
+        menu: Menu,
+        environment: EnvironmentValues
+    )
 
     /// Creates a labelled toggle that is either on or off. Predominantly used by
     /// ``Toggle``.
@@ -441,7 +451,8 @@ public protocol AppBackend {
         content: ResolvedMenu,
         environment: EnvironmentValues
     )
-    /// Shows the popover menu at a position relative to the given widget.
+    /// Shows the popover menu at a position relative to the given widget. May not be called
+    /// depending on the value of ``menuImplementationStyle``.
     func showPopoverMenu(
         _ menu: Menu,
         at position: SIMD2<Int>,
@@ -664,6 +675,14 @@ extension AppBackend {
         _ button: Widget,
         label: String,
         action: @escaping () -> Void,
+        environment: EnvironmentValues
+    ) {
+        todo()
+    }
+    public func updateButton(
+        _ button: Widget,
+        label: String,
+        menu: Menu,
         environment: EnvironmentValues
     ) {
         todo()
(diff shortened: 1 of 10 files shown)


Sources/UIKitBackend/UIKitBackend+Menu.swift
```diff
@@ -0,0 +1,66 @@
...
+extension UIKitBackend {
+    public final class Menu {
+        var uiMenu: UIMenu?
+    }
+
+    public func createPopoverMenu() -> Menu {
+        return Menu()
+    }
+
+    static func transformMenu(
```
stackotter: Rename to `createMenu` or `buildMenu` or something to that effect

Sources/UIKitBackend/UIKitBackend+Control.swift
```diff
@@ -2,7 +2,15 @@ import SwiftCrossUI
 import UIKit
 
 final class ButtonWidget: WrapperWidget<UIButton> {
-    var onTap: (() -> Void)?
+    private let event: UIControl.Event
+
+    var onTap: (() -> Void)? {
+        didSet {
+            if oldValue == nil {
+                child.addTarget(self, action: #selector(buttonTapped), for: event)
```
stackotter, line 10: Just out of interest, why did you move this to here instead of leaving it in the init?

Sources/SwiftCrossUI/Backend/MenuImplementationStyle.swift
```diff
@@ -0,0 +1,18 @@
...
+/// ``AppBackend/createPopoverMenu()``, ``AppBackend/updatePopoverMenu(_:content:environment:)``,
+/// and ``AppBackend/updateButton(_:label:action:environment:)``.
+public enum MenuImplementationStyle {
+    /// The backend can show popover menus arbitrarily.
+    ///
+    /// Backends that use this style must implement
+    /// ``AppBackend/showPopoverMenu(_:at:relativeTo:closeHandler:)``. For these backends,
+    /// ``AppBackend/createPopoverMenu()`` is not called until after the button is tapped.
+    case dynamicPopover
+    /// The backend requires menus to be constructed and passed to buttons ahead-of-time.
```
stackotter: Change passed to attached

Sources/SwiftCrossUI/Backend/AppBackend.swift
```diff
@@ -441,7 +451,8 @@ public protocol AppBackend {
         content: ResolvedMenu,
         environment: EnvironmentValues
     )
-    /// Shows the popover menu at a position relative to the given widget.
+    /// Shows the popover menu at a position relative to the given widget. May not be called
+    /// depending on the value of ``menuImplementationStyle``.
```
stackotter: Update this to clarify that this method is only required for `dynamicPopover` style

Sources/SwiftCrossUI/Backend/AppBackend.swift
```diff
@@ -343,6 +345,14 @@ public protocol AppBackend {
         action: @escaping () -> Void,
         environment: EnvironmentValues
     )
+    /// Sets a button's label and menu. May not be called depending on the value of
+    /// ``menuImplementationStyle``.
```
stackotter: Update this to clarify that this method is only required for `menuButton` style.

Sources/SwiftCrossUI/Backend/AppBackend.swift
```diff
@@ -84,6 +84,8 @@ public protocol AppBackend {
     /// manually rescale the image meaning that it must get rescaled when the
     /// scale factor changes.
     var requiresImageUpdateOnScaleFactorChange: Bool { get }
+    /// How the backend handles menu rendering. Affects which menu-related methods are called.
```
stackotter: `How the backend handlers menu rendering for menu buttons...`

Sources/SwiftCrossUI/Views/Menu.swift
```diff
@@ -71,27 +74,45 @@ public struct Menu: TypeSafeView {
...
+                )
+
+                if !dryRun {
+                    backend.setSize(of: widget, to: size)
+                    children.updateMenuIfShown(
+                        content: content, environment: environment, backend: backend)
+                }
+            case .menuButton:
+                let menu = children.menu as? Backend.Menu ?? backend.createPopoverMenu()
                 children.menu = menu
```
stackotter, line 115: I believe this is redundant?

Sources/SwiftCrossUI/Views/Menu.swift
```diff
@@ -71,27 +74,45 @@ public struct Menu: TypeSafeView {
...
+                    label: label,
+                    action: {
+                        let menu = backend.createPopoverMenu()
+                        children.menu = menu
+                        backend.updatePopoverMenu(
+                            menu,
+                            content: content,
+                            environment: environment
+                        )
+                        backend.showPopoverMenu(menu, at: SIMD2(0, size.y + 2), relativeTo: widget)
```
stackotter: Line wrap it like this (apologies that the formatter config doesn't do this automatically);

```swift
function(
    arg1,
    arg2,
    arg3
) {
    ...
}
```

Sources/SwiftCrossUI/Views/Menu.swift
```diff
@@ -71,27 +74,45 @@ public struct Menu: TypeSafeView {
...
+                            children.menu = nil
+                        }
+                    },
+                    environment: environment
+                )
+
+                if !dryRun {
+                    backend.setSize(of: widget, to: size)
+                    children.updateMenuIfShown(
+                        content: content, environment: environment, backend: backend)
```
stackotter: Line wrap this the same way as the above review comment

stackotter (changes requested):
Thanks looks good, just a few small naming and formatting changes

Sources/UIKitBackend/UIKitBackend+Control.swift
bbrk24: This way the target-action pair isn't registered for menu buttons.

Sources/SwiftCrossUI/Views/Menu.swift
bbrk24: This is a pattern I've used elsewhere before. The cast obscures it a little in this case, but the basic pattern is
```swift
let x = y.x ?? createX()
y.x = x
```
This ensures `x` exists and is attached to `y`. In this particular case it reduces the number of `createPopoverMenu` calls by caching the Menu object.

Sources/UIKitBackend/UIKitBackend+Menu.swift
bbrk24: I originally chose this name because it transforms the menu from `ResolvedMenu` to `UIMenu`, but I can see the argument.

Sources/UIKitBackend/UIKitBackend+Control.swift
stackotter: Ah ok yep, cause `onTap` isn't used at all by menu buttons

Sources/UIKitBackend/UIKitBackend+Menu.swift
stackotter: Yeah I can see that making sense too. I guess the part that makes `transform` feel strange to me is that the method performs one specific transformation and the result is an opaque-ish value (i.e. we treat `UIMenu` as an opaque blob to attach to buttons etc). In my mental model I think of that as building a final result or creating a resource rather than transforming from one data structure to another.

Sources/SwiftCrossUI/Views/Menu.swift
stackotter: Ah sorry I didn't pay enough attention. I didn't realise that `children.menu` was used in the expression for `let menu = ...`. I thought it wasn't used anywhere in the `menuButton` style. My bad 👍

stackotter (changes requested):
Just submitting this review to save my reviewing progress, no new requested changes other than the existing `transform` discussion.

stackotter (approved):
Looks good now, thanks for implementing this 🙏 I'll test it out locally and then merge

PR #119, opened 2025-03-25
Build UIKitBackend in CI

bbrk24:
Only includes the build step and not the test step because `xcodebuild -scheme swift-cross-ui-Package test` errors out looking for GTK headers.

### DIFF
--- a/.github/workflows/swift-uikit.yml
+++ b/.github/workflows/swift-uikit.yml
@@ -1,4 +1,4 @@
-name: Build macOS
+name: Build UIKitBackend
 
 on:
   push:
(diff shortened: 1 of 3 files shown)


bbrk24:
Maybe I should've done a matrix rather than a bash for loop 😅

bbrk24:
After piping the logs through xcpretty:

```
❌  /Users/runner/work/swift-cross-ui/swift-cross-ui/Sources/UIKitBackend/KeyboardToolbar.swift:197:30: 'UIToolbar' is unavailable in tvOS

final class KeyboardToolbar: UIToolbar {
                             ^~~~~~~~~
```

It seems to have found a legitimate error. I'll fix that I guess.

bbrk24:
As stupid as this looks:
```swift
@available(iOS 14, macCatalyst 14, tvOS 14, *)
@available(tvOS, unavailable)
```
It failed to compile if I removed `tvOS 14` from the former annotation.

PR #123, opened 2025-04-17
Add Path and Shape

bbrk24:
Not currently implemented in GtkBackend or Gtk3Backend, but I may be able to do the former in the near future

### DIFF
--- a/Sources/SwiftCrossUI/Backend/AppBackend.swift
+++ b/Sources/SwiftCrossUI/Backend/AppBackend.swift
@@ -44,6 +44,7 @@ public protocol AppBackend {
     associatedtype Widget
     associatedtype Menu
     associatedtype Alert
+    associatedtype Path
 
     /// Creates an instance of the backend.
     init()
@@ -527,6 +528,24 @@ public protocol AppBackend {
         gesture: TapGesture,
         action: @escaping () -> Void
     )
+
+    // MARK: Paths
+    /// Create a path. It will not be shown until ``renderPath(_:container:)`` is called.
+    func createPath() -> Path
+    /// Update a path. The updates do not need to be visible before ``renderPath(_:container:)``
+    /// is called.
+    /// - Parameters:
+    ///   - path: The path to be updated.
+    ///   - source: The source to copy the path from.
+    ///   - pointsChanged: If `false`, the ``Path/actions`` of the source have not changed.
+    func updatePath(_ path: Path, _ source: SwiftCrossUI.Path, pointsChanged: Bool)
+    /// Draw a path to the screen.
+    /// - Parameters:
+    ///   - path: The path to be rendered.
+    ///   - container: The container widget that the path will render in. It has no other
+    ///     children.
+    ///   - environment: The environment values, including color.
+    func renderPath(_ path: Path, container: Widget, environment: EnvironmentValues)
 }
 
 extension AppBackend {
@@ -841,6 +860,17 @@ extension AppBackend {
         gesture: TapGesture,
         action: @escaping () -> Void
     ) {
+        todo()
+    }
+
+    // MARK: Paths
+    func createPath() -> Path {
+        todo()
+    }
+    func updatePath(_ path: Path, _ source: SwiftCrossUI.Path, pointsChanged: Bool) {
+        todo()
+    }
+    func renderPath(_ path: Path, container: Widget, environment: EnvironmentValues) {
         todo()
     }
 }
--- a/Sources/SwiftCrossUI/Backend/AppBackend.swift
+++ b/Sources/SwiftCrossUI/Backend/AppBackend.swift
@@ -544,8 +544,16 @@ public protocol AppBackend {
     ///   - path: The path to be rendered.
     ///   - container: The container widget that the path will render in. It has no other
     ///     children.
-    ///   - environment: The environment values, including color.
-    func renderPath(_ path: Path, container: Widget, environment: EnvironmentValues)
+    ///   - strokeColor: The color to draw the path's stroke.
+    ///   - fillColor: The color to shade the path's fill.
+    ///   - overrideStrokeStyle: If present, a value to override the path's stroke style.
+    func renderPath(
+        _ path: Path,
+        container: Widget,
+        strokeColor: Color,
+        fillColor: Color,
+        overrideStrokeStyle: StrokeStyle?
+    )
 }
 
 extension AppBackend {
@@ -870,7 +878,13 @@ extension AppBackend {
     func updatePath(_ path: Path, _ source: SwiftCrossUI.Path, pointsChanged: Bool) {
         todo()
     }
-    func renderPath(_ path: Path, container: Widget, environment: EnvironmentValues) {
+    func renderPath(
+        _ path: Path,
+        container: Widget,
+        strokeColor: Color,
+        fillColor: Color,
+        overrideStrokeStyle: StrokeStyle?
+    ) {
         todo()
     }
 }
--- a/Sources/SwiftCrossUI/Backend/AppBackend.swift
+++ b/Sources/SwiftCrossUI/Backend/AppBackend.swift
@@ -530,6 +530,8 @@ public protocol AppBackend {
     )
 
     // MARK: Paths
+    /// Create a widget that can contain a path.
+    func createPathWidget() -> Widget
     /// Create a path. It will not be shown until ``renderPath(_:container:)`` is called.
     func createPath() -> Path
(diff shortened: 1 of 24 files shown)


Sources/SwiftCrossUI/Backend/AppBackend.swift
```diff
@@ -527,6 +528,34 @@ public protocol AppBackend {
...
+    /// Create a widget that can contain a path.
+    func createPathWidget() -> Widget
+    /// Create a path. It will not be shown until ``renderPath(_:container:)`` is called.
+    func createPath() -> Path
+    /// Update a path. The updates do not need to be visible before ``renderPath(_:container:)``
+    /// is called.
+    /// - Parameters:
+    ///   - path: The path to be updated.
+    ///   - source: The source to copy the path from.
+    ///   - pointsChanged: If `false`, the ``Path/actions`` of the source have not changed.
```
stackotter, line 542: I feel like `shapeChanged` could be a batter description, cause I can see it being ambiguous from the name of the property alone whether `pointsChanged` would be true or false when the points are the same but the interpolation has changed. `shape` might not be the best word but the only good word I can think of is `path` which is of course confusing in its own right due to SwiftCrossUI's path type including the stroke style.

stackotter: Another alternative could be `onlyStyleChanged`

Sources/SwiftCrossUI/Path.swift
```diff
@@ -0,0 +1,373 @@
...
+    /// - Parameters:
+    ///   - center: The location of the center of the circle.
+    ///   - radius: The radius of the circle.
+    ///   - startAngle: The angle of the start of the arc, measured in radians clockwise from
+    //      right. Must be between 0 and 2pi (inclusive).
+    ///   - endAngle: The angle of the end of the arc, measured in radians clockwise from right.
+    ///     Must be between 0 and 2pi (inclusive).
+    ///   - clockwise: `true` if the arc is to be drawn clockwise, `false` if the arc is to
+    ///     be drawn counter-clockwise. Used to determine whether to draw the larger arc or
+    ///     the smaller arc identified by the given start and end angles.
```
stackotter: I feel like referring to the larger arc and the smaller arc is probably a bit confusing because they could be the same size, the clockwise arc could be larger, or the clockwise arc could be smaller. Maybe just;

> ... Used to determine which of the two possible arcs to draw between the given start and end angles.

Sources/SwiftCrossUI/Views/Shapes/Shape.swift
```diff
@@ -0,0 +1,125 @@
+/// A 2-D shape that can be drawn as a view.
+///
+/// If no stroke color or fill color is specified, the default is no stroke and a fill of the
+/// current foreground color.
+public protocol Shape: View
+where Content == EmptyView {
```
stackotter: Put these on the same line

Sources/SwiftCrossUI/Views/Shapes/Shape.swift
```diff
@@ -0,0 +1,125 @@
...
+    /// Determine the ideal size of this shape given the proposed bounds.
+    ///
+    /// The default implementation accepts the proposal and imposes no practical limit on
+    /// the shape's size.
+    /// - Returns: Information about the shape's size. The ``ViewSize/size`` property is what
+    ///   frame the shape will actually be rendered with if the current layout pass is not
+    ///   a dry run, while the other properties are used to inform the layout engine how big
+    ///   or small the shape can be. The ``ViewSize/idealSize`` property should not vary with
+    ///   the `proposal`, and should only depend on the shape's contents. Pass `nil` for the
+    ///   maximum width/height if the shape has no maximum size (and therefore may occupy
```
stackotter: Remove the comment comment about occupying the entire screen because I can see it being a bit misleading; a shape with unlimited size may still occupy less than the entire screen if it likes to keep its aspect ratio etc.

Sources/UIKitBackend/UIColor+Color.swift
```diff
@@ -27,4 +27,9 @@ extension Color {
     var uiColor: UIColor {
         UIColor(color: self)
     }
+
+    var cgColor: CGColor {
+        CGColor(
+            red: CGFloat(red), green: CGFloat(green), blue: CGFloat(blue), alpha: CGFloat(alpha))
```
stackotter: Line wrap these

Sources/UIKitBackend/UIKitBackend+Path.swift
```diff
@@ -0,0 +1,177 @@
...
+        let widget = container as! PathWidget
+        let shapeLayer = widget.shapeLayer
+
+        shapeLayer.path = path.cgPath
+        shapeLayer.lineWidth = path.lineWidth
+        shapeLayer.miterLimit = path.miterLimit
+        shapeLayer.fillRule = path.usesEvenOddFillRule ? .evenOdd : .nonZero
+
+        shapeLayer.lineJoin =
+            switch path.lineJoinStyle {
```
stackotter: The CI checks are warning about CGLineJoin and CGLineCap not being frozen enums. It's probably best to handle the `@unknown default:` case with a printed warning and a sensible default.

Sources/WinUIBackend/WinUIBackend.swift
```diff
@@ -1163,6 +1162,259 @@ public final class WinUIBackend: AppBackend {
...
+                case .winding:
+                    .nonzero
+            }
+    }
+
+    func requirePathFigure(
+        _ collection: WinUI.GeometryCollection,
+        lastPoint: Point
+    ) -> PathFigure {
+        var pathGeo: PathGeometry
```
stackotter: Name this pathGeometry unless there some reason it can't be called that

Sources/WinUIBackend/WinUIBackend.swift
```diff
@@ -1163,6 +1162,259 @@ public final class WinUIBackend: AppBackend {
...
+        for action in actions {
+            switch action {
+                case .moveTo(let point):
+                    lastPoint = Point(x: Float(point.x), y: Float(point.y))
+
+                    if geometry.size > 0,
+                        let pathGeo = geometry.getAt(geometry.size - 1) as? PathGeometry,
+                        pathGeo.figures.size > 0
+                    {
+                        let figure = pathGeo.figures.getAt(pathGeo.figures.size - 1)!
```
stackotter: Move this into the if condition to avoid the force unwrap

stackotter (changes requested):
Sorry for taking a while to get around to this.

Thanks so much for getting this to where it is, it's looking great. I've just left a few small change requests but they're mostly formatting, naming and documentation (which were all great for the most part).

I'll test this out locally in a bit more depth once I've finished off my assignment due soon. I'm probably gonna try making a graphing example with bar charts, line charts and pie charts to stress test everything.

Sources/UIKitBackend/UIKitBackend+Path.swift
bbrk24: If I add a print statement then this can no longer be a switch expression :/

but yeah, what you said makes sense

Sources/WinUIBackend/WinUIBackend.swift
bbrk24: The comment from ~30 lines earlier also applies here. Should I change both? I don't think the result can ever be nil in practice.

Sources/SwiftCrossUI/Backend/AppBackend.swift
bbrk24: What about `actionsChanged`? Since it's identifying whether the `Path.actions` property has changed

Sources/SwiftCrossUI/Views/Shapes/Shape.swift
bbrk24: done

Sources/SwiftCrossUI/Views/Shapes/Shape.swift
bbrk24: done

Sources/UIKitBackend/UIColor+Color.swift
bbrk24: done

Sources/WinUIBackend/WinUIBackend.swift
bbrk24: done

stackotter (approved):
Looks good now, thanks for the PR!

PR #129, opened 2025-04-29
Implement textContentType modifier

bbrk24:


Sources/SwiftCrossUI/Environment/EnvironmentValues.swift
```diff
@@ -38,6 +38,9 @@ public struct EnvironmentValues {
     /// The scale factor of the current window.
     public var windowScaleFactor: Double
 
+    /// The type of on-screen keyboard to show when a text field is focused.
```
stackotter: Update to mention that it still affects auto complete on some non-mobile platforms.

Sources/AppKitBackend/AppKitBackend.swift
```diff
@@ -641,6 +641,22 @@ public final class AppKitBackend: AppBackend {
...
+                switch environment.textContentType {
+                    case .url:
+                        .URL
+                    case .phoneNumber:
+                        .telephoneNumber
+                    case .name:
+                        .name
+                    case .emailAddress:
+                        .emailAddress
+                    default:
```
stackotter: Switch over remaining cases so that we're forced to evaluate how to handle new text content types when we add them in future (same goes for other backend implementations).

stackotter (changes requested):
Looks great! Documentation is well written and the API is nice. Just a few tweaks and then I'm happy to merge.

Sources/AppKitBackend/AppKitBackend.swift
bbrk24: > same goes for other backend implementations

I'm updating AppKit now, but GTK doesn't use textContentType at all, and UIKit and WinUI don't have `default:` branches. Could you clarify what you meant by this?

stackotter: At the time I left that comment I hadn't read through the other backends yet, so I just added that as a catch all. Only happens to apply to AppKitBackend though, so ignore that

Sources/WinUIBackend/WinUIBackend.swift
```diff
@@ -801,6 +801,38 @@ public final class WinUIBackend: AppBackend {
...
+            }
+
+        setInputScope(for: textField, to: inputScope)
+    }
+
+    private func setInputScope(for textField: TextBox, to value: InputScopeNameValue) {
+        if let inputScope = textField.inputScope,
+            inputScope.names.count == 1
+        {
+            inputScope.names[0] = value
```
stackotter, line 830: Replacing `value` with `InputScopeName(value)` should hopefully fix the Windows CI, unless more issues pop up once you fix that. You'll have to update the occurrence at line 833 as well.

stackotter (approved):
Merging now 👍

PR #133, opened 2025-05-05
Fix height of window in RandomNumberGeneratorExample

valeriyvan:
Before change:

Screenshot from 2025-05-05 12-57-33

After change:

Screenshot from 2025-05-05 12-57-00


stackotter:
I'll fix this a different a way cause I think it's actually a symptom of an issue with the way GtkBackend implements default window sizing.

Issue #80, opened 2024-04-07
Figure out how to run tests without SwiftPM trying to build all targets


stackotter:
If you run `swift test` now, SwiftPM first tries to build all targets (even though the only tests target has a single dependency on the SwiftCrossUI target). This has the outcome of meaning that you need to have all backend system dependencies present (e.g. gtk4, qt@5) on your computer to run `swift test` (very annoying). Due to this, I've disabled the `swift test` step in macOS and Linux CI until a solution is found.

lhoward:
I ran into a similar issue with one of my repos, never found a solution, so curious if you find one.

stackotter:
My only idea so far is to make Package.swift remove all non-tested targets when a `TEST` environment variable is present so that you can run `TEST=1 swift test`, but that's pretty hacky.

DivineDominion:
I ran into the issue of `swift build` trying to compile the WinUI backend today, too. I believe the way to go is to assemble and mutate the packages and targets collections imperatively based on e.g. `#if os(...)` and similar flags.

Maybe that would be a good start? I've opened #144 with a concrete proposal.

stackotter:
I've commented on the conditional compilation approach over on #144, but I'll include my current workaround for `swift test` here so that I can close this issue.

```
swift test --test-product swift-cross-uiPackageTests
```

PR #144, opened 2025-05-10
Draft: Conditional compilation of Package parts

DivineDominion:
I haven't looked into how to detect GTK3, yet. Opening this for discussion after reading #80 -- which failed on my machine before I installed GTK4, for example.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -11,6 +11,10 @@ if let version = getGtk4MinorVersion(), version >= 10 {
     gtkSwiftSettings.append(.define("GTK_4_10_PLUS"))
 }
 
+// On macOS, test both code paths with `brew unlink gtk4`, `brew link gtk4` to make the lookup fail/succeed.
+let didFindGTK4 = getGtk4MinorVersion() != nil
+let didFindGTK3 = isGtk3Installed()
+
 let defaultBackend: String
 if let backend = ProcessInfo.processInfo.environment["SCUI_DEFAULT_BACKEND"] {
     defaultBackend = backend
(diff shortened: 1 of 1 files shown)


DivineDominion:
So a huge downside of this is that Package.resolved cannot be checked in anymore -- because it'll change on each and every machine. :/

stackotter:
> So a huge downside of this is that Package.resolved cannot be checked in anymore -- because it'll change on each and every machine. :/

Yeah that's the reason I removed conditional compilation from the package manifest (I did originally have it).

Another issue was that the detection sometimes failed when people used alternative installation methods for Gtk (making it impossible for them to use GtkBackend even though they have all the required dependencies).

And the issue that finally convinced me to remove conditional compilation was that if someone added `GtkBackend` as a dependency without gtk installed, they'd get "No such product 'GtkBackend'" instead of "gtk/gtk.h not found", the latter of which I think is more intuitive.

---

That said, as you've noticed, running `swift build` or `swift test` doesn't really work anymore. The former is workable but the latter is pretty annoying. Luckily I recently found a workaround for running `swift test` without building every single product in the package; `swift test --test-product swift-cross-uiPackageTests`

DivineDominion:
I really don't know whether checking in `Package.resolved` is an important in this setup. On one hand, it pins dependencies that are known to work. On the other, it makes working on the package a bit weird. I can't really weigh the trade-offs. I'm fine with closing this without merging, no harm done.

stackotter:
> I really don't know whether checking in `Package.resolved` is an important in this setup.

Yeah, `Package.resolved` wasn't the thing that caused me to remove the conditionals, it's of course a bit annoying not having it but it wasn't that important to me at the time (probably more important once we have a 1.0).

---

I'll close this PR, but I'll create a helper script for running tests and update `CONTRIBUTING.md` to comment on this issue and the workaround.

Issue #153, opened 2025-05-14
error: cannot find 'gtk_list_box_remove_all' in scope


valeriyvan:
Have this build error.What does it mean?

```
/*****/checkouts/swift-cross-ui/Sources/Gtk/Widgets/ListBox+ManualAdditions.swift:14:9: error: cannot find 'gtk_list_box_remove_all' in scope
12 |             index += 1
13 |         }
14 |         gtk_list_box_remove_all(opaquePointer)
   |         `- error: cannot find 'gtk_list_box_remove_all' in scope
15 |     }
16 | 
```

stackotter:
I'm assuming this was resolved, but feel free to open it again if not.

PR #154, opened 2025-05-14
Add wokrflow to test on ubuntu-22.04, rename swift-linux.yml -> ubuntu-24.04

valeriyvan:
Addresses #153

### DIFF
--- a/.github/workflows/swift-linux.yml
+++ b/.github/workflows/swift-linux.yml
@@ -10,7 +10,10 @@ on:
 
 jobs:
   build-linux:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


stackotter:
Given that both workflows are essentially identical, I think it'd be beneficial to instead just parameterise the original workflow using a matrix. Something like so should work:

```yaml
jobs:
  build-linux:
    strategy:
      matrix:
        os: [ubuntu-22.04, ubuntu-latest]
    runs-on: ${{ matrix.os }}
    steps:
      ...
```

stackotter (approved):
Thanks for doing this!

PR #157, opened 2025-05-17
[WinUIBackend] Prevent crash: remove empty paths

bbrk24:
After applying a transform, I appended an empty path to the geometry group, so that future lines and Bézier curves would go in that (untransformed) path. However, if you don't have any lines or curves after the transform, that path stays empty, and WinUI really doesn't like rendering empty paths. This PR adds a cleanup pass to remove any empty paths, and adds a custom curve to the PathsExample that previously would have triggered this crash.


stackotter (approved):
Thanks for fixing this 🙏

PR #168, opened 2025-05-23
Build for Catalyst

bbrk24:
Updates Package.resolved so it actually builds, add a CI job for it, and add a missing `#available`.

### DIFF
--- a/.github/workflows/build-test-and-docs.yml
+++ b/.github/workflows/build-test-and-docs.yml
@@ -127,14 +127,12 @@ jobs:
         xcodebuild-device-type: ${{ matrix.device-type }}
 
   uikit-catalyst:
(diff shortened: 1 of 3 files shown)


bbrk24:
I thought I'd have two such builds, one for x86 and one for ARM. I thought to run the x86 build on macOS 13 so it would be a native build instead of cross-compiling. I forgot macOS 13 can't run a new enough version of Xcode.

stackotter:
That seems a lot like an issue I was getting back when Swift Bundler used SwiftPM for iOS builds. I fixed that issue by using xcodebuild when targeting iOS, so it's weird that the CI job is getting a similar issue but with xcodebuild.

stackotter (approved):
Thanks for getting this working!

PR #175, opened 2025-06-03
Modern concurrency support

bbrk24:


stackotter:
I've tried building a few of my local projects with these changes, and the main things I've bumped into are;
- view initialisers should be `nonisolated` where possible (to allow for things such as storing/producing `Image`s from stored/computed properties)
- button actions should be just `@MainActor` (without `@Sendable`) if possible?

I'll try with a few more projects when I get the time to see if anything else comes up.

bbrk24:
I can try to make `Image`'s initializer nonisolated, but in my personal projects when I need to send an image around, I traffic in `ImageFormats.Image<RGBA>`, not `SwiftCrossUI.Image`.

stackotter (approved):
Sorry for the long delay reviewing this. Thanks for your patience 🙏

I'm happy to merge this now. Looks like there are just a couple of merge conflicts that need fixing now that I merged the TextEditor PR and the dynamic fonts PR; sorry about that.

stackotter (approved):
Looks good!

Issue #176, opened 2025-06-03
Alternatives for SwiftUI's Grid, LazyVGrid, and LazyHGrid


valeriyvan:
Are there any alternatives to for SwiftUI's `Grid`, `LazyVGrid`, and `LazyHGrid`?

stackotter:
I haven't got `LazyVGrid` or `LazyHGrid` alternatives yet, they will get a bit complicated to implement correctly, but a `Grid` isn't too far off being possible.

PR #177, opened 2025-06-03
Add simple implementation of Grid 

valeriyvan:
Partially solves #176


stackotter:
I believe that this implementation of Grid is just equivalent to using the following in today's SwiftCrossUI:

```swift
VStack(alignment: .leading) {
    ForEach(rows) { row in
        HStack {
            ForEach(row) { item in
                Text(item.description)
            }
        }
    }
}
```

I don't want to introduce a `Grid` implementation that differs from the intended eventual behaviour of `Grid` because that would lead to unnecessary breaking changes in the future for little gain.

For that reason I'll close this PR for now, but I'll let you know once I've got a proper grid implementation going!

Issue #194, opened 2025-06-24
Running Examples on iOS Simulator


Vitorsilveira31:
There is currently no information on how to run the examples on iOS. It would be great to add instructions to the README showing how to set this up using tools like mint and swift-bundler to run the examples easily on the iOS Simulator.

Vitorsilveira31:
I’ve created a pull request for this: https://github.com/stackotter/swift-cross-ui/pull/195

PR #195, opened 2025-06-24
#194: Add instructions to run examples on iOS Simulator using mint and swift-bundler

Vitorsilveira31:
This PR adds a new section to the README with step-by-step instructions on how to run the Swift package examples on the iOS Simulator. Since the examples are distributed as Swift packages, they can’t be run directly on iOS without additional setup.

Using tools like [mint](https://github.com/yonaskolb/Mint) and [swift-bundler](https://github.com/stackotter/swift-bundler) simplifies running these examples without needing to create an Xcode project manually. The instructions include installation steps and a sample command to run the CounterExample on the iPhone simulator.

This addition should help new users get started quickly and improve overall developer experience.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,6 @@ To dive right in with SwiftCrossUI, check out [the SwiftCrossUI quick start guid
 
 ## Overview
 
-- [Overview](#overview)
 - [Community](#community)
 - [Supporting SwiftCrossUI](#supporting-swiftcrossui)
 - [Documentation](#documentation)
--- a/README.md
+++ b/README.md
@@ -81,23 +81,47 @@ struct CounterApp: App {
 }
 ```
 
-Clone the SwiftCrossUI repository to test out this example, and many more;
+Clone the SwiftCrossUI repository to test out this example and others. Running the examples requires an installation of [Swift Bundler](https://github.com/stackotter/swift-bundler), which provides consistent behavior across platforms and enables running on iOS and tvOS simulators and devices.
 
-```sh
-git clone https://github.com/stackotter/swift-cross-ui
-cd swift-cross-ui/Examples
-swift run CounterExample
-```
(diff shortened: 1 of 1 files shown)


stackotter:
Thanks for pointing out this gap in the readme. I reckon it might even make sense to just replace the existing `swift run CounterExample` instructions and recommend Swift Bundler for all platforms.

It's probably worth linking to [Swift Bundler's official installation instructions](https://github.com/stackotter/swift-bundler?tab=readme-ov-file#installation-) as well.

We could maybe move the existing SwiftPM-based instructions to a small sentence along the lines of "These examples may be run on your host machine without Swift Bundler by running `swift run TargetName`, but resources may not end up where the examples expect them to be so do not expect everything to work."

---

> Here's what I reckon it could look like

Clone the SwiftCrossUI repository to test out this example, and many more. **Running the examples requires a [Swift Bundler installation](https://github.com/stackotter/swift-bundler?tab=readme-ov-file#installation-)**. *You may be able to run the examples directly with SwiftPM, but your mileage may vary depending on whether the example contains resources or requires features such as deep linking. You also won't be able to run them on iOS/tvOS devices.*

```sh
git clone https://github.com/stackotter/swift-cross-ui
cd swift-cross-ui/Examples
# Run on host machine
swift-bundler run CounterExample
# Run on iOS (or tvOS) device containing 'iPhone' in its name (macOS-only)
swift-bundler run CounterExample --device iPhone
# Run on iOS (or tvOS) simulator containing 'iPhone 16' in its name (macOS-only)
swift-bundler run CounterExample --simulator "iPhone 16"
```

Vitorsilveira31:
@stackotter hey, I've updated it. Do you think it works?

stackotter (approved):
Thanks for this 🙏 There are just a few changes I want to make but I'll merge into my own branch and then do them there

PR #200, opened 2025-06-25
Update view/scene builders to support up to 20 views/scenes (fixes #198)

### LINKED ISSUE #198: Extend ViewBuilder to accept more then 10 parameters
10 parameters isn't that much for desktop interfaces when `Divider` and `Spacer` are used. And when you pass 10 items limit you get so cryptic compile error message that it takes time to understand what's going on. Would be great to be able to build out of at least 20 items.

stackotter:
Just updates SwiftCrossUI's view/scene builders to support up to 20 views/scenes by updating some constants in our gyb files.

I've tested build performance with and without these extra tuple views/scenes and builder methods, and build times don't seem to have been affected.

Unfortunately these changes do make CounterExample's binary 2mb larger in both debug mode and release mode (about 20% bigger for such a simple app). Perhaps I can gate this behind an 'SCUI_EXTENDED_BUILDERS` environment variable for now and then make it a trait once we bump to Swift 6.1?

Before I merge this, I'll also attempt to reduce the size of the generated files by moving repeated logic into generic helper methods, because this PR currently has 10,000 additions, which is a ridiculous amount of repetitive code.


stackotter:
My latest commit reduced repetition in the generated code quite significantly (7500 lines less code) and resulted in 1mb smaller debug binaries and 0.5mb smaller release binaries.

## CounterExample binary size (AppKitBackend + Swift 6.0.3)

### Debug

- 10 builders: 8.7mb
- 20 builders: 11.8mb
- 20 builders (shrunk): 10.7mb

### Release

- 10 builders: 5.7mb
- 20 builders: 7.4mb
- 20 builders (shrunk): 6.9mb

Sources/SwiftCrossUI/Layout/LayoutSystem.swift
valeriyvan, line 1: I don't think changes in this file belong to this PR.

stackotter: I only made the change in this PR because it was required to get one of the size optimisations working. When I moved one of the repeated function bodies into  
a helper method, I needed to make the helper MainActor and then started getting all sorts of concurrency warning about LayoutableChild.

stackotter (commented):
Gonna merge this now, I think a 1mb loss is acceptable, and there's a chance we could improve that in future by introducing an option for people to limit themselves to 10 views/scenes via a compile time flag.

PR #208, opened 2025-07-25
Refactor all Tests from XCTest to Swift Testing

### LINKED ISSUE #196: Switch over to Swift Testing
The current tests are written using XCTest. We should move over to the more modern, and more better, [Swift Testing](https://github.com/swiftlang/swift-testing).

This shouldn't take too long cause there are only 6 tests or so... (whoops...)

MinerMinerMods:
Changes:
- Converted all tests to use `@Test`
- Split all Suites into separate files
- Tagged all existing Tests and Suites.
- Added to `.gitignore` DocC build artifacts and the `.fleet` folder

Additional Notes
- [ ] Does `DefaultBackend` work as a mock backend for the purpose of removing the `AppKitBackend` requirement for the contents of the `Backend` Suite?

Fixes #196


Tests/SwiftCrossUITests/SwiftCrossUIBackendTests.swift
```diff
@@ -0,0 +1,158 @@
...
+// TODO: Create mock backend so that this can be tested on all platforms. There's
+//       nothing AppKit-specific about it.
+@Suite(
+    "Testing for Graphical Backends",
+    .tags(.Backend)
+)
+struct BackendTests {
+    enum BackendTestError: Error {
+        case FailedBitmapBack
+        case FailedTiffRep
```
stackotter: Use lower camel case for enum cases (e.g. `failedBitmapBack` instead of `FailedBitmapBack`).

That said, due to test errors being purely for reporting, I think this error type would be better suited as a struct with a `message` field. Then tests can easily report detailed errors without having to create enum cases and update the switch (which adds a bit of a barrier to changing tests).

```swift
throw BackendTestError(message: "Failed to create backing bitmap")
```

Tests/SwiftCrossUITests/SwiftCrossUIBackendTests.swift
```diff
@@ -0,0 +1,158 @@
...
+            }
+        }
+    }
+
+    #if canImport(AppKitBackend)
+        @Test(
+            "Validates Observation will not give up the thread entirely when sleeping the thread",
+            .tags(.Observation)
+        )
+        func ThrottledStateObservation() async {
```
stackotter: Use lower camel case for function names. E.g. `throttledStateObservation` instead of `ThrottledStateObservation`.

Tests/SwiftCrossUITests/SwiftCrossUIBackendTests.swift
```diff
@@ -0,0 +1,158 @@
...
+
+            // Compute percentage of main thread's time taken up by updates.
+            let ratio = Double(await updateCount.count) * updateDuration / elapsed
+            #expect(
+                ratio <= 0.85,
+                """
+                Expected throttled updates to take under 85% of the main \
+                thread's time. Took \(Int(ratio * 100))%
+
+                Duration: \(elapsed), \(start) <=> \(end)
```
stackotter: Remove the `\(start) <=> \(end)` part. They'll just be big system uptime numbers that don't provide any more information than `\(elapsed)`.

Add `seconds` unit to duration text.

Tests/SwiftCrossUITests/SwiftCrossUIBackendTests.swift
```diff
@@ -0,0 +1,158 @@
...
+            switch self {
+                case .FailedBitmapBack: "Failed to create bitmap backing"
+                case .FailedTiffRep: "Failed to create tiff representation"
+            }
+        }
+    }
+
+    #if canImport(AppKitBackend)
+        @Test(
+            "Validates Observation will not give up the thread entirely when sleeping the thread",
```
stackotter: Replace with `"Ensures that Publisher.observeAsUIUpdater throttles state update observations"`

Tests/SwiftCrossUITests/SwiftCrossUIBackendTests.swift
```diff
@@ -0,0 +1,158 @@
...
+                result.size == ViewSize(fixedSize: SIMD2(92, 96)),
+                """
+                View update result mismatch
+                \(result.size) != \(SIMD2(92, 96))
+                """
+            )
+
+            #expect(
+                result.preferences.onOpenURL == nil,
+                "`onOpenURL` not `nil` it was not set yet"
```
stackotter: Replace with `onOpenURL should be nil as it was not set`

Tests/SwiftCrossUITests/SwiftCrossUIBackendTests.swift
```diff
@@ -0,0 +1,158 @@
...
+            let result = viewGraph.update(
+                proposedSize: SIMD2(200, 200),
+                environment: environment,
+                dryRun: false
+            )
+            let view: AppKitBackend.Widget = viewGraph.rootNode.widget.into()
+            backend.setSize(of: view, to: result.size.size)
+            backend.setSize(ofWindow: window, to: result.size.size)
+
+            // MARK: This Might be a logic error as `result` was proposed a Size of `SIMD2(200, 200)`, the maximum allotted to it's parent window
```
stackotter: Child views decide their own size and may be smaller or larger than the proposed size given to them. This is expected 👍

Tests/SwiftCrossUITests/SwiftCrossUIBackendTests.swift
```diff
@@ -0,0 +1,158 @@
...
+            #expect(
+                result.size == ViewSize(fixedSize: SIMD2(92, 96)),
+                """
+                View update result mismatch
```
stackotter: Remove this message as Swift Testing should print out enough context to infer this.

Tests/SwiftCrossUITests/SwiftCrossUIBackendTests.swift
```diff
@@ -0,0 +1,158 @@
...
+                """
+            )
+
+            #expect(
+                result.preferences.onOpenURL == nil,
+                "`onOpenURL` not `nil` it was not set yet"
+            )
+        }
+
+        /// Unused Test Function
```
stackotter: Replace with `/// Helper function to be used in future tests`

Tests/SwiftCrossUITests/SwiftCrossUIInternalsTests.swift
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    // NOTE: The Test ``CodableNavigationPath`` cannot be condensed into a parameterized test
+    // or a for loop without specifiying allowed Types individually. A Macro might be useful here
+    /// Makes sure that ``NavigationPath`` can retain data validating them with `compareComponents`.
+    @Test(
+        "The Codable NavigationPath can accurately store Codable data",
+        .tags(.NavPath)
+    )
+    func CodableNavigationPath() throws {
```
stackotter: Please match the original test closer. It was succinct and easy to read/maintain.

Tests/SwiftCrossUITests/SwiftCrossUIInternalsTests.swift
```diff
@@ -0,0 +1,180 @@
...
+                Button("Increase") { count += 1 }
+            }.padding()
+        }
+    }
+
+    // NOTE: The Test ``CodableNavigationPath`` cannot be condensed into a parameterized test
+    // or a for loop without specifiying allowed Types individually. A Macro might be useful here
+    /// Makes sure that ``NavigationPath`` can retain data validating them with `compareComponents`.
+    @Test(
+        "The Codable NavigationPath can accurately store Codable data",
```
stackotter: Replace with `Ensures that NavigationPath instances can be round tripped to JSON and back`.

Tests/SwiftCrossUITests/SwiftCrossUIInternalsTests.swift
```diff
@@ -0,0 +1,180 @@
...
+
+@testable import SwiftCrossUI
+
+@Suite(
+    "SwiftCrossUI's Internal Types",
+    .tags(.Internal)
+)
+struct InternalTests {
+
+    struct CounterView: View {
```
stackotter: This is from the backend tests

Tests/SwiftCrossUITests/SwiftCrossUIInternalsTests.swift
```diff
@@ -0,0 +1,180 @@
...
+import Foundation
+import Testing
+
+@testable import SwiftCrossUI
+
+@Suite(
+    "SwiftCrossUI's Internal Types",
+    .tags(.Internal)
+)
+struct InternalTests {
```
stackotter: These aren't internal, they are testing parts of SwiftCrossUI's public interface

Tests/SwiftCrossUITests/SwiftCrossUIInternalsTests.swift
```diff
@@ -0,0 +1,180 @@
...
+        }
+
+        return original == decoded
+    }
+
+    @Test(
+        "Observation Sanity Checks",
+        .tags(.Observation)
+    )
+    func StateObservationSanity() {
```
stackotter: Rename to `observableObjectObservation`

Tests/SwiftCrossUITests/SwiftCrossUIInternalsTests.swift
```diff
@@ -0,0 +1,180 @@
...
+            let decoded = decoded as? T
+        else {
+            return false
+        }
+
+        return original == decoded
+    }
+
+    @Test(
+        "Observation Sanity Checks",
```
stackotter: Replace with `Validates ObservableObject observation behaviour`

Tests/SwiftCrossUITests/TestTagDeclarations.swift
```diff
@@ -0,0 +1,12 @@
+import Testing
+
+extension Tag {
+    // Primary Group Tags
+    @Tag static var Internal: Self
```
stackotter: Make all of these lower camel case

Tests/SwiftCrossUITests/SwiftCrossUIBackendTests.swift
stackotter: This isn't a backend test, it's a state test, it just happens to require a reference to a backend. In this case, DefaultBackend should work, so feel free to replace the use of AppKitBackend if you want (in just this test, not the real backend test).

stackotter (changes requested):
Thanks for helping to transition SwiftCrossUI to Swift Testing! I've some comments requesting various changes.

In addition to the self contained changes I've requested, I would also like the tests to be rearranged into 3 suites, rather than the 2 suites that you've rearranged them into (which aren't a correct categorisation). The suite names I'd use are; `BackendTests`, `StateTests` and `NavigationPathTests`. The basic layout test should go in `BackendTests`, the two state tests should go in `StateTests` and the remaining navigation path codable test should go in `NavigationPathTests`.

---

**Please limit your use of AI in future PRs, and when addressing these comments, as it makes some strange decisions that I don't believe you would've made by yourself (e.g. the code around line 68 of SwiftCrossUIInternalsTests.swift). Submissions made with AI generally take me significantly longer to review, and have an increased chance that I'll have to request changes.**

Tests/SwiftCrossUITests/SwiftCrossUIBackendTests.swift
```diff
@@ -31,7 +31,8 @@ struct BackendTests {
     #if canImport(AppKitBackend)
         @Test(
             "Validates Observation will not give up the thread entirely when sleeping the thread",
-            .tags(.Observation)
+            .tags(.Observation),
+            .bug("https://github.com/stackotter/swift-cross-ui/issues/167", id: 167)
```
stackotter: This issue is completely unrelated to this test

stackotter:
I forgot to add, the tests currently don't compile (due to around 14 compilation errors).

MinerMinerMods:
The Line you referenced was just a byproduct of 3 hours of trying to make that function be completely parameterized. I will just remove it. I will be working on improving clarity before sending in a commit. I can understand that that was really a confusing section.

MinerMinerMods:
> I forgot to add, the tests currently don't compile (due to around 14 compilation errors).

This is mostly because of the conditional compilation on my end disabling the tests inside of the Backend Suite. Removing as much of the conditionally compiled tests should be a future goal to improve these issues

Tests/SwiftCrossUITests/SwiftCrossUIInternalsTests.swift
MinerMinerMods: I will be correcting the tag to `.interface` and will change the collection to `InterfaceTests`

Tests/SwiftCrossUITests/SwiftCrossUIBackendTests.swift
MinerMinerMods: I believe I found an option where you can treat it like a enum error while still using the struct type. It also has a `.new("")` function for explicitly declaring them in functions with an explicit throw type. If you don't like this idea just let me know.

These throw the same value
```swift
throw BackendTestError.new("Failed to create bitmap backing")
throw BackendTestError.FailedBitmapBack
```
Just like this
```swift
throw BackendTestError.new(nil)
throw BackendTestError.unknownIssue
```

Tests/SwiftCrossUITests/SwiftCrossUIBackendTests.swift
MinerMinerMods: It was just slightly confusing because `SIMD(92, 96)` acts like a magic value

Tests/SwiftCrossUITests/SwiftCrossUIInternalsTests.swift
MinerMinerMods: I removed the conditional compilation and the long Comment. Let me know if there are any more changes to make

Tests/SwiftCrossUITests/SwiftCrossUIBackendTests.swift
MinerMinerMods: I have corrected this already [MinerMinerMods/swift-cross-ui 26adf38](https://github.com/MinerMinerMods/swift-cross-ui/commit/26adf38484606c615a54899606d81ea0abb183a3).

Tests/SwiftCrossUITests/SwiftCrossUIInterfaceTests/SwiftCrossUINavPathTests.swift
```diff
@@ -0,0 +1,91 @@
...
+                String.self, Int.self, [Int].self, Double.self,
+            ] as! [any Codable.Type]
+
+        try #require(
+            Values.count == Types.count,
+            """
+            Test `CodableNavigationPath` is Malformed.
+            `Values` and `Types` must have a 1 to 1 match-up
+            """
+        )
```
MinerMinerMods: Here is a potentially more succinct variant of the test setup
```suggestion
        /// Specifies the values and types input into the `NavigationPath`
        let typeValuePairs: [(type: any Codable.Type, value: any Codable)] =
            [
                (String.self, "a"),
                (Int.self, 1),
                ([Int].self, [1, 2, 3]),
                (Double.self, 5.0),
            ] as! [(any Codable.Type, any Codable)]

        let Values: [any Codable] = typeValuePairs.map { $0.value }

        let Types: [any Codable.Type] = typeValuePairs.map { $0.type }
```

Tests/SwiftCrossUITests/SwiftCrossUIInterfaceTests/SwiftCrossUINavPathTests.swift
```diff
@@ -0,0 +1,91 @@
...
+            "An Issue with Navigation path data retainment occured"
+        )
+        #expect(
+            Self.compareComponents(ofType: [Int].self, components[2], decodedComponents[2]),
+            "An Issue with Navigation path data retainment occured"
+        )
+        #expect(
+            Self.compareComponents(ofType: Double.self, components[3], decodedComponents[3]),
+            "An Issue with Navigation path data retainment occured"
+        )
```
MinerMinerMods, line 64: I did consider using a loop but it would have been excessive and overkill as there is only one instance of each type

Tests/SwiftCrossUITests/SwiftCrossUIInternalsTests.swift
MinerMinerMods: I also have another correction [further down](https://github.com/stackotter/swift-cross-ui/pull/208#discussion_r2241525020) if you would like a more concise setup.

stackotter:
I've spent quite a while reviewing this PR across a few separate sessions, and I'm short on time. I've decided to just make the transition myself cause I've run out of time to review, and I needed the changes for something else I want to do (and there are still changes that I'd like you to make). Sorry about that

PR #210, opened 2025-08-06
Bring back the terminal backend to life.

migueldeicaza:
It renames 'CursesBackend' to TermKit, as now TermKit has various console drivers, and curses is just one of them.

The challenge is that TermKit deals with screen sizes like 80x24, not things like 600x400, so both the default sizes and spacing in the codebase will need to compensate for that.

The AppBackend introduces a defaultStackSpacingAmount which is the spacing that VStack and HStack would use, it is currently kept as a default to the existing value, but for the TermKit backend, this value is set to zero.

Now HStack and VStack rather than storing a resolved value like 'spacing: Int', they store the request 'spacing: Int?', which is resolved before it is actually needed by querying the backend for this data.

There is also a new 'limitScreenBounds' that is a no-op for most, but on the curses cases, ensures that a misuse of the defaultSize does not trickle up and down the stack with unrealistic view sizes.


Package.swift
```diff
@@ -252,10 +252,10 @@ let package = Package(
             name: "WinUIInterop",
             dependencies: []
         ),
-        // .target(
-        //     name: "CursesBackend",
-        //     dependencies: ["SwiftCrossUI", "TermKit"]
-        // ),
+         .target(
```
stackotter, line 255: Fix indentation (and for the TermKit package above and TermKitBackend product above that)

Package.swift
```diff
@@ -252,10 +252,10 @@ let package = Package(
...
+             name: "TermKitBackend",
+             dependencies: ["SwiftCrossUI", .product(name: "TermKit", package: "TermKit")]
```
stackotter, line 257: Use `"TermKit"` spelling instead of `.product(name: "TermKit", package: "TermKit")`

Sources/SwiftCrossUI/Backend/AppBackend.swift
```diff
@@ -1139,4 +1145,8 @@ extension AppBackend {
     ) {
         todo()
     }
+
+    public func limitScreenBounds(_ bounds: SIMD2<Int>) -> SIMD2<Int> {
```
stackotter, line 1149: I feel like this should be reworked to `func clampWindowSize(_ size: SIMD2<Int>, for window: Window) -> SIMD2<Int>`. Renamed for clarity, and extra parameter added for future proofing (future purposes could want the window in question to access properties such as scaling factors).

Sources/SwiftCrossUI/Scenes/WindowGroupNode.swift
```diff
@@ -88,8 +88,8 @@ public final class WindowGroupNode<Content: View>: SceneGraphNode {
         _ = update(
             newScene,
             proposedWindowSize: isFirstUpdate && isProgramaticallyResizable
-                ? (newScene ?? scene).defaultSize
-                : backend.size(ofWindow: window),
+            ? backend.limitScreenBounds((newScene ?? scene).defaultSize)
```
stackotter, line 91: Indent these two hanging lines

Sources/TermKitBackend/TermKitBackend.swift
```diff
@@ -0,0 +1,449 @@
...
+    public func setMinimumSize(ofWindow window: RootView, to minimumSize: SIMD2<Int>) {
+        // TODO
+    }
+    
+    public func activate(window: RootView) {
+        fatalError()
+    }
+
+    public func runInMainThread(action: @escaping @MainActor () -> Void) {
+#if os(macOS) || os(iOS) || os(tvOS) || os(watchOS) || os(visionOS)
```
stackotter, line 66: Indent this body two levels

stackotter (changes requested):
Just had a quick look out of interest and left a few comments (mostly formatting related). I'll change this PR to a draft for now

PR #215, opened 2025-09-21
[Feat] range foreach

MiaKoring:
https://github.com/stackotter/swift-cross-ui/issues/213
Just adds ForEach Initializers accepting Range<Int> or ClosedRange<Int> and updated HoverExample to use the new initializer instead of the old array hack.

I expect it to work seemlessly across all supported platforms, as its just 2 new initializers


stackotter (approved):
Thanks for this 🙏 I’ve found the `Array(range)` workaround annoying often enough that I do think it’s worth having these concenience initialisers

PR #218, opened 2025-10-02
Add compatibility for legacy iOS

JWIMaster:
Please do not merge unless you are happy to remove everything I have done. I advise you review my changes, most do not impact your framework whatsover. I have preserved Concurrency basically everywhere, and provided fallback functions for older iOS versions. Let me know of anything you want me to do. Once everything is complete, you can merge. I will change the Package.swift to something better if you like, but currently, the way I have it is the only way it compiles.


JWIMaster:
Also, I added my UIKitCompatKit which I am developing partially for this, which does a lot of the heavy lifting in the backend,

JWIMaster:
Update: It now builds, but crashes. I am not entirely sure why, because I don't know the details of this repo very well. I am assuming it is one of the classes I marked as iOS 13+, if one of you could let me know which it is that would cause crashes that'd be great, so I can work on either bypassing or shimming it.

JWIMaster:
Removed a bunch of commits related to iOS 7 compatibility, however, i have now patched the repo to seemingly be fully working on iOS 10+. Of course, all iOS 10+ methods are disabled because the target is iOS 7, but, this is a decent proof of concept that this can work. I have reached the end of what I can reasonably do without help, as I do not really understand the inner workings of this repo. I have hopes that this repo can be able to run on iOS 6+, but as it stands, iOS 10+ should 100% be considered as a reasonable target.

timi2506:
WOW! this would be GREAT!

Sources/SwiftCrossUI/SwiftCrossUI.docc/AppKitBackend.md
```diff
@@ -11,15 +11,15 @@ SwiftCrossUI's native macOS backend built on top of AppKit.
 ```swift
 ...
 
-let package = Package(
+let public = Package(
```
stackotter, line 14: You seem to have found and replaced all occurrences of `package` with `public` 😅

stackotter (changes requested):
Hey, sorry for letting this sit here for so long. It's a really cool piece of work, and although I don't think it should live on the main branch of the repository for now, I'd like to merge it into a dedicated `ios7_support` branch (which future contributors can rebase and work in if they desire iOS 7 support). I've only skim read the code as it's not going into the main branch (mostly just doing my due diligence). I've eyeballed everything but haven't checked for logic bugs.

Sources/SwiftCrossUI/SwiftCrossUI.docc/AppKitBackend.md
JWIMaster: oh weird, have you tested building with this change? I assume it doesn't break anything as it's just a mass rename

stackotter: I haven't, but that change is in documentation so it wouldn't get caught by building the project anyway. That snippet shows people how to set up their Package.swift file to use AppKitBackend, and in Package.swift files I'm pretty sure that the variable holding your package description has to be called `package`. Even if it doesn't *have* to be called `package`, it should be called `package`.

PR #219, opened 2025-10-09
Sheet Modifier

MiaKoring:
Is it already Christmas? No but I brought you a gift anyway.

Jokes aside.

This PR adds the .sheet Modifier, like we know it from SwiftUI with isPresented Binding and optional onDismiss callback.

Also some sheet modifiers:
presentationBackground
presentationDragIndicatorVisibility
presentationCornerRadius
presentationDetents
interactiveDismissDisabled

and Example usage in WindowingExample

support for the modifiers varies quite a bit. most are ignored if unsupported, interactiveDismissDisabled and presentationBackground are required as both shouldn’t be limited by Framework features.

Currently only UIKitBackend, AppKitBackend and GtkBackend support sheets, as I sadly couldn’t get them to work properly on WinUI. I hope someone else can build on the Foundation.

As was to expect, there are some weird quirks:
on UIKitBackend and AppKitBackend sheet content is placed top-leading. On GtkBackend on Linux Text elements are very small without further modification instead of allocation space. Same thing happened on macOS with GtkBackend in the 2nd and 3rd window (before and after my changes).

I expect those to be fixable through SCUI Frontend Code. If you know how to fix those in the Backends, I’d be happy to do so before a merge.

### DIFF
--- a/Sources/SwiftCrossUI/Views/Modifiers/SheetModifier.swift
+++ b/Sources/SwiftCrossUI/Views/Modifiers/SheetModifier.swift
@@ -1,7 +1,7 @@
 extension View {
     /// Presents a conditional modal overlay. `onDismiss` gets invoked when the sheet is dismissed.
     ///
-    /// Internal modalPresentationStyle falls back to .overFullScreen (non-opaque) on tvOS 18 and earlier.
+    /// Internal UIViewController.modalPresentationStyle falls back to .overFullScreen (non-opaque) on tvOS.
     public func sheet<SheetContent: View>(
         isPresented: Binding<Bool>, onDismiss: (() -> Void)? = nil,
         @ViewBuilder content: @escaping () -> SheetContent
--- a/Sources/UIKitBackend/UIKitBackend+Sheet.swift
+++ b/Sources/UIKitBackend/UIKitBackend+Sheet.swift
@@ -9,11 +9,7 @@ extension UIKitBackend {
         #if !os(tvOS)
             sheet.modalPresentationStyle = .formSheet
         #else
-            if #available(tvOS 26.0, *) {
-                sheet.modalPresentationStyle = .formSheet
-            } else {
-                sheet.modalPresentationStyle = .overFullScreen
-            }
+            sheet.modalPresentationStyle = .overFullScreen
         #endif
         sheet.view = content.view
         return sheet


Sources/SwiftCrossUI/Backend/AppBackend.swift
```diff
@@ -720,13 +806,25 @@ extension AppBackend {
     }
 }
 
+public protocol SheetImplementation {
+    var sheetSize: SIMD2<Int> { get }
+}
```
stackotter: Using a protocol here is nice, but I've generally tried to keep the backend API really consistent in its design. Unfortunately we can't use protocols for a lot of the widget-related stuff so I've gone with the decision of making all backend data types opaque and putting their getters and setters on the backend. It's not super Swifty, but the goal is to make backend implementation surprise-free, so it's kinda the best we can do with today's Swift.

tl;dr remove the protocol and move `sheetSize` to a `size(ofSheet:)` method on the  AppBackend protocol.

If there's a technical reason that it has to be done this way (e.g. to access sheet sizes in a place with no backend access or something), let me know

Sources/SwiftCrossUI/Backend/AppBackend.swift
```diff
@@ -603,6 +604,91 @@ public protocol AppBackend: Sendable {
     /// ``showAlert(_:window:responseHandler:)``.
     func dismissAlert(_ alert: Alert, window: Window?)
 
+    /// Creates a sheet object (without showing it yet). Sheets contain View Content.
+    /// They optionally execute provied code on dismiss and
```
stackotter: Make `View Content` lowercase (unless it links to a type).

Replace `provied` with `provided`.

Sources/SwiftCrossUI/Backend/AppBackend.swift
```diff
@@ -603,6 +604,91 @@ public protocol AppBackend: Sendable {
...
+    /// prevent users from interacting with the parent window until dimissed.
+    func createSheet() -> Sheet
+
+    /// Updates the content and appearance of a sheet
```
stackotter: End with a full stop (sorry, a bit pedantic cause it's for the public docs)

Sources/SwiftCrossUI/Backend/AppBackend.swift
```diff
@@ -603,6 +604,91 @@ public protocol AppBackend: Sendable {
...
+    func updateSheet(
+        _ sheet: Sheet,
+        content: Widget,
```
stackotter: For views with a single child we generally take the widget in the creation method instead of the update method. Each view graph node is guaranteed to keep the same widget for its whole lifecycle so you can give the sheet creation method a reference the the sheet content's underlying widget before updating the widget with the sheets size (so there shouldn't be any circular dependencies).

Sources/SwiftCrossUI/Backend/AppBackend.swift
```diff
@@ -603,6 +604,91 @@ public protocol AppBackend: Sendable {
...
+        onDismiss: @escaping () -> Void
+    )
+
+    /// Shows a sheet as a modal on top of or within the given window.
+    /// Users should be unable to interact with the parent window until the
+    /// sheet gets dismissed. The sheet will be closed once onDismiss gets called
```
stackotter:
```md
... `onDismiss` only gets called once the sheet has been closed.
```

Sources/SwiftCrossUI/Backend/AppBackend.swift
```diff
@@ -603,6 +604,91 @@ public protocol AppBackend: Sendable {
...
+    ///
+    /// If `window` is `nil`, the backend can either make the sheet a whole
+    /// app modal, a standalone window, or a modal for a window of its choosing.
+    func showSheet(
+        _ sheet: Sheet,
+        window: Window?
+    )
+
+    /// Dismisses a sheet programmatically.
+    /// Gets used by the SCUI sheet implementation to close a sheet.
```
stackotter: `... by the ``View/sheet`` modifier ...`

Sources/SwiftCrossUI/Backend/AppBackend.swift
```diff
@@ -603,6 +604,91 @@ public protocol AppBackend: Sendable {
...
+    func dismissSheet(_ sheet: Sheet, window: Window?)
+
+    /// Sets the corner radius for a sheet presentation.
+    ///
+    /// This method is called when the sheet content has a `presentationCornerRadius` modifier
```
stackotter: Doesn't have to be at the top level, just has to have the preference key set. (same for the other presentation methods down below)

Sources/SwiftCrossUI/Backend/AppBackend.swift
```diff
@@ -603,6 +604,91 @@ public protocol AppBackend: Sendable {
...
+    ///   - detents: An array of detents that the sheet can be resized to.
+    func setPresentationDetents(of sheet: Sheet, to detents: [PresentationDetent])
+
+    /// Sets the visibility for a sheet presentation.
+    ///
+    /// This method is called when the sheet content has a `presentationDragIndicatorVisibility`
+    /// modifier applied at its top level.
+    ///
+    /// - Parameters:
+    ///   - sheet: The sheet to apply the detents to.
```
stackotter: Drag indicator visibility (not detents). Easy mistake hahah

Sources/SwiftCrossUI/Backend/AppBackend.swift
```diff
@@ -603,6 +604,91 @@ public protocol AppBackend: Sendable {
...
+        to visibility: PresentationDragIndicatorVisibility
+    )
+
+    /// Sets the background color for a sheet presentation.
+    ///
+    /// This method is called when the sheet content has a `presentationBackground`
+    /// modifier applied at its top level.
+    ///
+    /// - Parameters:
+    ///   - sheet: The sheet to apply the detents to.
```
stackotter: Same as above (says detents instead of background)

Sources/SwiftCrossUI/Backend/AppBackend.swift
```diff
@@ -603,6 +604,91 @@ public protocol AppBackend: Sendable {
...
+    ///   - color: rgba background color
```
stackotter: `Background color for the sheet` (I avoid mentioning rgba cause Color won't necessarily always be simple rgba, it will eventually support dynamic colors as well)

Sources/SwiftCrossUI/Backend/AppBackend.swift
```diff
@@ -603,6 +604,91 @@ public protocol AppBackend: Sendable {
...
+    /// Sets the interactive dismissablility of a sheet.
+    /// when disabled the sheet can only be closed programmatically,
+    /// not through users swiping, escape keys or similar.
+    ///
+    /// This method is called when the sheet content has a `interactiveDismissDisabled`
+    /// modifier applied at its top level.
+    ///
+    /// - Parameters:
+    ///   - sheet: The sheet to apply the detents to.
+    ///   - disabled: wether its disabled
```
stackotter: `The sheet to apply the interactive dismissibility to.`

`Whether interactive dismissing is disabled.`

Sources/SwiftCrossUI/Backend/AppBackend.swift
```diff
@@ -603,6 +604,91 @@ public protocol AppBackend: Sendable {
...
+    ///
+    /// This method is called when the sheet content has a `presentationBackground`
+    /// modifier applied at its top level.
+    ///
+    /// - Parameters:
+    ///   - sheet: The sheet to apply the detents to.
+    ///   - color: rgba background color
+    func setPresentationBackground(of sheet: Sheet, to color: Color)
+
+    /// Sets the interactive dismissablility of a sheet.
```
stackotter: `dismissibility`

Examples/Sources/WindowingExample/WindowingApp.swift
```diff
@@ -64,6 +64,103 @@ struct AlertDemo: View {
     }
 }
 
+// kind of a stress test for the dismiss action
```
stackotter:
```swift
/// A demo displaying SwiftCrossUI's `View.sheet` modifier.
```

Sources/AppKitBackend/AppKitBackend.swift
```diff
@@ -1685,6 +1686,113 @@ public final class AppKitBackend: AppBackend {
         let request = URLRequest(url: url)
         webView.load(request)
     }
+
+    public func createSheet() -> NSCustomSheet {
+        // Initialize with a default contentRect, similar to window creation (lines 58-68)
```
stackotter: Refer to method name instead of lines (lines may change)

Sources/AppKitBackend/AppKitBackend.swift
```diff
@@ -1685,6 +1686,113 @@ public final class AppKitBackend: AppBackend {
...
+            backing: .buffered,
+            defer: true
+        )
+        return sheet
+    }
+
+    public func updateSheet(
+        _ sheet: NSCustomSheet, content: NSView, onDismiss: @escaping () -> Void
+    ) {
+        let contentSize = naturalSize(of: content)
```
stackotter: Content size should be passed in from SwiftCrossUI because even if this works for AppKitBackend, not all backends will be able to infer the sizing that SwiftCrossUI has assigned to their sheet content.

This may resolve the Gtk issues you were facing.

Sources/AppKitBackend/AppKitBackend.swift
```diff
@@ -1685,6 +1686,113 @@ public final class AppKitBackend: AppBackend {
...
+    }
+
+    public func showSheet(_ sheet: NSCustomSheet, window: NSCustomWindow?) {
+        guard let window else {
+            print("warning: Cannot show sheet without a parent window")
+            return
+        }
+        // critical sheets stack
+        // beginSheet only shows a nested
+        // sheet after its parent gets dismissed
```
stackotter: `// Critical sheets stack. beginSheet only shows a nested sheet after its parent gets dismissed` (wrapped at around 80 chars if necessary

Sources/AppKitBackend/AppKitBackend.swift
```diff
@@ -1685,6 +1686,113 @@ public final class AppKitBackend: AppBackend {
...
+        window.beginCriticalSheet(sheet)
+    }
+
+    public func dismissSheet(_ sheet: NSCustomSheet, window: NSCustomWindow?) {
+        if let window {
+            window.endSheet(sheet)
+        } else {
+            NSApplication.shared.stopModal()
```
stackotter: Is this necessary if showSheet never shows sheets without parent windows?

Sources/SwiftCrossUI/Backend/AppBackend.swift
```diff
@@ -603,6 +604,91 @@ public protocol AppBackend: Sendable {
...
+    /// Users should be unable to interact with the parent window until the
+    /// sheet gets dismissed. The sheet will be closed once onDismiss gets called
+    ///
+    /// Must only get called once for any given sheet.
+    ///
+    /// If `window` is `nil`, the backend can either make the sheet a whole
+    /// app modal, a standalone window, or a modal for a window of its choosing.
+    func showSheet(
+        _ sheet: Sheet,
+        window: Window?
```
stackotter: Do sheets ever get shown without a parent window?

Sources/AppKitBackend/AppKitBackend.swift
```diff
@@ -1685,6 +1686,113 @@ public final class AppKitBackend: AppBackend {
...
+        }
+    }
+
+    public func setPresentationBackground(of sheet: NSCustomSheet, to color: Color) {
+        let backgroundView = NSView()
+        backgroundView.wantsLayer = true
+        backgroundView.layer?.backgroundColor = color.nsColor.cgColor
+
```
stackotter, line 1757: Does it work to set the [`backgroundColor`](https://developer.apple.com/documentation/appkit/nswindow/backgroundcolor) of the window instead?

Sources/GtkBackend/GtkBackend.swift
```diff
@@ -48,6 +50,59 @@ public final class GtkBackend: AppBackend {
...
+            // TRUE (1) = consume event
+            if keyval == GDK_KEY_Escape {
+                guard let userData else { return 1 }
+                let box = Unmanaged<ValueBox<() -> Void>>.fromOpaque(userData).takeUnretainedValue()
+                box.value()
+                return 1
+            }
+            return 0
+        }
+
```
stackotter: Update the `Gtk.Window` implementation in `Sources/Gtk` instead so that everyone can make use of a nice wrapper for these signals. See `Sources/Gtk/Generated/Button.swift::Button.didMoveToParent` for an example of how these signal handler wrappers are implemented.

Sources/GtkBackend/GtkBackend.swift
```diff
@@ -1569,6 +1624,101 @@ public final class GtkBackend: AppBackend {
...
+    public func createSheet() -> Gtk.Window {
+        return Gtk.Window()
+    }
+
+    public func updateSheet(_ sheet: Gtk.Window, content: Widget, onDismiss: @escaping () -> Void) {
+        sheet.setChild(content)
+
+        let key: OpaquePointer = OpaquePointer(sheet.widgetPointer)
+
+        //add a slight border to not be just a flat corner
```
stackotter: `// Add ...`

Sources/SwiftCrossUI/Values/PresentationDetent.swift
```diff
@@ -0,0 +1,18 @@
+/// Represents the available detents (heights) for a sheet presentation.
+public enum PresentationDetent: Sendable, Hashable {
+    /// A detent that represents a medium height sheet.
+    case medium
+
+    /// A detent that represents a large (full-height) sheet.
+    case large
+
+    /// A detent at a custom fractional height of the available space.
+    /// falling back to medium on iOS 15
```
stackotter: `Falls back to medium on iOS 15`

Sources/SwiftCrossUI/Values/PresentationDetent.swift
```diff
@@ -0,0 +1,18 @@
...
+    /// - Parameter fraction: A value between 0 and 1 representing the fraction of available height.
+    case fraction(Double)
+
+    /// A detent at a specific fixed height in pixels.
```
stackotter: This is probably `points` rather than `pixels` on most backends

Sources/SwiftCrossUI/Values/PresentationDragIndicatorVisibility.swift
```diff
@@ -0,0 +1,3 @@
+public enum PresentationDragIndicatorVisibility: Sendable {
```
stackotter: Apple seems to use [a generic `Visibility` enum](https://developer.apple.com/documentation/swiftui/visibility) for this situation. Probably makes sense to copy them there, cause I can see such an enum being useful in other contexts as well. We should have an `automatic` case as well (which would be whatever the backend's default is).

Sources/SwiftCrossUI/ViewGraph/PreferenceValues.swift
```diff
@@ -2,13 +2,44 @@ import Foundation
...
     public var onOpenURL: (@Sendable @MainActor (URL) -> Void)?
 
-    public init(onOpenURL: (@Sendable @MainActor (URL) -> Void)?) {
+    /// The available detents for a sheet presentation. Only applies to the top-level view in a sheet.
+    public var presentationDetents: [PresentationDetent]?
+
+    /// The corner radius for a sheet presentation. Only applies to the top-level view in a sheet.
+    public var presentationCornerRadius: Double?
+
+    /// The drag indicator visibility for a sheet presentation. Only applies to the top-level view in a sheet.
```
stackotter: `. Applies to enclosing sheets.` (same for the other related properties)

Sources/SwiftCrossUI/ViewGraph/PreferenceValues.swift
```diff
@@ -2,13 +2,44 @@ import Foundation
...
+    public var presentationDragIndicatorVisibility: PresentationDragIndicatorVisibility?
+
+    /// The backgroundcolor of a sheet. Only applies to the top-level view in a sheet
+    public var presentationBackground: Color?
+
+    public var interactiveDismissDisabled: Bool?
```
stackotter, line 28: `/// Controls whether the user can interactively dismiss enclosing sheets.`

Sources/SwiftCrossUI/ViewGraph/PreferenceValues.swift
```diff
@@ -21,5 +52,13 @@ public struct PreferenceValues: Sendable {
                 }
             }
         }
+
+        // For presentation modifiers, take the first (top-level) value only
```
stackotter: `// For presentation modifiers, take the outer-most value (using child ordering to break ties).`

Sources/SwiftCrossUI/ViewGraph/PreferenceValues.swift
```diff
@@ -21,5 +52,13 @@ public struct PreferenceValues: Sendable {
...
+        // This ensures only the root view's presentation modifiers apply to the sheet
+        presentationDetents = children.first?.presentationDetents
+        presentationCornerRadius = children.first?.presentationCornerRadius
+        presentationDragIndicatorVisibility = children.first?.presentationDragIndicatorVisibility
+        presentationBackground = children.first?.presentationBackground
+        interactiveDismissDisabled = children.first?.interactiveDismissDisabled
```
stackotter: Perform a merge over all of the children and take the first defined value when multiple children define a value for a given property. Similar to `navigationTitle`, I think these preferences should flow up the view hierarchy freely so that people can abstract out their presentation preferences into other views easily. E.g. adding spacers above and below the sheet's content should silently override the presentation modifiers set on the sandwiched content.

Something like `children.map(\.presentationDetents).first { $0 != nil }` should work well?

Sources/SwiftCrossUI/Views/Modifiers/PresentationModifiers.swift
```diff
@@ -0,0 +1,67 @@
+extension View {
+    /// Sets the available detents (heights) for a sheet presentation.
+    ///
+    /// This modifier only affects the sheet presentation itself when applied to the
+    /// top-level view within a sheet. It allows users to resize the sheet to different
+    /// predefined heights.
+    ///
+    /// supported platforms: iOS (ignored on unsupported platforms)
+    /// ignored on: older than iOS 15
```
stackotter:
```
- Supported platforms: iOS 15+ (ignored on unsupported platforms)
  - `.fraction` and `.height` fall back to `.medium` on iOS 15 and earlier
```

Apply similar formatting to the support notes for the other modifiers as well.

Sources/SwiftCrossUI/Views/Modifiers/PresentationModifiers.swift
```diff
@@ -0,0 +1,67 @@
+extension View {
+    /// Sets the available detents (heights) for a sheet presentation.
+    ///
+    /// This modifier only affects the sheet presentation itself when applied to the
+    /// top-level view within a sheet. It allows users to resize the sheet to different
+    /// predefined heights.
```
stackotter: Remove the part about top-level views (not true anymore with my suggested change to the preference value merging implementation).

Sources/SwiftCrossUI/Views/Modifiers/PresentationModifiers.swift
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    /// Sets the corner radius for a sheet presentation.
+    ///
+    /// This modifier only affects the sheet presentation itself when applied to the
+    /// top-level view within a sheet. It does not affect the content's corner radius.
+    ///
+    /// supported platforms: iOS 15+, Gtk4 (ignored on unsupported platforms)
+    ///
+    /// - Parameter radius: The corner radius in pixels.
```
stackotter: Probably `points` not `pixels` on most backends

Sources/SwiftCrossUI/Views/Modifiers/PresentationModifiers.swift
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    /// Sets the visibility of a sheet's drag indicator.
+    ///
+    /// This modifier only affects the sheet presentation itself when applied to the
+    /// top-level view within a sheet.
+    ///
+    /// supported platforms: iOS 15+ (ignored on unsupported platforms)
+    ///
+    /// - Parameter visibiliy: visible or hidden
```
stackotter: `visibility` (typo)

Sources/SwiftCrossUI/Views/Modifiers/PresentationModifiers.swift
```diff
@@ -0,0 +1,67 @@
...
+        preference(key: \.presentationDragIndicatorVisibility, value: visibility)
+    }
+
+    /// Sets the background of a sheet.
+    ///
+    /// This modifier only affects the sheet presentation itself when applied to the
+    /// top-level view within a sheet.
+    ///
+    /// - Parameter color: the background color
+    /// - Returns: A view with the presentation corner radius preference set.
```
stackotter: Background not corner radius

Sources/SwiftCrossUI/Views/Modifiers/PresentationModifiers.swift
```diff
@@ -0,0 +1,67 @@
...
+    public func presentationBackground(_ color: Color) -> some View {
+        preference(key: \.presentationBackground, value: color)
+    }
+
+    /// Sets wether the user should be able to dismiss the sheet themself.
```
stackotter: `whether`

Sources/SwiftCrossUI/Views/Modifiers/PresentationModifiers.swift
```diff
@@ -0,0 +1,67 @@
...
+    ///
+    /// This modifier only affects the sheet presentation itself when applied to the
+    /// top-level view within a sheet.
+    ///
+    /// - Parameter isDisabled: is it disabled
```
stackotter: `Whether interactive dismissal is disabled`

Sources/SwiftCrossUI/Views/Modifiers/PresentationModifiers.swift
```diff
@@ -0,0 +1,67 @@
...
+    /// - Returns: A view with the presentation corner radius preference set.
```
stackotter: Interactive dismiss not corner radius

Sources/UIKitBackend/UIKitBackend+Sheet.swift
```diff
@@ -0,0 +1,151 @@
...
+        if #available(iOS 15.0, *) {
+            #if !os(visionOS)
+                if let sheetController = sheet.sheetPresentationController {
+                    sheetController.prefersGrabberVisible = visibility == .visible ? true : false
+                }
+            #endif
+        } else {
+            #if DEBUG
+                print(
+                    "Your current OS Version doesn't support setting sheet drag indicator visibility.\n Setting this is only available from iOS 15.0"
```
stackotter: Remove the `\n`, best to keep it together I think cause it's not too long (same for above).

Sources/WinUIBackend/WinUIBackend.swift
```diff
@@ -1879,6 +1880,12 @@ public class CustomWindow: WinUI.Window {
     var grid: WinUI.Grid
     var cachedAppWindow: WinAppSDK.AppWindow!
 
+    //only for AppBackend conformance, no support yet
```
stackotter: `// Only ...`

Sources/SwiftCrossUI/Views/Modifiers/SheetModifier.swift
```diff
@@ -0,0 +1,170 @@
+extension View {
+    /// Presents a conditional modal overlay
+    /// onDismiss optional handler gets executed before
+    /// dismissing the sheet
```
stackotter:
```swift
/// Presents a conditional modal overlay. `onDismiss` gets invoked when the sheet is dismissed.
```

Sources/SwiftCrossUI/Views/Modifiers/SheetModifier.swift
```diff
@@ -0,0 +1,170 @@
...
+    public func sheet<SheetContent: View>(
+        isPresented: Binding<Bool>, onDismiss: (() -> Void)? = nil,
+        @ViewBuilder content: @escaping () -> SheetContent
+    ) -> some View {
+        SheetModifier(
+            isPresented: isPresented, body: TupleView1(self), onDismiss: onDismiss,
```
stackotter: Wrap each parameter onto its own line, and the closing `)` onto its own as well (only cause it's wrapped);

```
foo(
    param1,
    param2,
    ...
)
```

Sources/SwiftCrossUI/Views/Modifiers/SheetModifier.swift
```diff
@@ -0,0 +1,170 @@
...
+            environment: environment
+        )
+        let bodyNode = AnyViewGraphNode(bodyViewGraphNode)
+
+        let sheetViewGraphNode = ViewGraphNode(
+            for: sheetContent(),
+            backend: backend,
+            environment: environment
+        )
+        let sheetContentNode = AnyViewGraphNode(sheetViewGraphNode)
```
stackotter: Delay sheet content display until the sheet is presented (so that `onAppear` and `onDisappear` work correctly). You'll also have to set the sheet node to nil when the sheet is dismissed (so that the sheet node gets dropped and `onDisappear` gets triggered).

Sources/SwiftCrossUI/Views/Modifiers/SheetModifier.swift
```diff
@@ -0,0 +1,170 @@
...
+            let sheetEnvironment = environment.with(\.dismiss, dismissAction)
+
+            let dryRunResult = children.sheetContentNode.update(
+                with: sheetContent(),
+                proposedSize: sheet.sheetSize,
+                environment: sheetEnvironment,
+                dryRun: true
+            )
+
+            let preferences = dryRunResult.preferences
```
stackotter: Even though these should be the same as the final run, it's best to use the preferences from the final non-dryrun to avoid confusion if things go wrong.

stackotter (changes requested):
I love how much of the SwiftUI sheets API you've managed to implement in this PR! Sorry about all the requested changes; they're generally about the internals rather than the API surface. You've had to use a lot of SwiftCrossUI's more obscure internals like the preferences system and the Gtk bindings signal system, so it's completely understandable 😅 I'll test this out on my own devices locally when I get the time (probably tomorrow).

Sources/AppKitBackend/AppKitBackend.swift
MiaKoring: Where do I get the correct size from?

Sources/AppKitBackend/AppKitBackend.swift
MiaKoring: Not sure why, but it doesn't

Sources/GtkBackend/GtkBackend.swift
MiaKoring: I did it, but aint sure if its a nice solution. close request, sure, but the escape key handler is a little bit less straightforward because it uses the event handler. If you don’t like how its done now please show me how it should be done and I’ll replace it asap.

At least it seems to be working :)

Sources/SwiftCrossUI/Views/Modifiers/SheetModifier.swift
MiaKoring: onAppear now works correctly. onDisappear only on interactive dismiss. I’m not sure why, the code setting them to nil gets called. a completion handler on programmatic dismiss (i.e. UIViewController/dismiss) also didn’t lead to onDisappear getting called. iirc SwiftUI has a similar issue. Although its not too bad, since programmatic dismissal like the name implies only gets called through code by the developer. This behaviour is consistent across platforms.

MiaKoring: https://stackoverflow.com/questions/60468040/ondisappear-not-called-when-a-modal-view-is-dismissed 

in my opinion we can leave it like that. there are enough possibilities to run code both on interactive and programmatic dismissal, its not really an issue imo. possibly we could document it (if you think its necessary). 

If you know a fix of course we can change it too.

Sources/AppKitBackend/AppKitBackend.swift
stackotter: Just before you call `updateSheet` in your sheet modifier implementation, you compute `result`. You should be able to pass the size from that result (`result.size.size`) into `updateSheet` in order to use that size in backend implementations.

stackotter:
I'm getting package resolution errors when I try to build this PR locally (due to an issue fixed by #224). I believe that merging in the latest changes from main should fix that issue.

Sources/SwiftCrossUI/Backend/AppBackend.swift
```diff
@@ -603,6 +604,94 @@ public protocol AppBackend: Sendable {
...
+    )
+
+    /// Dismisses a sheet programmatically.
+    /// Gets used by the ``View/sheet`` modifier to close a sheet.
+    func dismissSheet(_ sheet: Sheet, window: Window)
+
+    /// Get the dimensions of a sheet
+    func sizeOf(_ sheet: Sheet) -> SIMD2<Int>
```
stackotter: Make this `func size(ofSheet sheet: Sheet) -> SIMD2<Int>` to match naming convention of other similar methods.

Sources/SwiftCrossUI/Backend/AppBackend.swift
```diff
@@ -603,6 +604,94 @@ public protocol AppBackend: Sendable {
...
+
+    /// Sets the corner radius for a sheet presentation.
+    ///
+    /// This method is called when the sheet content has a `presentationCornerRadius`
```
stackotter: `has the`

Sources/SwiftCrossUI/Backend/AppBackend.swift
stackotter: I see you updated `dismissSheet` to take a non-optional parent window. `showSheet`'s window parameter should be non-optional too

Sources/AppKitBackend/AppKitBackend.swift
```diff
@@ -1685,6 +1686,105 @@ public final class AppKitBackend: AppBackend {
...
+            existingContentView.translatesAutoresizingMaskIntoConstraints = false
+            existingContentView.leadingAnchor.constraint(equalTo: container.leadingAnchor)
+                .isActive = true
+            existingContentView.topAnchor.constraint(equalTo: container.topAnchor).isActive = true
+            existingContentView.trailingAnchor.constraint(equalTo: container.trailingAnchor)
+                .isActive = true
+            existingContentView.bottomAnchor.constraint(equalTo: container.bottomAnchor).isActive =
+                true
+
+            sheet.contentView = container
```
stackotter, line 1783: I feel like this function should be able to be called twice (e.g. if we start respecting live updates to the preference values of the sheet's content). In its current form, that would mean that we'd repeatedly add levels of NSView containers one on top of another each time the background colour changes (or even worse, every time that it could possibly have changed).

For that reason, I think AppKitBackend should construct this nested background+content container thing as its `contentView` in `createSheet`. You can store a reference to that actual use contentView in a new property on `NSCustomSheet`. That should make this update-background method relatively trivial.

Sources/Gtk/Widgets/Window.swift
```diff
@@ -82,5 +82,77 @@ open class Window: Widget {
...
+                if let data {
+                    Unmanaged<ValueBox<() -> Void>>.fromOpaque(data).release()
+                }
+            },
+            .init(0)
+        )
+
+        gtk_widget_add_controller(widgetPointer, keyEventController)
+        escapeKeyEventController = keyEventController
+        escapeKeyHandlerData = boxedHandler
```
stackotter: You can avoid the horrible C interop stuff by generating a wrapper for `EventControllerKey` (which should generate the signal handler property) and then using `Widget.addEventController`. Check out GtkBackend's implementations of `createTapGestureTarget` and `updateTapGestureTarget` for examples of using some generated wrappers for event controller subclasses (`GestureClick`).

You can generate the `EventControllerKey` by adding it to `allowListedClasses` in GtkCodeGen.swift and running `./generate.sh && ./format_and_lint.sh` (I believe you've done that before but I figured I should say it again just in case).

Sources/GtkBackend/GtkBackend.swift
```diff
@@ -1569,6 +1585,96 @@ public final class GtkBackend: AppBackend {
 
         return properties
     }
+    public func createSheet(content: Widget) -> Gtk.Window {
```
stackotter, line 1589: Add newline between this function and the one above

Sources/SwiftCrossUI/Values/PresentationDetent.swift
stackotter: Resolved for the one below but not this one. Please add a full stop at the end of the `Falls back to ...` sentence as well (in both).

Sources/SwiftCrossUI/ViewGraph/PreferenceValues.swift
```diff
@@ -2,13 +2,45 @@ import Foundation
...
+    /// The corner radius for a sheet presentation. Applies to enclosing sheets.
+    public var presentationCornerRadius: Double?
+
+    /// The drag indicator visibility for a sheet presentation. Applies to enclosing sheets.
+    public var presentationDragIndicatorVisibility: Visibility?
+
+    /// The backgroundcolor of a sheet. Applies to enclosing sheets.
+    public var presentationBackground: Color?
+
+    /// Controls whether the user can interactively dismiss enclosing sheets.  Applies to enclosing sheets.
```
stackotter, line 27: Don't need the `Applies to enclosing sheets` here due to the wording of the previous sentence.

Sources/SwiftCrossUI/Views/Modifiers/PresentationModifiers.swift
```diff
@@ -0,0 +1,62 @@
...
+    /// Sets the available detents (heights) for a sheet presentation.
+    ///
+    /// This modifier only affects the sheet presentation itself.
+    /// It allows users to resize the sheet to different predefined heights.
+    ///
+    /// - Supported platforms: iOS 15+ (ignored on unsupported platforms)
+    /// - `.fraction` and `.height` fall back to `.medium` on iOS 15 and earlier
+    ///
+    /// - Parameter detents: A set of detents that the sheet can be resized to.
+    /// - Returns: A view with the presentation detents preference set.
```
stackotter: update this `Returns` and the one below to refer to the preference by property name as you did for the bottom 3 modifiers.

Sources/SwiftCrossUI/Views/Modifiers/SheetModifier.swift
stackotter: I'll look into the onDisappear issue. I think it's best if we get it working

stackotter:
On macOS it appears that the `dismiss` button in a nested sheet breaks if you dismiss the parent sheet first.

Sources/SwiftCrossUI/Views/Modifiers/SheetModifier.swift
stackotter: Ok it turns out that onDisappear is just broken in general, don't worry about it in this PR.

Sources/AppKitBackend/AppKitBackend.swift
MiaKoring:
```swift
let result = children.sheetContentNode!.update(
    with: sheetContent(),
    proposedSize: SIMD2(x: 400, y: 0),
    environment: sheetEnvironment,
    dryRun: false
)

backend.updateSheet(
    sheet,
    size: result.size.size,
    onDismiss: { handleDismiss(children: children) }
)
```

this works.
Before I had proposedSize: backend.size(ofSheet: sheet)
pretty logical it didn’t work. When the size was read it was still (0,0). That lead to SCUI squishing it as small as possible. I first tried setting y to 400 as well, but this lead to the sheet keeping 400 as height even though the content was smaller. 0 for y results in a sheet only as high as it needs to be. This renders correctly in WindowingExample. 

Imo the y value can stay constant 0 for now and changed if it needs to be. For the x value however we need to determine a good constant. It needs to be big enough its not dictating the width while being small enough there are no integer overflow problems (Int.max for example crashes). 

I’m going to set it to 10k for now. That should temporarily be fine for almost any resolution, but may need further investigation and optimization.

Sources/Gtk/Widgets/Window.swift
MiaKoring: I should’ve thought of that *facepalm*

thanks, thats so much nicer

MiaKoring:
> On macOS it appears that the `dismiss` button in a nested sheet breaks if you dismiss the parent sheet first.

Dismissing a sheet up the graph should now lead to every sheet below it getting dismissed too.
On all supported platforms.

MiaKoring:
The new sheetParent environment value was added due to UIKitBackend.Sheet not being compatible with UIKitBackend.Window, so environment.with(\.window, sheet) couldn’t be used.

UIKitBackend is also the only (currently sheet-supporting) platform where Backend.Window can’t be used to add a sheet. It needs a UIViewController, accessible trough UIWindow.rootViewController when creating the first sheet. to avoid compiler flags/UIKitBackend specific code in SCUI I decided to make it Any? and let each backend resolve the property/type it needs to create a sheet.

This should also be future proof for any other Backend with Window-Sheet incompatibility.

I noticed I created sheets always from the main Window before (definetely AppKit, Gtk I’m not sure).
This lead to the sheet z order occasionally getting messed up when dismissing a sheet down the graph.

Now using the deepest window to present a sheet should solve this and I was able to switch to a non-critical sheet on AppKitBackend (not sure if it would make a difference with the current implementation).

I don’t know how we could test wether the z order issue actually is fixed now, due to it appearing to be non deterministic from the layer we are working on.

Another nice side effect is a theoretical slight performance improvement in cases where there is a big stack of sheets, because it gets passed the deepest UIViewController instead of needing to loop like before.

I was afraid setting the managedAttachedSheet to nil on scope exiting could cause dismiss animation bugs, this doesn’t seem to be the case, but should be kept in mind just in case. Doing it is important though, because otherwise the root window on GtkBackend and AppKitBackend would keep a reference to the entire last presented Sheet Graph:

RootWindow
┗ managedAttachedSheet
      ┗ managedAttachedSheet
            ┗ managedAttachedSheet
                  ┗ managedAttachedSheet
                        ┗ …
                        
Thus it couldn’t be freed up.

Like you suggested an AppKitBackend.Sheet now keeps a reference to its background node, if set.
So now it only gets its color changed when the function gets called instead of creating a new node each time.

tvOS only accepts UIModalPresentationStyle.formSheet starting on 26+. I decided to fall back to UIModalPresentationStyle.overFullscreen for now to allow developers to decide wether they cover the whole View with opaque content or just part of it (overFullscreen isn't opaque by itself).

https://developer.apple.com/documentation/uikit/uimodalpresentationstyle/formsheet
https://developer.apple.com/documentation/uikit/uimodalpresentationstyle/overfullscreen

CI fails with (formSheet unavailable on tvOS), locally it works,
```swift
if #available(tvOS 26.0, *) {
    sheet.modalPresentationStyle = .formSheet
} else {
    sheet.modalPresentationStyle = .overFullScreen
}
```
so I only used .overFullScreen for now.

I believe thats everything important to know about todays changes.

Sources/SwiftCrossUI/ViewGraph/PreferenceValues.swift
```diff
@@ -21,7 +21,7 @@ public struct PreferenceValues: Sendable {
     /// The drag indicator visibility for a sheet presentation. Applies to enclosing sheets.
     public var presentationDragIndicatorVisibility: Visibility?
 
-    /// The backgroundcolor of a sheet. Applies to enclosing sheets.
+    /// The backgroundcolor of a sheet.
```
stackotter, line 24: This one should have `Applies to enclosing sheets`. The comment below was for the next property. Make `backgroundcolor` two words as well.

Sources/SwiftCrossUI/Views/Modifiers/SheetModifier.swift
```diff
@@ -80,17 +82,22 @@ struct SheetModifier<Content: View, SheetContent: View>: TypeSafeView {
             let dismissAction = DismissAction(action: { [isPresented] in
                 isPresented.wrappedValue = false
             })
-            let sheetEnvironment = environment.with(\.dismiss, dismissAction)
+
+            var sheetEnvironment =
```
stackotter, line 86: This variable can be a `let` now.

Sources/SwiftCrossUI/Views/Modifiers/SheetModifier.swift
```diff
@@ -1,5 +1,7 @@
 extension View {
     /// Presents a conditional modal overlay. `onDismiss` gets invoked when the sheet is dismissed.
+    ///
+    /// Internal UIViewController.modalPresentationStyle falls back to .overFullScreen (non-opaque) on tvOS.
```
stackotter, line 4: Replace this sentence with something like `On most platforms sheets appear as form-style modals. On tvOS <26, sheets appear as full screen overlays (non-opaque).`

Sources/SwiftCrossUI/Backend/AppBackend.swift
```diff
@@ -626,19 +627,19 @@ public protocol AppBackend: Sendable {
     /// app modal, a standalone window, or a modal for a window of its choosing.
     func showSheet(
         _ sheet: Sheet,
-        window: Window?
+        sheetParent: Any
```
stackotter: To avoid the `Any`, I think a better signature would be `func showSheet(_ sheet: Sheet, window: Window?, parentSheet: Sheet?)`. Then on iOS the equivalent of sheetParent can be retrieved via `parentSheet ?? window!.rootViewController`.

Sources/SwiftCrossUI/Backend/AppBackend.swift
```diff
@@ -626,19 +627,19 @@ public protocol AppBackend: Sendable {
...
     )
 
     /// Dismisses a sheet programmatically.
     /// Gets used by the ``View/sheet`` modifier to close a sheet.
-    func dismissSheet(_ sheet: Sheet, window: Window)
+    func dismissSheet(_ sheet: Sheet, sheetParent: Any)
```
stackotter, line 635: Update this to match `showSheet` as well

Sources/AppKitBackend/AppKitBackend.swift
```diff
@@ -1689,6 +1690,122 @@ public final class AppKitBackend: AppBackend {
...
+            return SIMD2(x: 0, y: 0)
+        }
+        return SIMD2(x: Int(size.width), y: Int(size.height))
+    }
+
+    public func showSheet(_ sheet: NSCustomSheet, sheetParent: Any) {
+        // Critical sheets stack. beginSheet only shows a nested sheet
+        // after its parent gets dismissed.
+        let window = sheetParent as! NSCustomWindow
+        window.beginSheet(sheet)
```
stackotter, line 1732: Investigate `beginSheet` vs `beginCriticalSheet`

stackotter:
- Updated the sheet APIs to avoid Any.
- Audited the backend implementations to ensure behaviour regarding the onDismiss handlers of implicitly dismissed nested sheets. I've decided that they should be called even though the parent sheet was dismissed programmatically. I also noticed that many backends trigger dismissals on nested sheets automatically so we could eliminate the need for keeping references to child sheets in said backends.
- Audited the backend implementations to ensure consistent onDismiss ordering. onDismiss now gets called *after* the sheet gets dismissed, across all backends (AppKitBackend was an odd one out).
- Merged presentation preference setting into updateSheet because I realised that most of the `setPresentation...` AppBackend methods should still be called in the case of a nil value to properly reset their corresponding backend property back to its default value (rather than retaining their previous non-nil backend). This made it make sense to merge the `setPresentation` methods into `updateSheet` because they were always all called at in succession.
- Fixed issue in windowing example where dismissing the 5 second sheet interactively didn't cancel the associated 5 second task, so if you reopened the sheet within 5 seconds, it would get cancelled sooner than 5 seconds.
- Updated the windowing example to allow infinite nesting of sheets, and updated 'dismiss parent sheet' to 'dismiss root sheet' because that makes more sense with infinite nesting.
- Addressed the comments from my review.
- And probably a few other things I missed in this summary

stackotter:
- Centered sheet contents horizontally on iOS
- Handled the case of an empty detents array on iOS (now defaults to [.large]. Without this, UIKit crashes.
- Fixed nested iOS sheet dismissal (possibly an issue introduced by my code). We need to call dismiss twice when there's a nested sheet, because the first time it dismisses the nested sheet, and the next time it defers to the parent to dismiss the sheet itself. This is because the presenting controller is the one responsible for dismissing presented controllers.
- Added a default sheet background colour on iOS (otherwise it's transparent and looks really weird).
- Updated UIKitBackend's interpretation of automatic drag indicator visibility to show the indicator when there is more than one detent.

PR #220, opened 2025-10-11
View/textSelectionEnabled

MiaKoring:
As Philocalyst requested I added a modifier to control wether Text() could be selected by the user.

On UIKitBackend I switched from UILabel to a new Class inheriting from UILabel to show a Menu on long press. 
Oob selectability is not built into UILabels. 
I tested using UITextView like NSTextView in the AppKitBackend, but it lead to rendering issues. 
The custom class solution should work as a drop in replacement, rendering exactly the same as before, only with more funktionality.

The isTextSelectionEnabled boolean gets set in the environment.

AppKitBackend, WinUIBackend and GtkBackend required a single line change in the updateTextView function, setting the flag on the native View.

On AppKitBackend the cursor is not always reacting to a change in the reactability instantly and it also doesn’t unselect it (Gtk and WinUI do unselect it) but this seems to be an AppKit Limitation and shouldn't affect the majority of usecases.

I’m sorry for the huge amount of confusing added and removed top level files, I assume a compile process placed them there for some reason.

It should be mergable with [my other pr](https://github.com/stackotter/swift-cross-ui/pull/219) without conflicts.

### DIFF
--- a/Sources/SwiftCrossUI/Views/Modifiers/TextSelectionModifier.swift
+++ b/Sources/SwiftCrossUI/Views/Modifiers/TextSelectionModifier.swift
@@ -1,6 +1,5 @@
 extension View {
-    /// Set selectability of contained text.
-    /// ignored on tvOS
+    /// Set selectability of contained text. Ignored on tvOS.
     public func textSelectionEnabled(_ isEnabled: Bool = true) -> some View {
         EnvironmentModifier(
             self,
--- a/Sources/UIKitBackend/UIKitBackend+Control.swift
+++ b/Sources/UIKitBackend/UIKitBackend+Control.swift
@@ -214,6 +214,7 @@ final class TappableWidget: ContainerWidget {
         }
     }
 #endif
+
 @available(tvOS, unavailable)
 final class SliderWidget: WrapperWidget<UISlider> {
     var onChange: ((Double) -> Void)?
--- a/Sources/SwiftCrossUI/Views/ForEach.swift
+++ b/Sources/SwiftCrossUI/Views/ForEach.swift
@@ -28,7 +28,7 @@ extension ForEach where Items == [Int] {
         self.elements = Array(range)
         self.child = child
     }
-    
+
     /// Creates a view that creates child views on demand based on a given Range
     @_disfavoredOverload
     public init(


Sources/UIKitBackend/UIKitBackend+Passive.swift
```diff
@@ -36,8 +36,7 @@ extension UIKitBackend {
     }
 
     public func createTextView() -> Widget {
-        let widget = WrapperWidget<UILabel>()
-        widget.child.numberOfLines = 0
```
bbrk24, line 40: Does the subclass do something that makes it unnecessary? With plain `UILabel`, forgetting this means multiline text doesn't work, and I don't see anywhere in the subclass that sets this property.

MiaKoring: Thank you, I read the description of this property wrong, leading to me thinking 0 was the default value.

philocalyst:
yas

Examples/Sources/GreetingGeneratorExample/GreetingGeneratorApp.swift
```diff
@@ -26,9 +27,11 @@ struct GreetingGeneratorApp: App {
                         }
                     }
 
+                    Toggle("Selectable Greeting", active: $isGreetingSelectable)
                     if let latest = greetings.last {
                         Text(latest)
                             .padding(.top, 5)
+                            .textSelectionEnabled(isGreetingSelectable)
```
stackotter, line 34: I noticed that this differs from SwiftUI's equivalent `.textSelection(_:)` api. That said, I do think that the spelling you've chosen makes more sense.

I feel like it's worth going with your better spelling, and I also feel like your spelling should be relatively discoverable for those using LSPs, but I still felt it was worth commenting on this for others to think about (and to document this decision).

Sources/SwiftCrossUI/Environment/EnvironmentValues.swift
```diff
@@ -95,6 +95,9 @@ public struct EnvironmentValues {
     /// The style of toggle to use.
     public var toggleStyle: ToggleStyle
 
+    /// Wether a Text should be selectable. Set by ``View/textSelectionEnabled(_:)``.
```
stackotter: Change to:

```swift
/// Whether the text should be selectable. Set by ``View/textSelectionEnabled(_:)``.
```

Sources/SwiftCrossUI/Environment/EnvironmentValues.swift
```diff
@@ -95,6 +95,9 @@ public struct EnvironmentValues {
...
+    public var isTextSelectionEnabled: Bool = false
```
stackotter: This file generally follows the pattern of setting the default values in the initialiser. Don't really have specific justification for that, but it'd be best to follow that to keep things neat.

Sources/SwiftCrossUI/Views/Modifiers/TextSelectionModifier.swift
```diff
@@ -0,0 +1,12 @@
+extension View {
+    /// Set SwiftUI.Text selectability
```
stackotter: Change to:

```swift
/// Set selectability of contained text.
```

In future this modifier may affect other forms of text views such as markdown views, so it's best to keep the description abstract.

Sources/SwiftCrossUI/Views/Modifiers/TextSelectionModifier.swift
```diff
@@ -0,0 +1,12 @@
...
+    public func textSelectionEnabled(_ isEnabled: Bool = true) -> some View {
+        EnvironmentModifier(
+            self,
+            modification: { environment in
+                var newEnvironment = environment
+                newEnvironment.isTextSelectionEnabled = isEnabled
+                return newEnvironment
```
stackotter: You can shorten this to `environment.with(\.isTextSelectionEnabled, isEnabled)` using the handy `with` helper method.

Sources/UIKitBackend/UIKitBackend+Passive.swift
```diff
@@ -104,3 +105,65 @@ extension UIKitBackend {
...
+    private func setupTextSelection() {
+        let longPress = UILongPressGestureRecognizer(target: self, action: #selector(didLongPress))
+        addGestureRecognizer(longPress)
+        isUserInteractionEnabled = true
+    }
+
+    @objc private func didLongPress(_ gesture: UILongPressGestureRecognizer) {
+        guard isSelectable, gesture.state == .began, let text = self.attributedText?.string,
+            !text.isEmpty
+        else {
```
stackotter: Format long guards like so for clarity; (just a preference thing)

```swift
guard
    condition1,
    condition2,
    condition3,
    ...
else {
```

Sources/UIKitBackend/UIKitBackend+Passive.swift
```diff
@@ -104,3 +105,65 @@ extension UIKitBackend {
...
+    }
+
+    override func canPerformAction(_ action: Selector, withSender sender: Any?) -> Bool {
+        return action == #selector(copy(_:))
+    }
+
+    private func textRect() -> CGRect {
+        let inset: CGFloat = -4
+        return textRect(forBounds: bounds, limitedToNumberOfLines: numberOfLines).insetBy(
+            dx: inset, dy: inset)
```
stackotter: Wrap the `.insetBy` onto the next line to avoid the awkward wrap;

```swift
return textRect(...)
    .insetBy(...)
```

Examples/Sources/GreetingGeneratorExample/GreetingGeneratorApp.swift
MiaKoring: Yes, I noticed the SwiftUI equivalent expects enabled or disabled. Since this is basically representable through a Bool I chose that naming, similar to something like scrollDisabled(_:). 

So in the case I imagine to be the most common one (adding selectability) it should read slightly nicer due to not needing content inside the brackets. 

(also a bool and different modifier naming is less vulnerable to lsp/intellisense problems, because its only one time something needs to be found instead of two)

stackotter:
Thanks for implementing all of the requested changes. I'm currently in the process of testing this PR out on iOS (faced some difficulties cause my phone wasn't paired to my current Xcode version and I didn't have the iOS platform support installed yet).

The modifier seems to work well on both Windows and Linux. ~~On macOS the text seems to remain selectable forever after it's been made selectable at least once.~~

I've found a fix. Add these lines before setting `isSelectable`;

```swift
if field.isSelectable && !environment.isTextSelectionEnabled {
    field.abortEditing()
}
```

This fix also clears the current selection when text selection gets disabled.

Each editable/selectable text field in the app has a `currentEditor()` method which returns the editor being used on that text field if applicable. That instance stores the selected range etc. I guess `NSTextField.isSelectable` just controls whether or not someone can *begin* selecting, but doesn't actually stop them from continuing to select if they already started.

stackotter:
Finally managed to test it on my iPhone. What would we have to do to support proper iOS text selection? For small text like the greetings generator this copy button is ok, but for large paragraphs the text selection ux would be nicer. Do we have to switch to a different text view type?

MiaKoring:
> Finally managed to test it on my iPhone. What would we have to do to support proper iOS text selection? For small text like the greetings generator this copy button is ok, but for large paragraphs the text selection ux would be nicer. Do we have to switch to a different text view type?

We would need to, yes. Like this its actually equal to the swiftui implementation https://developer.apple.com/documentation/swiftui/view/textselection(_:) so in my opinion it can stay like that. (Aside from maybe adding the sharebutton)

stackotter:
Ah interesting, yeah I agree that it can stay like this for now then.

MiaKoring:
Your selection removing suggestion worked :)

Sources/SwiftCrossUI/Views/Modifiers/TextSelectionModifier.swift
```diff
@@ -0,0 +1,11 @@
+extension View {
+    /// Set selectability of contained text.
+    /// ignored on tvOS
```
stackotter: `/// Set selectability of contained text. Ignored on tvOS.

Sources/UIKitBackend/UIKitBackend+Control.swift
```diff
@@ -214,9 +212,8 @@ final class HoverableWidget: ContainerWidget {
                 default: break
             }
         }
-    #endif
-}
-
+    }
+#endif
```
stackotter, line 216: Add a newline after the `#endif` (to put a blank line between types)

stackotter (approved):
Looks good now

PR #224, opened 2025-10-16
Fix tvOS CI

### LINKED ISSUE #223: Apple TV workflow step is failing, but reporting a success
The Apple TV workflow [is currently failing silently](https://github.com/stackotter/swift-cross-ui/actions/runs/17899975995/job/50891652716#step:6:1579). We need to first make it actually fail, and then fix the underlying reason that it's failing.

bbrk24:
Fixes #223 

Turns out the `set -e` call in the script was almost immediately followed by `set +e`, so I had to move it further down to make the tests fail correctly. Then I updated the code so now it actually builds on tvOS.

### DIFF
--- a/Sources/UIKitBackend/UIKitBackend+Control.swift
+++ b/Sources/UIKitBackend/UIKitBackend+Control.swift
@@ -61,9 +61,13 @@ final class TextEditorWidget: WrapperWidget<UITextView>, UITextViewDelegate {
     var onChange: ((String) -> Void)?
     var isEditable: Bool = true {
         didSet {
-            if !isEditable {
-                child.resignFirstResponder()
-            }
+            #if os(tvOS)
+                if !isEditable {
+                    child.resignFirstResponder()
+                }
+            #else
+                child.isEditable = isEditable
+            #endif
         }
     }
 


bbrk24:
Looks like about six hours ago, a commit was pushed to https://github.com/thebrowsercompany/swift-windowsfoundation that means it can no longer be pulled in by xcodebuild. (Not sure why it's succeeding on Linux; you'd think xcodebuild and SPM would resolve dependencies the same way.) Sometime thereafter, the repository was archived, so we shouldn't expect an update to come and fix it. I can't point it at an older commit because one of our dependencies also points at `main`, so SPM will complain about the mismatched commits.

Sources/UIKitBackend/UIKitBackend+Control.swift
```diff
@@ -332,7 +347,7 @@ extension UIKitBackend {
     ) {
         let textEditorWidget = textEditor as! TextEditorWidget
 
-        textEditorWidget.child.isEditable = environment.isEnabled
```
stackotter, line 335: Would we have to still set `.child.isEditable` somewhere? This change seems like it'd break something.

bbrk24: I tested this on an iOS Simulator and it seemed to work fine. The delegate method prevents you from starting to type in it, and the `didSet` listener forcibly unfocuses the field if you're already typing when it becomes disabled.

stackotter: Ok sounds good. Would not setting the underlying editability affect accessibility metadata at all? I haven’t done accessibility stuff on iOS before, so I have no clue

bbrk24: That's a good question and I hadn't considered that. I will double-check when I'm back at my computer later today.

bbrk24: So it turns out that (on iOS) deactivating the text editor in this way still lets you navigate to it, it's just that typing does nothing once you do. Setting `isEditable` doesn't let you navigate to it. I've updated the `didSet` accordingly, leaving the behavior I had for tvOS and using `child.isEditable` on other systems.

stackotter (approved):
Thanks for looking into that. Will merge once CI passes.

PR #238, opened 2025-10-25
Fix/progess spinner resizing

MiaKoring:
Fixes https://github.com/stackotter/swift-cross-ui/issues/237

**Visually Breaking Change for some existing UIs using ProgressSpinner**

This PR makes ProgressSpinners resizable on supported backends. Unsupported Backends change the container size without changing how the Widgets is actually being rendered to the user. 

All changes are SCUI-Level, so even Backends I didn’t test (Gtk3) should benefit from it without breaking.

It may break some existing UIs though, as it now can be squished smaller than its natural size. To make it not entirely disappear (without setting a frame manually) I set its width/height to a minimum of 10 points, it can‘t shrink smaller.

To determine its rendered size it takes the smaller axis length and applies it to both width and height (making sure it always fits inside the applied constraints).

This can lead to it being rendered smaller than a set frame, if only one axis is set and its being squished from the opposite site. setting both axes in .frame eliminates this „problem“.

Gtk seems to have a min size for its Spinner. This can lead to clipping on too small externally applied constraints (at least on Ubuntu). I don’t really consider this to be an issue though as it can easily be avoided by the developer and other views can clip too.

### DIFF
--- a/Examples/Sources/ControlsExample/ControlsApp.swift
+++ b/Examples/Sources/ControlsExample/ControlsApp.swift
@@ -88,7 +88,7 @@ struct ControlsApp: App {
                         VStack {
                             Toggle(
                                 "Enable ProgressView resizability", isOn: $isProgressViewResizable)
-                            Slider($progressViewSize, minimum: 10, maximum: 100)
+                            Slider(value: $progressViewSize, in: 10...100)
                             ProgressView()
                                 .resizable(isProgressViewResizable)
                                 .frame(width: progressViewSize, height: progressViewSize)


bbrk24:
My concern with resizable spinners is that I'll often do something like
```swift
HStack {
  Spacer()

  ProgressView()

  Spacer()
}
```
to center a spinner within a larger view. Would this lead to a ridiculously large spinner on a large enough screen?

stackotter:
> My concern with resizable spinners is that I'll often do something like
> 
> ```swift
> HStack {
>   Spacer()
> 
>   ProgressView()
> 
>   Spacer()
> }
> ```
> 
> to center a spinner within a larger view. Would this lead to a ridiculously large spinner on a large enough screen?

Perhaps this resizability can be gated behind a `.resizable()` modifier similar to Image resizability?

MiaKoring:
> > My concern with resizable spinners is that I'll often do something like
> > ```swift
> > HStack {
> >   Spacer()
> > 
> >   ProgressView()
> > 
> >   Spacer()
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > to center a spinner within a larger view. Would this lead to a ridiculously large spinner on a large enough screen?
> 
> Perhaps this resizability can be gated behind a `.resizable()` modifier similar to Image resizability?

Exactly what I did.

I’m going to replace the function with how its done on the Image in a moment. I like it better.

bbrk24:
It doesn't need to be done now (I could do this in a future PR), but maybe some code could be added to UIKitBackend to change the spinner between `.medium` and `.large` depending on its frame? I don't think there's an `updateProgressSpinner` backend method, so this would probably have to go on the widget's `layoutSubviews` method.

Whatever threshold we choose, we have to make sure it's large enough that the `.large` style is never chosen when it could be cut off on any platform (not just OS, but also based on OS version and maybe even accessibility settings. If you don't want to test it all I totally understand; I can get to it this weekend.)

MiaKoring:
> maybe some code could be added to UIKitBackend to change the spinner between `.medium` and `.large` depending on its frame?

totally. I decided to not do it in this pr because the scope was scui isolated and I wanted to stay it that way. (Also I wasn't sure yet how I'd like the api to look for it. Single Backend support is way harder for api design imo. Especially when its kind of overlapping with another modifier but not really.)

Maybe .style() or whatever it would be called could apply .resizable() and .frame(constant size) depending on .medium or .large? so it wouldn't be UIKit Backend specific anymore (even though the other backends wouldn't need specific code)

Examples/Sources/ControlsExample/ControlsApp.swift
```diff
@@ -82,6 +90,7 @@ struct ControlsApp: App {
                 Toggle(enabled ? "Disable all" : "Enable all", active: $enabled)
                     .padding()
             }
+
```
stackotter: Remove this blank line

Sources/SwiftCrossUI/Views/ProgressView.swift
```diff
@@ -50,6 +51,14 @@ public struct ProgressView<Label: View>: View {
         self.kind = .bar
         self.progress = value.map(Double.init)
     }
+
+    /// Makes the ProgressView resize to fit the available space.
+    /// Only affects `Kind.spinner`.
```
stackotter: Update the reference to

```
``Kind/spinner``
```

so that it links through.

Sources/SwiftCrossUI/Views/ProgressView.swift
```diff
@@ -101,7 +110,10 @@ extension ProgressView where Label == Text {
 }
 
 struct ProgressSpinnerView: ElementaryView {
-    init() {}
+    let isResizable: Bool
```
stackotter, line 113: And blank line after this var

Sources/SwiftCrossUI/Views/ProgressView.swift
```diff
@@ -114,8 +126,33 @@ struct ProgressSpinnerView: ElementaryView {
...
-        ViewUpdateResult.leafView(
-            size: ViewSize(fixedSize: backend.naturalSize(of: widget))
+        let naturalSize = backend.naturalSize(of: widget)
+        guard isResizable else {
+            // Required to reset its size when resizability
+            // gets changed at runtime
+            backend.setSize(of: widget, to: naturalSize)
+            return ViewUpdateResult.leafView(size: ViewSize(fixedSize: naturalSize))
+        }
+        let min = max(min(proposedSize.x, proposedSize.y), 10)
```
stackotter: Rename to `minimumDimension` for clarity and to avoid shadowing the `min` function.

stackotter: Do we want to have a lower bound of 10? I feel like we should just let the spinner vanish to 0 by 0 if sufficiently squished (like we do for other resizable views such as images or shapes). 10 is also backend-specific, so may not be a suitable minimum for all backends.

Sources/SwiftCrossUI/Views/ProgressView.swift
```diff
@@ -114,8 +126,33 @@ struct ProgressSpinnerView: ElementaryView {
...
+        let size = SIMD2(
+            min,
+            min
+        )
+        if !dryRun {
+            // Doesn't change the rendered size of ProgressSpinner
+            // on UIKitBackend, but still sets container size to
+            // (width: n, height: n) n = min(proposedSize.x, proposedSize.y)
```
stackotter: Does this centre the spinner in the proposed frame under UIKitBackend?

Sources/SwiftCrossUI/Views/ProgressView.swift
```diff
@@ -114,8 +126,33 @@ struct ProgressSpinnerView: ElementaryView {
...
+            backend.setSize(of: widget, to: size)
+        }
+        return ViewUpdateResult.leafView(
+            size: ViewSize(
+                size: size,
+                idealSize: naturalSize,
+                minimumWidth: 10,
+                minimumHeight: 10,
```
stackotter: Set these to zero after applying the above feedback for the `min` var.

stackotter:
Here's what I've found from testing locally;

## GtkBackend and Gtk3Backend

- Work perfectly.

## AppKitBackend

- Spinner animates its size kinda weirdly when switching between enabled and disabled. I think it shouldn't animate at all (to make our job easier when we add an animation system).

- The `ControlsExample` crashes when I use the spinner size slider. It appears to be due to a feedback loop in the window sizing code (that eventually leads to running out of stack space). The feedback loop is likely unrelated to your changes. If you can reproduce the crash locally then it'd be worth making a reproducer for it. Otherwise I can once I get the time.
- I've confirmed that the crash doesn't occur when the ControlsExample is inside a ScrollView. I think it should be in a ScrollView anyway given that it's so tall now. So feel free to pop it in a ScrollView and I'll look into the crash later.

## UIKitBackend

- Works other than the behaviour you documented (that the spinner doesn't resize).

## WinUIBackend

- Works perfectly.

Examples/Sources/ControlsExample/ControlsApp.swift
```diff
@@ -22,75 +22,81 @@ struct ControlsApp: App {
...
-                    Toggle("Enable ProgressView resizability", active: $isProgressViewResizable)
-                    Slider($progressViewSize, minimum: 10, maximum: 100)
-                    ProgressView()
-                        .resizable(isProgressViewResizable)
-                        .frame(width: progressViewSize, height: progressViewSize)
+                        Toggle("Enable ProgressView resizability", active: $isProgressViewResizable)
+                        Slider($progressViewSize, minimum: 10, maximum: 100)
+                        Button("Randomize progress view size") {
+                            progressViewSize = Int.random(in: 10...100)
+                        }
```
stackotter: I think this should be removed. I assume it was probably just used to test for size animation when fixing the AppKitBackend issues?

Sources/SwiftCrossUI/Backend/AppBackend.swift
```diff
@@ -541,6 +541,15 @@ public protocol AppBackend: Sendable {
     /// Creates an indeterminate progress spinner.
     func createProgressSpinner() -> Widget
 
+    /// Changes the Spinner's Size.
+    /// Required due to AppKitBackend needing special treatment.
+    /// Forward to ``AppBackend/setSize(of widget: Widget, to size: SIMD2<Int>)``
+    /// on other Backends.
+    func setProgressSpinnerSize(
```
stackotter: Rename to `setSize(ofProgressSpinner:to:)` for consistency.

Sources/SwiftCrossUI/Backend/AppBackend.swift
```diff
@@ -541,6 +541,15 @@ public protocol AppBackend: Sendable {
     /// Creates an indeterminate progress spinner.
     func createProgressSpinner() -> Widget
 
+    /// Changes the Spinner's Size.
+    /// Required due to AppKitBackend needing special treatment.
+    /// Forward to ``AppBackend/setSize(of widget: Widget, to size: SIMD2<Int>)``
+    /// on other Backends.
```
stackotter: Replace these docs with the following.

```swift
/// Sets the size of a progress spinner.
///
/// This method exists because AppKitBackend requires special handling to resize progress spinners.
///
/// The default implementation forwards to ``AppBackend/setSize(of:to:)``.
```

I know that most of the other backend methods are just one big blob of text, but I'm planning on changing that soon, and I figured that given that I was already suggesting wording changes and fixing the symbol link, I might as well update it to the new formatting while I was at it.

Putting the blank lines splits the documentation into multiple paragraphs. The first paragraph is the one that LSPs show, and it's the short blurb used in generated DocC documentation under symbol links I think.

Examples/Sources/ControlsExample/ControlsApp.swift
MiaKoring: yess, my bad

Sources/SwiftCrossUI/Views/ProgressView.swift
```diff
@@ -120,45 +120,52 @@ struct ProgressSpinnerView: ElementaryView {
...
+
+        let minimumDimension: Double
+
+        if let proposedWidth = proposedSize.width,
+            let proposedHeight = proposedSize.height
+        {
+            minimumDimension = max(min(proposedWidth, proposedHeight), 0)
+        } else {
+            minimumDimension = Double(min(naturalSize.x, naturalSize.y))
         }
```
stackotter: Suppose that the natural size is 10x10, and the proposed Size is 50xnil. This code would give the spinner a size of 10x10, whereas it should be 50x50.

Replace with;

```swift
let minimumDimension = max(min(
    proposedSize.width ?? naturalSize.x,
    proposedSize.height ?? naturalSize.y
), 0)
```

Sources/SwiftCrossUI/Views/ProgressView.swift
```diff
@@ -120,45 +120,52 @@ struct ProgressSpinnerView: ElementaryView {
...
-                size: size,
-                idealSize: naturalSize,
-                minimumWidth: 0,
-                minimumHeight: 0,
-                maximumWidth: nil,
-                maximumHeight: nil
-            )
+
+        return ViewLayoutResult.leafView(
+            size: .init(minimumDimension, minimumDimension)
```
stackotter: Use `ViewSize` instead of `.init` as a stylistic thing

stackotter (changes requested):
The layout code broke a bit during the rebase, but it should be a simple enough fix

Sources/SwiftCrossUI/Views/ProgressView.swift
MiaKoring: You are right, in the moment I had in my mind that either both ore neither would be set, even though this obviously wouldn’t make sense if its not constrained in one axis.

This should still result in 10 though right? because 10 is smaller than 50, so it would be used by min() 

I updated the else instead, to use the set value on proposedSize, if existent, otherwise defaulting to 0, compared by max() against the min() of natural size. I think this solves it, but please confirm.

stackotter: I think the correct result in this case is 50x50, because having a single unspecified axis is essentially asking "what's the ideal size for this axis, given that the other axis has been proposed blah pixels?".

My gut feeling is that a resizable spinner should behave the same as `            Color.blue.aspectRatio(1, contentMode: .fit)`. A resizable spinner ideally wants to grow as big as it can, unless both axes are unspecified, in which case it has nothing to base its size on and has to fall back to the widget's natural size.

When given a partial proposal, we want to figure out the biggest square that fits the proposal. One axis is unspecified, so we can make it as big as we want. To do so, we would make the spinner a square with its size dictated by the specified dimension.

My code is actually wrong; it doesn't match my described behaviour; and your code also doesn't match this behaviour.

I think what we need is this;

```swift
if let proposedWidth = proposedSize.width, let proposedHeight = proposedSize.height {
    dimension = min(proposedWidth, proposedHeight)
} else if let proposedWidth = proposedSize.width {
    dimension = proposedWidth
} else if let proposedHeight = proposedSize.height {
    dimension = proposedHeight
} else {
    dimension = min(naturalSize.x, naturalSize.y)
}
```

That can be compressed to this;

```swift
if let proposedWidth = proposedSize.width, let proposedHeight = proposedSize.height {
    dimension = min(proposedWidth, proposedHeight)
} else {
    dimension = proposedSize.width ?? proposedSize.height ?? min(naturalSize.x, naturalSize.y)
}
```

But that feels a bit obtuse to me, and takes a little while to understand, so the former may be better.

stackotter (changes requested):

Examples/Sources/ControlsExample/ControlsApp.swift
```diff
@@ -83,11 +85,21 @@ struct ControlsApp: App {
                             Text("Value: \(text)")
                         }
 
+                        VStack {
+                            Toggle(
+                                "Enable ProgressView resizability", isOn: $isProgressViewResizable)
+                            Slider($progressViewSize, minimum: 10, maximum: 100)
```
stackotter: This API has changed on main now I think. Should be (value:in:) to avoid deprecation warning.

stackotter (dismissed):
Code looks good now. I'll test locally (when I get home) and then merge if the sizing-related changes had the intended effect.

stackotter (approved):
Tested locally and the sizing works as expected. Will merge once the CI finishes

PR #241, opened 2025-11-03
Renamed StateProperty to ObservableProperty and made the protocol public

MiaKoring:
Like the title says, I changed the protocol to be publicly available, allowing for more third party view update triggering property wrappers, i.e. @Query or @AppStorage

Also I renamed it do ObservableProperty, decoupling it verbally from @State.

Doing this involved declaring the protocol-required properties and functions as public too and replacing StateProperty with ObservableProperty wherever it gets used.

0 sideeffects expected, as there were no logic changes, aside from making some functions/values more accessible.


Sources/SwiftCrossUI/State/State.swift
```diff
@@ -129,7 +129,10 @@ public struct State<Value>: DynamicProperty, StateProperty {
     }
 }
 
-protocol StateProperty {
+/// Declaring conformance to ObservableProperty is required for
+/// SwiftCrossUI to automatically register an observer on the
+/// conforming object's Publisher.
+public protocol ObservableProperty {
```
stackotter: Move this protocol its own file now that it's part of the public API.

Sources/SwiftCrossUI/State/State.swift
```diff
@@ -129,7 +129,10 @@ public struct State<Value>: DynamicProperty, StateProperty {
     }
 }
 
-protocol StateProperty {
+/// Declaring conformance to ObservableProperty is required for
+/// SwiftCrossUI to automatically register an observer on the
+/// conforming object's Publisher.
```
stackotter: Update to;

```
/// View properties that conform to ObservableProperty are automatically observed by SwiftCrossUI.
///
/// This protocol is intended to be implemented by property wrappers. You shouldn't
/// have to implement it for your own model types.
```

Sources/SwiftCrossUI/State/State.swift
stackotter: I think that `ObservableProperty` should inherit from `DynamicProperty`, because it doesn't make sense to implement an `ObservableProperty` whose state doesn't persist across view recalculations.

stackotter (changes requested):
Just requested a few small changes

Sources/SwiftCrossUI/State/State.swift
MiaKoring: You are right, I didn’t think of it, just changed the name on what it used before

stackotter (approved):
Thanks, looks great now.

PR #242, opened 2025-11-03
View/environment modifier, support for both EnvironmentValues and EnvironmentKey (used for custom storage in the environment)

MiaKoring:
This PR adds:
- EnvironmentValues/with(key:, value:) //equivalent to with(keyPath: newValue:) but for non-builtin values
- view modifiers for both builtin (key path) and not builtin (EnvironmentKey) values

For demonstration and testing purposes I changed GreetingGeneratorExample’s latest greeting display to get its value over an environment key.

The @Environment property wrapper was changed to additionally support EnvironmentKeys.

I did some limited testing which worked like expected. The modifiers just use the EnvironmentModifier View just like we do with most other modifiers.


Examples/Sources/GreetingGeneratorExample/GreetingGeneratorApp.swift
```diff
@@ -51,3 +52,15 @@ struct GreetingGeneratorApp: App {
         }
     }
 }
+
+struct EnvironmentDisplay: View {
+    @Environment(TestKey.self) var value: String?
```
stackotter: Add a newline after this.

Examples/Sources/GreetingGeneratorExample/GreetingGeneratorApp.swift
```diff
@@ -51,3 +52,15 @@ struct GreetingGeneratorApp: App {
         }
     }
 }
+
+struct EnvironmentDisplay: View {
```
stackotter: Add a doc comment along the lines of `/// This intermediate view exists to show the usage of custom environment keys. In reality it is not necessary.` so that future readers don't get too confused.

Examples/Sources/GreetingGeneratorExample/GreetingGeneratorApp.swift
```diff
@@ -51,3 +52,15 @@ struct GreetingGeneratorApp: App {
...
+    @Environment(TestKey.self) var value: String?
+    var body: some View {
+        Text(value ?? "nil")
+    }
+}
+
+struct TestKey: EnvironmentKey {
```
stackotter: Rename to `LatestGreetingKey` to make its usage more self-evident.

Examples/Sources/GreetingGeneratorExample/GreetingGeneratorApp.swift
stackotter: Rename to `LatestGreetingDisplay` for clarity.

Sources/SwiftCrossUI/Environment/Environment.swift
```diff
@@ -36,21 +36,26 @@
 /// ```
 @propertyWrapper
 public struct Environment<Value>: DynamicProperty {
-    var keyPath: KeyPath<EnvironmentValues, Value>
+    var keyPath: KeyPath<EnvironmentValues, Value>?
+    var environmentKey: (any EnvironmentKey.Type)?
```
stackotter: Introduce an enum to encode the requirement that we want one property or the other to be non-nil (not neither, and not both). This enum would just be an internal detail.

```swift
enum EnvironmentLocation<Value> {
    case keyPath(KeyPath<EnvironmentValues, Value>)
    case environmentKey(any EnvironmentKey.Type)
}
```

You can then replace these two properties with a single `var location: EnvironmentLocation<Value>` property, and then replace the existing if lets with switch statements.

Sources/SwiftCrossUI/Environment/Environment.swift
```diff
@@ -36,21 +36,26 @@
...
 
     public func update(
         with environment: EnvironmentValues,
         previousValue: Self?
     ) {
-        value.value = environment[keyPath: keyPath]
+        if let keyPath {
+            value.value = environment[keyPath: keyPath]
+        } else if let environmentKey {
+            value.value = (environment[environmentKey] as! Value)
```
stackotter: I've thought about how to avoid this force unwrap a little bit, and it seems like we can make `Value` a primary associated type of `EnvironmentKey`, and then update the environmentKey property to have type `any EnvironmentKey<Value>.Type`. This should make things a bit safer, and appears to be a backwards compatible change (because users of `EnvironmentKey` aren't forced to use the primary associated type syntax)?

I had to Google the syntax, so I figured I might as well include it here. To make `Value` a primary associatetype do the following;

```swift
protocol EnvironmentKey<Value> {
    associatedtype Value
    ...
}
```

Sources/SwiftCrossUI/Environment/EnvironmentValues.swift
```diff
@@ -222,6 +222,14 @@ public struct EnvironmentValues {
         environment[keyPath: keyPath] = newValue
         return environment
     }
+
+    /// Returns a copy of the environment with the specified key set to the
+    /// provided new value.
+    public func with<T: EnvironmentKey>(key: T.Type, value: T.Value) -> Self {
```
stackotter: Give the parameters each the underscore label to match the keypath variant (and rename `value` to `newValue`)

Sources/SwiftCrossUI/Views/Modifiers/EnvironmentModifier.swift
```diff
@@ -37,3 +37,21 @@ package struct EnvironmentModifier<Child: View>: View {
         )
     }
 }
+
+extension View {
+    /// Modifies the environment of the View its applied to.
+    public func environment<T: EnvironmentKey>(key: T.Type, value: T.Value) -> some View {
```
stackotter: Remove the parameter labels to match the following method, and rename `value` to `newValue`

yqhbzvyzbw-cloud:
Environment

PR #244, opened 2025-11-06
Date picker

### LINKED ISSUE #233: Date Picker

bbrk24:
Fixes #233

Each backend has its own shortcomings:
- UIKitBackend has weird availability constraints on each picker style, doesn't support `hourMinuteAndSecond`, and doesn't respect font color
  - Furthermore, UIDatePicker doesn't exist on tvOS 
- AppKitBackend doesn't support the `wheel` picker style
- WinUIBackend  doesn't support `hourMinuteAndSecond`, doesn't respect font color, and doesn't respect `environment.timeZone` (instead always using the current time zone)
- GtkBackend only supports the `graphical` picker style, doesn't respect `environment.timeZone` (instead always using the current time zone), and only supports picking the date, not the time
- Gtk3Backend is unimplemented as I have no way to test it.

### DIFF
--- a/Sources/AppKitBackend/AppKitBackend.swift
+++ b/Sources/AppKitBackend/AppKitBackend.swift
@@ -168,6 +168,8 @@ public final class AppKitBackend: AppBackend {
                     renderedItem.target = wrappedAction
 
                     return renderedItem
+                case .separator:
+                    return NSCustomMenuItem.separator()
                 case .submenu(let submenu):
                     return renderSubmenu(submenu)
             }
@@ -404,53 +406,46 @@ public final class AppKitBackend: AppBackend {
 
     public func show(widget: Widget) {}
 
-    class NSContainerView: NSView {
-        var children: [NSView] = []
-
-        override func addSubview(_ view: NSView) {
-            children.append(view)
-            super.addSubview(view)
-        }
-
-        func removeSubview(_ view: NSView) {
-            children.removeAll { other in
-                view === other
-            }
-            view.removeFromSuperview()
-        }
-
-        func removeAllSubviews() {
-            for child in children {
-                child.removeFromSuperview()
-            }
-            children = []
-        }
-    }
-
     public func createContainer() -> Widget {
-        let container = NSContainerView()
+        let container = NSView()
         container.translatesAutoresizingMaskIntoConstraints = false
         return container
     }
 
     public func removeAllChildren(of container: Widget) {
-        let container = container as! NSContainerView
-        container.removeAllSubviews()
+        container.subviews = []
     }
 
-    public func addChild(_ child: Widget, to container: Widget) {
-        container.addSubview(child)
+    public func insert(_ child: Widget, into container: Widget, at index: Int) {
+        container.subviews.insert(child, at: index)
         child.translatesAutoresizingMaskIntoConstraints = false
     }
 
-    public func setPosition(ofChildAt index: Int, in container: Widget, to position: SIMD2<Int>) {
-        let container = container as! NSContainerView
-        guard container.children.indices.contains(index) else {
-            logger.warning("attempted to set position of non-existent container child")
-            return
-        }
+    public func swap(childAt firstIndex: Int, withChildAt secondIndex: Int, in container: NSView) {
+        assert(
+            container.subviews.indices.contains(firstIndex)
+            && container.subviews.indices.contains(secondIndex),
+            """
+            attempted to swap container child out of bounds; container count \
+            = \(container.subviews.count); firstIndex = \(firstIndex); \
+            secondIndex = \(secondIndex)
+            """
+        )
 
-        let child = container.children[index]
+        container.subviews.swapAt(firstIndex, secondIndex)
+    }
+
+    public func setPosition(ofChildAt index: Int, in container: Widget, to position: SIMD2<Int>) {
+        assert(
+            container.subviews.indices.contains(index),
+            """
+            attempted to set position of non-existent container child; container \
+            count = \(container.subviews.count); index = \(index); position = \
+            \(position)
+            """
+        )
+
+        let child = container.subviews[index]
 
         var foundConstraint = false
         for constraint in container.constraints {
@@ -489,9 +484,8 @@ public final class AppKitBackend: AppBackend {
         }
     }
 
-    public func removeChild(_ child: Widget, from container: Widget) {
-        let container = container as! NSContainerView
-        container.removeSubview(child)
+    public func remove(childAt index: Int, from container: Widget) {
+        container.subviews.remove(at: index)
     }
 
     public func createColorableRectangle() -> Widget {
--- a/Sources/GtkBackend/GtkBackend.swift
+++ b/Sources/GtkBackend/GtkBackend.swift
@@ -286,7 +286,9 @@ public final class GtkBackend: AppBackend {
         actionNamespace: String,
         actionPrefix: String?
     ) -> GMenu {
-        let model = GMenu()
+        var currentSection = GMenu()
+        var previousSections: [GMenu] = []
+
         for (i, item) in menu.items.enumerated() {
             let actionName =
                 if let actionPrefix {
@@ -301,15 +303,27 @@ public final class GtkBackend: AppBackend {
                         actionMap.addAction(GSimpleAction(name: actionName, action: action))
                     }
 
-                    model.appendItem(label: label, actionName: "\(actionNamespace).\(actionName)")
+                    currentSection.appendItem(
+                        label: label,
+                        actionName: "\(actionNamespace).\(actionName)"
+                    )
                 case .toggle(let label, let value, let onChange):
                     actionMap.addAction(
                         GSimpleAction(name: actionName, state: value, action: onChange)
                     )
 
-                    model.appendItem(label: label, actionName: "\(actionNamespace).\(actionName)")
+                    currentSection.appendItem(
+                        label: label,
+                        actionName: "\(actionNamespace).\(actionName)"
+                    )
+                case .separator:
+                    // GTK[3] doesn't have explicit separators per se, but instead deals with
+                    // sections (actually quite similar to what you can do in SwiftUI with the
+                    // Section view). It'll automatically draw separators between sections.
+                    previousSections.append(currentSection)
+                    currentSection = GMenu()
                 case .submenu(let submenu):
-                    model.appendSubmenu(
+                    currentSection.appendSubmenu(
                         label: submenu.label,
                         content: renderMenu(
                             submenu.content,
@@ -320,7 +334,17 @@ public final class GtkBackend: AppBackend {
                     )
             }
         }
-        return model
+
+        if previousSections.isEmpty {
+            // There are no dividers; just return the current section to keep the menu tree flat.
+            return currentSection
+        } else {
+            let model = GMenu()
+            for section in previousSections + [currentSection] {
+                model.appendSection(label: nil, content: section)
+            }
+            return model
+        }
     }
 
     private func renderMenuBar(_ submenus: [ResolvedMenu.Submenu]) -> GMenu {
@@ -457,9 +481,9 @@ public final class GtkBackend: AppBackend {
         container.removeAllChildren()
     }
 
-    public func addChild(_ child: Widget, to container: Widget) {
+    public func insert(_ child: Widget, into container: Widget, at index: Int) {
         let container = container as! Fixed
-        container.put(child, x: 0, y: 0)
+        container.put(child, index: index, x: 0, y: 0)
     }
 
     public func setPosition(ofChildAt index: Int, in container: Widget, to position: SIMD2<Int>) {
@@ -467,9 +491,20 @@ public final class GtkBackend: AppBackend {
         container.move(container.children[index], x: Double(position.x), y: Double(position.y))
     }
 
-    public func removeChild(_ child: Widget, from container: Widget) {
+    public func remove(childAt index: Int, from container: Widget) {
         let container = container as! Fixed
+        let child = container.children[index]
         container.remove(child)
+    }
+
+    public func swap(childAt firstIndex: Int, withChildAt secondIndex: Int, in container: Widget) {
+        // Gtk.Fixed doesn't let us rearrange children, so we just swap them in
+        // our own list so that at least everything works on the SCUI side. The
+        // only side effect of this approach is that overlapping widgets may
+        // end up with unexpected z ordering. If that becomes an issue we may
+        // have to make a custom replacement for Gtk.Fixed.
+        let container = container as! Fixed
+        container.children.swapAt(firstIndex, secondIndex)
     }
 
     public func createColorableRectangle() -> Widget {
--- a/Sources/SwiftCrossUI/Backend/AppBackend.swift
+++ b/Sources/SwiftCrossUI/Backend/AppBackend.swift
@@ -246,12 +246,15 @@ public protocol AppBackend: Sendable {
     func createContainer() -> Widget
     /// Removes all children of the given container.
     func removeAllChildren(of container: Widget)
-    /// Adds a child to a given container at an exact position.
-    func addChild(_ child: Widget, to container: Widget)
+    /// Inserts a child into a given container at a given index.
+    func insert(_ child: Widget, into container: Widget, at index: Int)
+    /// Swaps the child at firstIndex with the child at secondIndex. May crash if either
+    /// index is out of bounds.
+    func swap(childAt firstIndex: Int, withChildAt secondIndex: Int, in container: Widget)
     /// Sets the position of the specified child in a container.
     func setPosition(ofChildAt index: Int, in container: Widget, to position: SIMD2<Int>)
-    /// Removes a child widget from a container (if the child is a direct child of the container).
-    func removeChild(_ child: Widget, from container: Widget)
+    /// Removes the child at the given index from the given container.
+    func remove(childAt index: Int, from container: Widget)
 
     /// Creates a rectangular widget with configurable color.
     func createColorableRectangle() -> Widget
--- a/Sources/WinUIBackend/WinUIBackend.swift
+++ b/Sources/WinUIBackend/WinUIBackend.swift
@@ -279,6 +279,8 @@ public final class WinUIBackend: AppBackend {
                         onChange(widget.isChecked)
                     }
                     return widget
+                case .separator:
+                    return MenuFlyoutSeparator()
                 case .submenu(let submenu):
                     let widget = MenuFlyoutSubItem()
                     widget.text = submenu.label
@@ -358,9 +360,28 @@ public final class WinUIBackend: AppBackend {
         container.children.clear()
     }
 
-    public func addChild(_ child: Widget, to container: Widget) {
+    public func insert(_ child: Widget, into container: Widget, at index: Int) {
         let container = container as! Canvas
-        container.children.append(child)
+        container.children.insertAt(UInt32(index), child)
+    }
+
+    public func swap(childAt firstIndex: Int, withChildAt secondIndex: Int, in container: Widget) {
+        // TODO: Find out if there's an efficient way to do this without WinUI
+        //   getting annoyed at us for having the same element in the list twice.
+        let container = container as! Canvas
+        let largerIndex = UInt32(max(firstIndex, secondIndex))
+        let smallerIndex = UInt32(min(firstIndex, secondIndex))
+        let element1 = container.children[Int(smallerIndex)]
+        let element2 = container.children[Int(largerIndex)]
+        container.children.removeAt(largerIndex)
+        container.children.removeAt(smallerIndex)
+        container.children.insertAt(smallerIndex, element2)
+        container.children.insertAt(largerIndex, element1)
+    }
+
+    public func remove(childAt index: Int, from container: Widget) {
+        let container = container as! Canvas
+        container.children.removeAt(UInt32(index))
     }
 
     public func setPosition(ofChildAt index: Int, in container: Widget, to position: SIMD2<Int>) {
@@ -372,19 +393,6 @@ public final class WinUIBackend: AppBackend {
 
         Canvas.setTop(child, Double(position.y))
         Canvas.setLeft(child, Double(position.x))
-    }
-
-    public func removeChild(_ child: Widget, from container: Widget) {
-        let container = container as! Canvas
-        let count = container.children.size
-        for index in 0..<count {
-            if container.children.getAt(index) == child {
-                container.children.removeAt(index)
-                return
-            }
-        }
-
-        logger.warning("child to remove not found")
     }
 
     public func createColorableRectangle() -> Widget {
@@ -1411,7 +1419,7 @@ public final class WinUIBackend: AppBackend {
             fatalError("Unsupported gesture type \(gesture)")
         }
         let tapGestureTarget = TapGestureTarget()
-        addChild(child, to: tapGestureTarget)
+        insert(child, into: tapGestureTarget, at: 0)
         tapGestureTarget.child = child
 
         // Set a background so that the click target's entire area gets hit
@@ -1445,7 +1453,7 @@ public final class WinUIBackend: AppBackend {
 
     public func createHoverTarget(wrapping child: Widget) -> Widget {
         let hoverTarget = HoverGestureTarget()
-        addChild(child, to: hoverTarget)
+        insert(child, into: hoverTarget, at: 0)
         hoverTarget.child = child
 
         // Ensure the hover target covers the full area of the child.
--- a/Examples/Sources/ForEachExample/ForEachApp.swift
+++ b/Examples/Sources/ForEachExample/ForEachApp.swift
@@ -42,23 +42,24 @@ struct ForEachApp: App {
                     .padding(10)
 
                     ScrollView {
-                        ForEach(Array(items.enumerated()), id: \.element.id) { (index, item) in
-                            ItemRow(
-                                item: item,
-                                isFirst: index == 0,
-                                isLast: index == items.count - 1
-                            ) {
-                                items.remove(at: index)
-                            } moveUp: {
-                                guard index != items.startIndex else { return }
-                                items.swapAt(index, index - 1)
-                            } moveDown: {
-                                guard index != items.endIndex else { return }
-                                items.swapAt(index, index + 1)
+                        VStack(alignment: .trailing) {
+                            ForEach(Array(items.enumerated()), id: \.element.id) { (index, item) in
+                                ItemRow(
+                                    item: $items[index],
+                                    isFirst: index == 0,
+                                    isLast: index == items.count - 1
+                                ) {
+                                    items.remove(at: index)
+                                } moveUp: {
+                                    guard index != items.startIndex else { return }
+                                    items.swapAt(index, index - 1)
+                                } moveDown: {
+                                    guard index != items.endIndex else { return }
+                                    items.swapAt(index, index + 1)
+                                }
                             }
                         }
-                        .frame(maxWidth: .infinity)
-                    }
+                    }.padding()
                 }
             }
         }
@@ -67,7 +68,11 @@ struct ForEachApp: App {
 }
 
 struct ItemRow: View {
-    var item: Item
+    @Binding var item: Item
+
+    @State var isEditing = false
+    @State var value = ""
+
     let isFirst: Bool
     let isLast: Bool
     var remove: () -> Void
@@ -76,7 +81,26 @@ struct ItemRow: View {
 
     var body: some View {
         HStack {
-            Text(item.value)
+            if isEditing {
+                TextField("Value", text: $value)
+                    .frame(width: 100)
+                Spacer()
+                Button("Save") {
+                    item.value = value
+                    isEditing = false
+                }
+                Button("Cancel") {
+                    isEditing = false
+                }
+            } else {
+                Text(item.value)
+                    .frame(width: 100, alignment: .leading)
+                Spacer()
+                Button("Edit") {
+                    isEditing = true
+                    value = item.value
+                }
+            }
             Button("Delete") { remove() }
             Button("⌃") { moveUp() }
                 .disabled(isFirst)
--- a/Examples/Sources/WindowingExample/WindowingApp.swift
+++ b/Examples/Sources/WindowingExample/WindowingApp.swift
@@ -208,8 +208,9 @@ struct WindowingApp: App {
         .commands {
             CommandMenu("Demo menu") {
                 Button("Menu item") {}
-
                 Toggle("Toggle", active: $toggle)
+
+                Divider()
 
                 Menu("Submenu") {
                     Button("Item 1") {}
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,5 +1,5 @@
 {
-  "originHash" : "dd3bde13d6de041e108582ce8b4dc7c7553f4468dd6cbc495b986a1bddeaeebd",
+  "originHash" : "d67afda47318a1da55f159038bea8c95778993612dfd2e6dc54b4c35e9f98151",
   "pins" : [
     {
       "identity" : "jpeg",
@@ -44,15 +44,6 @@
       "state" : {
         "revision" : "45a7b500c24e7243da7b02e19cf436d0107a0c1b",
         "version" : "0.2.0"
-      }
-    },
-    {
-      "identity" : "swift-collections",
-      "kind" : "remoteSourceControl",
-      "location" : "https://github.com/apple/swift-collections.git",
-      "state" : {
-        "revision" : "8c0c0a8b49e080e54e5e328cc552821ff07cd341",
-        "version" : "1.2.1"
       }
     },
     {
--- a/Package.swift
+++ b/Package.swift
@@ -123,10 +123,6 @@ let package = Package(
             revision: "42c47f4e4129c8b5a5d9912f05e1168c924ac180"
         ),
         .package(
-            url: "https://github.com/apple/swift-collections.git",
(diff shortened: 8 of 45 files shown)


bbrk24:
The thought just occurred to me to observe [NSSystemTimeZoneDidChange](https://developer.apple.com/documentation/foundation/nsnotification/name-swift.struct/nssystemtimezonedidchange) to properly redraw views that use `@Environment(\.timeZone)`, but I'm unsure how that would work, and regardless I don't see an equivalent for the calendar.

stackotter:
There’s a backend method called setEnvironmentChangeHandler or something like that. That’s where we currently observe system theme changes and would be where timezone observation should live as well.

bbrk24:
Right, I'll go see if I can get that to work in UIKitBackend and AppKitBackend first, and then maybe the other backends. I don't think it'll necessarily work on Linux though

bbrk24:
Marking this as draft because UIDatePicker (at least on iOS 17) completely misbehaves if the user changes the time zone in settings

Examples/Sources/DatePickerExample/DatePickerApp.swift
```diff
@@ -0,0 +1,53 @@
...
+                #if os(iOS) || os(visionOS) || canImport(WinUIBackend)
+                    allStyles.append(.wheel)
+                #endif
+            }
+        #endif
+    }
+
+    var body: some Scene {
+        WindowGroup("Date Picker") {
+            VStack {
```
stackotter: Add `#hotReloadable { ... }` around this `VStack`

Examples/Sources/DatePickerExample/DatePickerApp.swift
```diff
@@ -0,0 +1,53 @@
...
+        }
+
+        #if !canImport(GtkBackend)
+            if #available(iOS 13.4, macCatalyst 13.4, *) {
+                allStyles.append(.compact)
+                #if os(iOS) || os(visionOS) || canImport(WinUIBackend)
+                    allStyles.append(.wheel)
+                #endif
+            }
+        #endif
```
stackotter: Given that the differences in platform support for each style are kinda complicated, it could be a good idea to add a `supportedDataPickerStyles` property to the environment so that devs can implement runtime compatibility checks a bit more easily than this. The environment variable should be a get-only property with a getter that references some internal property that SwiftCrossUI can write to. Then when creating the default environment values, SwiftCrossUI should query the backend for its list of supported date picker styles (via a new `supportedDatePickerStyles` property of some sort).

Examples/Sources/DatePickerExample/DatePickerApp.swift
```diff
@@ -0,0 +1,53 @@
...
+
+                Picker(of: allStyles, selection: $style)
+
+                DatePicker(
+                    "Test Picker",
+                    selection: $date
+                )
+                .datePickerStyle(style ?? .automatic)
+
+                Button("Reset date") {
```
stackotter: `"Reset date to now"`

Sources/SwiftCrossUI/Environment/EnvironmentValues.swift
```diff
@@ -190,6 +190,16 @@ public struct EnvironmentValues {
...
     }
 
+    /// The current calendar that views should use when handling dates.
+    public var calendar: Calendar
+
+    /// The current time zone that views should use when handling dates.
+    public var timeZone: TimeZone
+
+    #if !os(tvOS)
+        public var datePickerStyle: DatePickerStyle
```
stackotter: Document this property.

Sources/AppKitBackend/AppKitBackend.swift
```diff
@@ -1689,6 +1697,80 @@ public final class AppKitBackend: AppBackend {
...
+        let datePicker = CustomDatePicker()
+        datePicker.delegate = datePicker.strongDelegate
+        return datePicker
+    }
+
+    // Depending on the calendar, era is either necessary or must be omitted. Making the wrong
+    // choice for the current calendar means the cursor position is reset after every keystroke. I
+    // know of no simple way to tell whether NSDatePicker requires or forbids eras for a given
+    // calendar, so in lieu of that I have hardcoded the calendar identifiers.
+    private let calendarsWithEras: Set<Calendar.Identifier> = [
```
stackotter: According to this thread, the Gregorian calendar also has eras: https://www.reddit.com/r/iOSProgramming/comments/7brq81/what_is_the_importance_of_the_calendar_component/

It could make more sense to name this property to `calendarsRequiringEra` or something?

Sources/AppKitBackend/AppKitBackend.swift
```diff
@@ -1689,6 +1697,80 @@ public final class AppKitBackend: AppBackend {
...
+            elementFlags.insert(.yearMonthDay)
+            if calendarsWithEras.contains(environment.calendar.identifier) {
+                elementFlags.insert(.era)
+            }
+        }
+        if components.contains(.hourMinuteAndSecond) {
+            elementFlags.insert(.hourMinuteSecond)
+        } else {
+            elementFlags.insert(.hourMinute)
+        }
```
stackotter, line 1896: Does AppKitBackend not support date-only pickers?

Sources/Gtk/Utility/GDateTime.swift
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    deinit {
+        g_date_time_unref(pointer)
+    }
+
+    public func toDate() -> Date {
+        let offset = g_date_time_to_unix(pointer)
+        return Date(timeIntervalSince1970: Double(offset))
+    }
```
stackotter, line 56: Do these unix-based conversions correctly handle historical dates? The Gtk docs aren't super clear about whether they support negative unix epochs.

Sources/Gtk/Utility/GDateTime.swift
```diff
@@ -0,0 +1,55 @@
...
+
+    public init?(_ pointer: OpaquePointer?) {
+        guard let pointer else { return nil }
+        self.pointer = pointer
+    }
+
+    public convenience init?(unixEpoch: TimeInterval) {
+        // g_date_time_new_from_unix_utc_usec appears to be too new
+        self.init(g_date_time_new_from_unix_utc(gint64(unixEpoch)))
+    }
```
stackotter: This should take an `Int` given that it discards everything below seconds

Sources/Gtk/Utility/GTimeZone.swift
```diff
@@ -0,0 +1,19 @@
...
+    }
+
+    public convenience init?(_ timeZone: TimeZone) {
+        self.init(identifier: timeZone.identifier)
+    }
+
+    deinit {
+        g_time_zone_unref(pointer)
+    }
+}
```
stackotter, line 19: Do Gtk timezone identifiers line up with Foundation's timezone identifiers. From a quick look at the docs it seems like Foundation's timezone identifiers don't really follow the same format. It's possible that this could be the reason behind the GtkBackend implementation appearing to not support timezones?

Sources/SwiftCrossUI/Backend/AppBackend.swift
```diff
@@ -538,6 +538,19 @@ public protocol AppBackend: Sendable {
     /// Sets the index of the selected option of a picker.
     func setSelectedOption(ofPicker picker: Widget, to selectedOption: Int?)
 
+    func createDatePicker() -> Widget
+
+    #if !os(tvOS)
```
stackotter: AppBackend methods shouldn't be conditional on platform. I think it'd be better to include this on tvOS and just not implement it (or implement it with a fatalError and a descriptive error message). Is there a reason that this method wouldn't be supported on tvOS but createDatePicker would?

Sources/SwiftCrossUI/Views/Modifiers/DatePickerStyleModifier.swift
```diff
@@ -0,0 +1,13 @@
...
+    public func datePickerStyle(_ style: DatePickerStyle) -> some View {
+        #if os(tvOS)
+            assertionFailure()
+            return EmptyView()
+        #else
+            EnvironmentModifier(self) { environment in
+                environment.with(\.datePickerStyle, style)
+            }
+        #endif
+    }
```
stackotter: I think it's best to keep this modifier and the underlying environment property available everywhere, because it's harmless to read/write the current datePickerStyle if you can't make a date picker. It reduces the amount of conditional compilation people have to use, without introducing additional unsafety.

Sources/UIKitBackend/UIKitBackend+Control.swift
```diff
@@ -239,6 +239,26 @@ final class SliderWidget: WrapperWidget<UISlider> {
     }
 }
 
+@available(tvOS, unavailable)
+final class DatePickerWidget: WrapperWidget<UIDatePicker> {
+    var onChange: ((Date) -> Void)? {
+        didSet {
+            if oldValue == nil {
+                child.addTarget(self, action: #selector(dateChanged), for: .valueChanged)
+            }
```
stackotter, line 248: I think this should be performed on initialisation of the view (`loadView` iirc?), because otherwise setting `onChange` to nil and then back to a value again would cause `dateChanged` to start triggering twice unless `addTarget` does some deduplication. Either way, we can assume that under UIKitBackend's usage, `onChange` will always get set to a value, so there's no harm in setting up the observation prematurely.

Sources/GtkBackend/GtkBackend.swift
```diff
@@ -1581,3 +1612,169 @@ extension UnsafeMutablePointer {
...
+    }
+
+    private static func transformToRange(_ value: Int, hourCycle: Locale.HourCycle) -> Int {
+        switch hourCycle {
+            case .zeroToEleven: value % 12
+            case .oneToTwelve: (value + 11) % 12 + 1
+            case .zeroToTwentyThree: value % 24
+            case .oneToTwentyFour: (value + 23) % 24 + 1
+            #if os(macOS)
+                @unknown default: fatalError()
```
stackotter: Add a custom error message

Sources/GtkBackend/GtkBackend.swift
```diff
@@ -1581,3 +1612,169 @@ extension UnsafeMutablePointer {
...
+        } else {
+            if let amPmPicker {
+                remove(amPmPicker)
+                self.amPmPicker = nil
+            }
+        }
+
+        hourPicker.text =
+            "\(TimePicker.transformToRange(components.hour!, hourCycle: self.hourCycle))"
+        hourPicker.valueChanged = { [unowned self] hourPicker in
```
stackotter, line 1967: Why use `unowned` instead of `weak` here? I've never used `unowned` because in objc it's quite unsafe, but after some googling it seems to be safe in Swift by default, and it's just basically a pre-force-unwrapped version of `weak`?

Sources/GtkBackend/GtkBackend.swift
```diff
@@ -1581,3 +1612,169 @@ extension UnsafeMutablePointer {
...
+            guard let value = Int(exactly: hourPicker.value),
+                let newDate = calendar.date(bySetting: .hour, value: value, of: date)
+            else {
```
stackotter, line 1970: I assume that this isn't meant to fail? If so, update it to print out a suitable warning when it early exits (with some info such as the value of the hour picker and the date).

Sources/GtkBackend/GtkBackend.swift
```diff
@@ -1581,3 +1612,169 @@ extension UnsafeMutablePointer {
...
+        hourPicker.setRange(
+            min: effectiveHours?.min().map(Double.init(_:))
+                ?? TimePicker.minHour(for: self.hourCycle),
+            max: effectiveHours?.max().map(Double.init(_:))
+                ?? TimePicker.maxHour(for: self.hourCycle)
+        )
+
+        if self.hourCycle == .oneToTwelve || self.hourCycle == .zeroToEleven {
+            if let amPmPicker {
+                // update strings if necessary
```
stackotter, line 1953: Is this a todo?

Sources/GtkBackend/GtkBackend.swift
```diff
@@ -1500,6 +1500,37 @@ public final class GtkBackend: AppBackend {
...
+            print("Warning: time picker is unimplemented on GtkBackend")
+        }
+        if environment.datePickerStyle != .automatic && environment.datePickerStyle != .graphical {
+            print("Warning: only DatePickerStyle.graphical is implemented in GtkBackend")
+        }
+
+        let calendarWidget = datePicker as! Gtk.Calendar
+        calendarWidget.date = date
+        calendarWidget.daySelected = { calendarWidget in
+            onChange(calendarWidget.date)
```
stackotter: Could `range` be supported by clamping it in this signal handler and setting the picker's date to the clamped value if it differs from the current value? It'd be a little janky, but it's probably better than not supporting it at all for the time being.

Sources/SwiftCrossUI/Views/DatePicker.swift
```diff
@@ -0,0 +1,152 @@
...
+        self.rawValue = 0
+    }
+
+    public static let date = DatePickerComponents(rawValue: 0x1C)
+    public static let hourAndMinute = DatePickerComponents(rawValue: 0x60)
+
+    @available(iOS, unavailable)
+    @available(visionOS, unavailable)
+    @available(macCatalyst, unavailable)
+    public static let hourMinuteAndSecond = DatePickerComponents(rawValue: 0xE0)
```
stackotter, line 32: Add a command somewhere here that explains that these magic raw values came from SwiftUI, and that hourMinuteAndSecond contains hourAndMinute by design.

Sources/SwiftCrossUI/Views/DatePicker.swift
```diff
@@ -0,0 +1,152 @@
...
+}
+
+@available(tvOS, unavailable)
+public enum DatePickerStyle: Sendable, Hashable {
+    /// A date input chosen by the backend.
```
stackotter: `A date input that adapts to the current platform and context.`

Sources/SwiftCrossUI/Views/DatePicker.swift
```diff
@@ -0,0 +1,152 @@
...
+    private var components: DatePickerComponents
+    private var style: DatePickerStyle = .automatic
+
+    /// Displays a date input.
+    /// - Parameters:
+    ///   - selection: The currently-selected date.
+    ///   - range: The range of dates to display. The backend takes this as a hint but it is not
+    ///     necessarily enforced. As such this parameter should be treated as an aid to validation
+    ///     rather than a replacement for it.
+    ///   - displayedComponents: What parts of the date/time to display in the input.
```
stackotter: I think that `Which parts...` reads a bit better

Sources/SwiftCrossUI/Views/DatePicker.swift
```diff
@@ -0,0 +1,152 @@
...
+    ///   - label: The view to be shown next to the date input.
+    public nonisolated init(
+        selection: Binding<Date>,
+        range: ClosedRange<Date> = Date.distantPast...Date.distantFuture,
```
stackotter: Label this parameter `in` to match SwiftUI. (same with the alternative init below)

Sources/SwiftCrossUI/Views/DatePicker.swift
```diff
@@ -0,0 +1,152 @@
...
+    func update<Backend: AppBackend>(
+        _ widget: Backend.Widget,
+        proposedSize: SIMD2<Int>,
+        environment: EnvironmentValues,
+        backend: Backend,
+        dryRun: Bool
+    ) -> ViewUpdateResult {
+        #if os(tvOS)
+            preconditionFailure()
+        #else
```
stackotter: With my comment about making `updateDatePicker` available on tvOS, I think this conditional compilation should be made redundant.

Sources/UIKitBackend/UIKitBackend.swift
```diff
@@ -87,6 +89,7 @@ public final class UIKitBackend: AppBackend {
         var environment = defaultEnvironment
 
         environment.toggleStyle = .switch
+        environment.timeZone = .current
```
stackotter, line 135: It seems like AppKitBackend would also prefer that timeZone is set to `.current`. Should AppKitBackend be doing that? Or maybe should EnvironmentValues be doing that? I can see reasons for and against;

- If the default was current, then backends that don't implement timezone change observation (such as GtkBackend) would get out of sync
- If the default was autoupdatingCurrent, then whenever you accessed the environment property you would technically get the current timezone, but it wouldn't capture the current timezone in the way that you'd expect cause the value's meaning could change out from under you.

Now that I think about it, having it be autoupdatingCurrent could cause issues further down the line when we do more surgical propagation of environment changes. From SwiftCrossUI's perspective, the old timezone and the new one would always seem to be identical.

Sources/WinUIBackend/WinUIBackend.swift
```diff
@@ -1704,6 +1725,57 @@ public final class WinUIBackend: AppBackend {
...
+
+        if components.contains(.hourMinuteAndSecond) {
+            print(
+                "DatePickerComponents.hourMinuteAndSecond is not supported in WinUIBackend. Falling back to .hourAndMinute."
+            )
+        }
+
+        customDatePicker.toggleTimeView(shown: components.contains(.hourAndMinute))
+
+        if environment.timeZone != .autoupdatingCurrent {
```
stackotter: This would have to be updated to `current` if the environment timezone default value gets changed to `.current` (as discussed in one of my comments on UIKitBackend.

Sources/WinUIBackend/WinUIBackend.swift
```diff
@@ -1940,3 +2012,276 @@ public final class GeometryGroupHolder {
...
+            case .hebrew: "HebrewCalendar"
+            case .islamicTabular: "HijriCalendar"
+            case .islamicUmmAlQura: "UmAlQuraCalendar"
+            case .japanese: "JapaneseCalendar"
+            case .persian: "PersianCalendar"
+            case .republicOfChina: "TaiwanCalendar"
+            #if compiler(>=6.2)
+                case .vietnamese: "VietnameseLunarCalendar"
+            #endif
+            case let id: fatalError("Unsupported calendar identifier \(id)")
```
stackotter: I feel like this fatalError would be relatively easy to come across when trying to make an app support many locales and calendars. It'd be annoying to find out by getting a bug report from someone softlocked out of using the app because of their locale, so I think it'd be best to print a warning here and fallback on gregorian. That way people can at least use the app even if they try to intentionally (or unintentionally via locale) use an unsupported calendar.

Examples/Sources/DatePickerExample/DatePickerApp.swift
bbrk24: Fair enough, this button was a last-minute addition when I got lost using calendars and time zones

Sources/AppKitBackend/AppKitBackend.swift
bbrk24: Yeah that's why I don't just check the length of the eras property on the calendar object. Gregorian has two (AD/BC or CE/BCE depending on your preference), and some of the ones that require it only have one. The suggested rename makes sense to me.

Sources/AppKitBackend/AppKitBackend.swift
bbrk24: I will have to double-check as there's a nonzero chance I just forgot to handle that case. But I think it does at least for .graphical, even if not for .compact.

Sources/Gtk/Utility/GDateTime.swift
bbrk24: image

Appears to!

Sources/Gtk/Utility/GDateTime.swift
bbrk24: I think the entire reason I use TimeInterval here is to make `init!(_ date: Date)` simpler. I can update this.

Sources/SwiftCrossUI/Backend/AppBackend.swift
bbrk24: The problem is that the types involved in the method signature are `@available(tvOS, unavailable)`, and you can't put availability annotations on a non-ObjC protocol. `#if` here was me giving up.

Sources/SwiftCrossUI/Views/Modifiers/DatePickerStyleModifier.swift
bbrk24: I guess that makes sense. Should the individual enum members still have availability for iOS and Catalyst?

Sources/UIKitBackend/UIKitBackend+Control.swift
bbrk24: I only used `didSet` here to mirror existing code in `ButtonWidget`, `TappableWidget`, etc. Seeing as how the `onChange` parameter for `updateDatePicker` isn't optional, I don't think it can ever be set back to nil.

Sources/GtkBackend/GtkBackend.swift
bbrk24: This entire class is not really complete, but I can do that for the `fatalError`s I guess.

Sources/GtkBackend/GtkBackend.swift
bbrk24: `unowned` is also faster under high contention -- converting a weak reference to a strong one needs to do more work than converting an unowned reference to a strong one, and refcounts are synchronized. I was following my own advice from [this forum post](https://forums.swift.org/t/do-you-need-weak-capture-in-escaping-closure/64869/2):

> If the closure won’t outlive `self`, but a strong capture would create a circular reference, use `[unowned self]`.

The logic being that I saw no way for `hourPicker` to outlive `self`, and `hourPicker.valueChanged` shouldn't outlive `hourPicker`.

Sources/GtkBackend/GtkBackend.swift
bbrk24: Yes, again this entire class is half-baked.

Sources/SwiftCrossUI/Views/DatePicker.swift
bbrk24: Fun fact: this is actually a bitfield! The values in binary are:
```
date 00011100
h&m  01100000
hms  11100000
     smhdMy--
```

I should probably mention that in the comment too.

Sources/UIKitBackend/UIKitBackend.swift
bbrk24: The entire extent of my logic was that I booted up a Swift REPL and ran
```swift
import SwiftUI
EnvironmentValues().timeZone == .autoupdatingCurrent
```
And seeing that that's `true` (and knowing `.autoupdatingCurrent != .current`), I put that as the default in SwiftCrossUI.

Sources/WinUIBackend/WinUIBackend.swift
bbrk24: That makes sense to me.

Examples/Sources/DatePickerExample/DatePickerApp.swift
bbrk24: Makes sense to me, will implement

Examples/Sources/DatePickerExample/DatePickerApp.swift
bbrk24: Added

Sources/SwiftCrossUI/Environment/EnvironmentValues.swift
bbrk24: Added documentation

Sources/Gtk/Utility/GTimeZone.swift
bbrk24: Oh you're right, I kind of just assumed it was IANA tz db identifiers. I'll look more into this.

bbrk24: No, this isn't it. First of all, new_identifier does support IANA tz db identifiers, as it supports the `TZ` environment variable formats which includes the "geographical format". But furthermore I don't even appear to use it: `GTimeZone` is only used by the convenience init on `GDateTime`, and that convenience init is currently unused. That's the real reason timezones aren't supported: I always call `g_date_time_new_from_unix_utc`, which uses, well, UTC.

Sources/SwiftCrossUI/Views/DatePicker.swift
bbrk24: Forgot to say anything but I already did this

Sources/SwiftCrossUI/Views/DatePicker.swift
bbrk24: Same here, already done

Sources/SwiftCrossUI/Views/DatePicker.swift
bbrk24: Updated

Sources/SwiftCrossUI/Views/DatePicker.swift
bbrk24: Updated

Sources/AppKitBackend/AppKitBackend.swift
bbrk24: It does support date-only pickers, updated this

Sources/UIKitBackend/UIKitBackend.swift
bbrk24: Weirdly, UIKitBackend still behaves fine if I change the default values from `.autoupdatingCurrent` to `.current` in EnvironmentValues.swift... but not if I remove the part where I set it to `.current` here. Heck if I know.

Sources/SwiftCrossUI/Backend/AppBackend.swift
stackotter: If you make DatePickerComponents available on tvOS would it be possible? I feel like there's not an issue with making that available.

Sources/SwiftCrossUI/Environment/EnvironmentValues.swift
```diff
@@ -195,6 +195,18 @@ public struct EnvironmentValues {
...
+    /// The current calendar that views should use when handling dates.
+    public var calendar: Calendar
+
+    /// The current time zone that views should use when handling dates.
+    public var timeZone: TimeZone
+
+    /// The display style used by ``DatePicker``.
+    public var datePickerStyle: DatePickerStyle
+
+    /// The display styles supported by ``DatePicker``. ``datePickerStyle`` must be one of these.
```
stackotter, line 219: The second sentence isn't true cause under Gtk3Backend the supported styles list is empty, but the datePickerStyle is `.automatic` by default (which isn't in the list).

Maybe automatic should always be allowed even if the backend doesn't state so (cause the date picker style is kinda inconsequential if DatePicker doesn't work anyway). Another reason for doing so is that it feels like `.datePickerStyle(environment.datePickerStyle)` should probably be allowed in all circumstances, cause it's kinda an identity operation (and allowing `automatic` in all circumstances would make it act like one).

Sources/SwiftCrossUI/Views/DatePicker.swift
stackotter: GitHub is showing conditional compilation here, but I am trying the beta pr review layout, so not sure whether it's telling the truth or not.

Sources/Gtk/Utility/GDateTime.swift
```diff
@@ -0,0 +1,55 @@
...
+                gint(month),
+                gint(day),
+                gint(hour),
+                gint(minute),
+                second
+            )
+        )
+    }
+
+    public convenience init!(_ date: Date) {
```
stackotter, line 45: Document that this discards fractional seconds.

Sources/SwiftCrossUI/Views/Modifiers/DatePickerStyleModifier.swift
stackotter: Yeah I think it's ok to keep the enum member availability for iOS and Catalyst cause we do it in a few other places. I haven't quite decided on a policy for availability annotations yet. My hesitation towards them is because it's possible that we'll want to make a custom rendering backend at some point, and it may not have the same feature availability as the existing UIKitBackend.

Sources/UIKitBackend/UIKitBackend+Control.swift
stackotter: Yeah, I guess as long as it stays an internal detail then that's a reasonable assumption. And given the other views do it it does make sense to do it here too.

stackotter:
I think I've gone through and closed all of the review comments that have been addressed and commented on ones that I have more questions about.

Regarding the GtkBackend TimePicker implementation: when I originally reviewed it I didn't realise that it's unused. If you'd like, feel free to turn my current comments on it into todo comments and document clearly that the class is neither ready nor used, and I'd count those comments as resolved. I more just want to avoid another contributor coming along and trying to use it assuming that it's meant to work properly in its current form.

Sources/SwiftCrossUI/Backend/AppBackend.swift
bbrk24: I've done that to avoid the `#if`, but it does mean that the availability annotations aren't quite consistent (cf. `hourMinuteAndSecond`) -- should I add `@available(tvOS, unavailable)` to the individual static properties?

Sources/SwiftCrossUI/Views/DatePicker.swift
bbrk24: I removed it elsewhere but not here, oops :sweat_smile: fixed

Sources/SwiftCrossUI/Backend/AppBackend.swift
stackotter: I think it's ok to leave it as is

stackotter (changes requested):
Somehow GitHub just had some of these comments pending from a while ago. I guess I must've left the review halfway through?

There's also this comment unresolved from a while ago: https://github.com/stackotter/swift-cross-ui/pull/244#discussion_r2570831922 . What do you reckon about that idea?

Looking good now though. Happy to merge once the GDateTime documentation comment is addressed and these changes have been rebased onto `main`.

stackotter:
I think you'll have to run `./generate.sh && ./format_and_lint.sh` to fix those Gtk build issues. I changed some things about the generation

Sources/GtkBackend/GtkBackend.swift
bbrk24: I had skipped over this comment because it was right after two or three other threads on the same class where I had explained this is half-baked and felt like stating it again would be repeating myself too much. Now that the comment above the class documents that it doesn't implement `range`, is that sufficient to resolve this thread?

stackotter: Is this related to that unused class below? I thought it was just used the built-in Gtk calendar widget which doesn't itself support ranges.

Sources/Gtk/Utility/GDateTime.swift
bbrk24: Done

bbrk24:
> Now that the DatePicker example is relatively simple, I reckon it could make sense to merge it into ControlsExample.

Done, and added a `#if` so that it can still run in Gtk3Backend (which doesn't implement pickers).

Examples/Sources/ControlsExample/ControlsApp.swift
```diff
@@ -69,14 +74,35 @@ struct ControlsApp: App {
...
+                            }
+                            Text("You chose: \(flavor ?? "Nothing yet!")")
                         }
-                        Text("You chose: \(flavor ?? "Nothing yet!")")
-                    }
+
+                        VStack {
+                            Text("Selected date: \(date)")
+
+                            Picker(of: supportedDatePickerStyles, selection: $datePickerStyle)
```
stackotter: Please put this in an HStack along with a label (like above for the flavour picker) so that it's clear what the picker is for

Examples/Sources/ControlsExample/ControlsApp.swift
```diff
@@ -69,14 +74,35 @@ struct ControlsApp: App {
...
+
+                            DatePicker(
```
stackotter: Please put the entirety of ControlsExample into a ScrollView because it's gotten pretty tall now (too tall for my laptop display). I probably should've done that a while ago

stackotter:
I've just been testing out the ControlsExample on my various machines, and noticed that the GtkBackend really spams the time picker not implemented warning. It might be worth updating it to only log that message the first time it's triggered. I believe that one of the other backends has a mechanism to do that already, maybe UIKitBackend from some of Mia's changes?

stackotter:
When I click a day in the GtkBackend Calendar picker, it selects the day prior for some reason 🤔

stackotter:
And when I click 'reset date to now', the focus ring stays on the date that I previously selected, rather than the programatically selected date. Maybe that's just a Gtk thing though?

stackotter (approved):
Thanks for your patience with this one 🙏

stackotter:
I'll test locally while waiting for CI, then merge if all goes well.

stackotter:
Damn, merging in the changes from main caused WinUIBackend compilation to fail. It appears to be because I moved the implementation of naturalSize to a static method for convenience when implementing WinUIElementRepresentable. This patch fixes things for me;

[WinUIBackendFixes.patch](https://github.com/user-attachments/files/24524213/WinUIBackendFixes.patch)

I also removed the picker's label because it was line wrapping and causing the date picker to be really tall with no visible reason when the window was skinny enough. I tried making the label fixedSize but for some reason that did nothing.

stackotter:
And while testing on Windows I found that the graphical calendar view lets you unselect your selection without selecting anything else. I think in that case we should just programatically reselect the selected day again.

stackotter (dismissed):

stackotter:
Will merge once CI finishes (it got re-triggered because I had to resolve a merge conflict)

PR #245, opened 2025-11-06
Fix ForEach Collection change rendered View correctness

MiaKoring:
This pr fixes the issue of only the last items getting removed from the rendered content, even if they werent the ones changing and item insertion/reordering correctness like mentioned in https://github.com/stackotter/swift-cross-ui/issues/243

I added apple/swift-collections as a dependency, I’m using OrderedSet in the changes.

If a duplicate identifier is detected every node gets replaced with a new one, diffing is not possible.

For unique identifiers:
- If it was included in the previous update, the node gets reused. 
It checks if there is an identifier it doesn’t know from the previous update anywhere before it, if there is, ongoing every existing node gets removed from the view graph and reinserted at its new place.
- if it wasn’t included, a new node gets created and ongoing every existing node gets readded. 

- if its the first update every node gets added to the view graph
- if there was no duplicate the nodes included in the previous update not included in the current one get removed from the view graph
- if there were duplicate(s) every node gets replaced

Also I added some new initializers: 
for Identifiable, setting the identifier key path to \.id if not specified otherwise
for existing identifier a new one allowing customization of the id KeyPath

Sadly this PR needs to be a breaking change. I was forced to add labels to the elements property on some existing initializers. Otherwise the Swift Compiler wouldn’t select the right initializer (or even select one). disfavouredOverload didn’t help. While updating you can decide between adding the label (if needed) or a keyPath. Choosing the label option uses the same update method as before. With adding a keyPath for the Identifier you choose the new, improved update method. ForEach with Range or [Identifiable] automatically recieve the new method.

I added a new Example App, ForEachExample, showcasing deletion, insertion, appending and reordering.

I tested iOS, macOS, macCatalyst, GtkBackend on Linux and WinUIBackend on Windows successfully.

While it works with non-unique Identifiers I strongly recommend using unique and constant Identifiers. It should be considerably more performant due to it making as few as possible operations on the view graph. 

In my tests it still was consistently about 11% faster than the previous implementation. 

As soon as the reason for a view update is available ForEach’s update method should be optimized to do only whats necessary. For Example the whole diffing, reordering,… is obsolete when only the size changed. For now at least the correctness got fixed and performance at least slightly improved.

### DIFF
--- a/Examples/Sources/GreetingGeneratorExample/GreetingGeneratorApp.swift
+++ b/Examples/Sources/GreetingGeneratorExample/GreetingGeneratorApp.swift
@@ -29,7 +29,8 @@ struct GreetingGeneratorApp: App {
 
                     Toggle("Selectable Greeting", active: $isGreetingSelectable)
                     if let latest = greetings.last {
-                        Text(latest)
+                        LatestGreetingDisplay()
+                            .environment(\.latestGreeting, latest)
                             .padding(.top, 5)
                             .textSelectionEnabled(isGreetingSelectable)
 
@@ -49,5 +50,26 @@ struct GreetingGeneratorApp: App {
                 .padding(10)
             }
         }
+    }
+}
+
+/// This intermediate view exists to show the usage of custom environment keys. In reality it is not necessary.
+struct LatestGreetingDisplay: View {
+    @Environment(\.latestGreeting) var value: String?
+
+    var body: some View {
+        Text(value ?? "nil")
+    }
+}
+
+struct LatestGreetingKey: EnvironmentKey {
+    typealias Value = String?
+    static let defaultValue: Value = nil
+}
+
+extension EnvironmentValues {
+    var latestGreeting: String? {
+        get { self[LatestGreetingKey.self] }
+        set { self[LatestGreetingKey.self] = newValue }
     }
 }
--- a/Package.swift
+++ b/Package.swift
@@ -130,6 +130,10 @@ let package = Package(
             url: "https://github.com/stackotter/swift-benchmark",
             .upToNextMinor(from: "0.2.0")
         ),
+        .package(
+            url: "https://github.com/apple/swift-log.git",
+            exact: "1.6.4"
+        ),
         // .package(
         //     url: "https://github.com/stackotter/TermKit",
         //     revision: "163afa64f1257a0c026cc83ed8bc47a5f8fc9704"
@@ -149,7 +153,8 @@ let package = Package(
             dependencies: [
                 "HotReloadingMacrosPlugin",
                 .product(name: "ImageFormats", package: "swift-image-formats"),
-                .product(name: "OrderedCollections", package: "swift-collections")
+                .product(name: "Logging", package: "swift-log"),
+                .product(name: "OrderedCollections", package: "swift-collections"),
             ],
             exclude: [
                 "Builders/ViewBuilder.swift.gyb",
--- a/Examples/Package.resolved
+++ b/Examples/Package.resolved
@@ -220,8 +220,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/apple/swift-log",
       "state" : {
-        "revision" : "9cb486020ebf03bfa5b5df985387a14a98744537",
-        "version" : "1.6.1"
+        "revision" : "ce592ae52f982c847a4efc0dd881cc9eb32d29f2",
+        "version" : "1.6.4"
       }
     },
     {
--- a/Examples/Sources/CounterExample/CounterApp.swift
+++ b/Examples/Sources/CounterExample/CounterApp.swift
@@ -18,6 +18,11 @@ struct CounterApp: App {
                         count -= 1
                     }
                     Text("Count: \(count)")
+                        .overlay {
+                            GeometryReader { proxy in
+                                Color.blue.opacity(0.5)
+                            }
+                        }
                     Button("+") {
                         count += 1
                     }
--- a/Package.resolved
+++ b/Package.resolved
@@ -91,6 +91,15 @@
       }
     },
     {
+      "identity" : "swift-log",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/apple/swift-log.git",
+      "state" : {
+        "revision" : "ce592ae52f982c847a4efc0dd881cc9eb32d29f2",
+        "version" : "1.6.4"
+      }
+    },
+    {
       "identity" : "swift-macro-toolkit",
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/stackotter/swift-macro-toolkit",
--- a/Sources/AppKitBackend/AppKitBackend.swift
+++ b/Sources/AppKitBackend/AppKitBackend.swift
@@ -1,12 +1,30 @@
 import AppKit
+import Logging
 import SwiftCrossUI
 import WebKit
+
+/// The storage behind `logger`.
+///
+/// `nil` if the logger hasn't been set yet.
+///
+/// > Safety: This is only set once, before it is ever read.
+nonisolated(unsafe) private var _logger: Logger?
+
+/// The global logger for this backend.
+var logger: Logger {
+    guard let _logger else { fatalError("logger not yet initialized") }
+    return _logger
+}
 
 extension App {
     public typealias Backend = AppKitBackend
 
     public var backend: AppKitBackend {
-        AppKitBackend()
+        _logger = Logger(
+            label: "AppKitBackend",
+            factory: Self.logHandler(label:metadataProvider:)
+        )
+        return AppKitBackend()
     }
 }
 
@@ -419,8 +437,7 @@ public final class AppKitBackend: AppBackend {
     public func setPosition(ofChildAt index: Int, in container: Widget, to position: SIMD2<Int>) {
         let container = container as! NSContainerView
         guard container.children.indices.contains(index) else {
-            // TODO: Create proper logging system.
-            print("warning: Attempted to set position of non-existent container child")
+            logger.warning("attempted to set position of non-existent container child")
             return
         }
 
@@ -1259,7 +1276,7 @@ public final class AppKitBackend: AppBackend {
             }
 
             guard response != .abort, response != .cancel else {
-                print("warning: Got abort or cancel modal response, unexpected and unhandled")
+                logger.warning("got abort or cancel modal response, unexpected and unhandled")
                 return
             }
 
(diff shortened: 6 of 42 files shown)


MiaKoring:
Edit: Fixed

oh, apparently I mixed some of my branches… there are changes that not belong to this pr… I’m going to try and remove them...

Examples/Sources/ForEachExample/ForEachApp.swift
```diff
@@ -0,0 +1,105 @@
...
+@main
+@HotReloadable
+struct ForEachApp: App {
+    @State var items = {
+        var items = [Item]()
+        for i in 0..<20 {
+            items.append(.init("\(i)"))
+        }
+        return items
+    }()
```
stackotter: Replace with
```swift
@State var items = (0..<20).map { Item("\($0)") }
```

Examples/Sources/ForEachExample/ForEachApp.swift
```diff
@@ -0,0 +1,105 @@
...
+                                biggestValue += 1
+                                items.insert(.init("\(biggestValue)"), at: insertionPosition)
+                            }
+
+                            Slider($insertionPosition, minimum: 0, maximum: items.count - 1)
+                                .onChange(of: items.count) {
+                                    guard insertionPosition > items.count - 1 else {
+                                        return
+                                    }
+                                    insertionPosition = max(items.count - 1, 0)
```
stackotter: It feels odd having a `guard` to protect the 'bad' path. I think it should be

```swift
let upperLimit = max(items.count - 1, 0)
insertionPosition = min(insertionPosition, upperLimit)
```

If you want you can add an if statement to only set the insertionPosition if it has changed, but I don't believe that that will be necessary for performance at all once we do dependency analysis stuff.

Examples/Sources/ForEachExample/ForEachApp.swift
```diff
@@ -0,0 +1,105 @@
...
+                                }
+                        #endif
+
+                        ForEach(items) { item in
+                            ItemRow(
+                                item: item, isFirst: Optional(item.id) == items.first?.id,
```
stackotter: Move `isFirst` onto a new line

Examples/Sources/ForEachExample/ForEachApp.swift
```diff
@@ -0,0 +1,105 @@
...
+                                isLast: Optional(item.id) == items.last?.id
+                            ) {
+                                items.removeAll(where: { $0.id == item.id })
+                            } moveUp: {
```
stackotter, line 48: I think it'd be best to have the `ForEach` iterate over `items.enumerated` so that each ItemRow can know its own index and avoid this linear operation in what could be a constant-time `moveUp` implementation (same for `moveDown`). I'm pointing this out cause I can see ForEachExample being used to stress test SwiftCrossUI's performance, in which case these move functions should be implemented with best practice.

Examples/Sources/ForEachExample/ForEachApp.swift
```diff
@@ -0,0 +1,105 @@
...
+                                    insertionPosition = max(items.count - 1, 0)
+                                }
+                        #endif
+
+                        ForEach(items) { item in
+                            ItemRow(
+                                item: item, isFirst: Optional(item.id) == items.first?.id,
+                                isLast: Optional(item.id) == items.last?.id
+                            ) {
+                                items.removeAll(where: { $0.id == item.id })
```
stackotter: The comment I left on `moveUp` applies here too. We can make this a little more efficient by directly removing at an index (although the gain won't be as big as for `moveUp` and `moveDown`.

Examples/Sources/ForEachExample/ForEachApp.swift
```diff
@@ -0,0 +1,105 @@
...
+                    .padding(10)
+                }
+            }
+        }
+        .defaultSize(width: 400, height: 800)
+    }
+}
+
+struct ItemRow: View {
+    @State var item: Item
```
stackotter: Item shouldn't be `State` here cause its true source of truth is from the external array. Can we make the `Item` class into a `struct` instead of an observable object?

Examples/Sources/GreetingGeneratorExample/GreetingGeneratorApp.swift
```diff
@@ -38,7 +38,7 @@ struct GreetingGeneratorApp: App {
                                 .padding(.top, 20)
 
                             ScrollView {
-                                ForEach(greetings.reversed()[1...]) { greeting in
+                                ForEach(items: greetings.reversed()[1...]) { greeting in
```
stackotter: Fix this up to use the new non-deprecated syntax.

Examples/Sources/WebViewExample/WebViewApp.swift
```diff
@@ -29,10 +29,12 @@ struct WebViewApp: App {
                     }
                     .padding()
 
-                    WebView($url)
-                        .onChange(of: url) {
-                            urlInput = url.absoluteString
-                        }
+                    #if !os(tvOS)
+                        WebView($url)
```
stackotter, line 41: Add an `#else` with `Text("WebView isn't supported on tvOS")`.

Package.swift
```diff
@@ -110,6 +110,10 @@ let package = Package(
             url: "https://github.com/stackotter/swift-winui",
             revision: "1695ee3ea2b7a249f6504c7f1759e7ec7a38eb86"
         ),
+        .package(
+            url: "https://github.com/apple/swift-collections.git",
+            exact: "1.2.1"
```
stackotter: Is there a reason that this has to be `exact` and not `from`? We should avoid `exact` dependencies because then it's harder to e.g. use swift-collections and swift-cross-ui in the same project.

Examples/Sources/ForEachExample/ForEachApp.swift
MiaKoring: The new ForEach would look like this:

```swift
ForEach(Array(items.enumerated()), id: \.element.id) { (index, item) in
```

EnumeratedSequence<Bla>.Element is not identifiable, so I have to specify the identifier path

Also EnumeratedSequence is only conform to Collection on macOS 26+, so I had to wrapp it in an array. 

If this is fine with you we can leave it like that. Otherwise additional initializers could do it for you and I could remove the necessity to specify a path on enumerated sequences where the element is identifiable. 

Just let me know ^^

Examples/Sources/ForEachExample/ForEachApp.swift
MiaKoring: Tbh I’m not sure why I made this decision while testing haha
I believe I intended to do something with editing which I didn’t end up doing.

The change to a struct works

Package.swift
MiaKoring: true. maybe we should change to a fork with exact eventually to guarantee stability.
I changed it to upToNextMajor(from: ) for now

Package.swift
```diff
@@ -112,7 +124,11 @@ let package = Package(
         ),
         .package(
             url: "https://github.com/apple/swift-collections.git",
-            exact: "1.2.1"
+            .upToNextMinor(from: "1.2.1")
```
stackotter, line 127: Could this be `from: "1.2.1"` instead? I generally do that for projects that are passed `1.0.0`, because semantic versioning, if done right, should guarantee a certain amount of stability.

Sources/SwiftCrossUI/Views/ForEach.swift
```diff
@@ -214,39 +195,34 @@ extension ForEach: TypeSafeView, View where Child: View {
 
         children.isFirstUpdate = false
 
-        return LayoutSystem.updateStackLayout(
+        var cache = children.stackLayoutCache
```
stackotter: Pass this directly to `computeStackLayout` as `cache: &children.stackLayoutCache`, otherwise the cache changes get discarded every time.

In VStack we do this is a weird way because we have to unwrap an optional, but here you can simply pass the stack layout cache directly.

Examples/notes.json
```diff
@@ -0,0 +1 @@
+[{"id":"0269CAC7-700F-474B-AED4-83B077E9023F","title":"Hello, world!","content":"Welcome SwiftCrossNotes!"},{"id":"6B885050-E1F6-4112-9A60-F5DC4E09B5D5","title":"Shopping list","content":"Carrots, mushrooms, and party pies"},{"id":"D52FADA2-BC47-4B22-B07B-4385358DCE71","title":"","content":""},{"id":"613181B3-8AA8-44C4-9170-670FF40EDBD0","title":"","content":""},{"id":"5FF6AA23-E908-4D62-AD27-E739E3B8366B","title":"","content":""},{"id":"5391E1DF-DD76-43FF-98B8-B609619A7C8F","title":"lol","content":""},{"id":"F715854B-31E7-4D34-B3F7-62A30E1AED1A","title":"","content":""},{"id":"7B7C9A53-F8F0-46BE-B490-87FE600451AE","title":"","content":""}]
```
stackotter: Remove this file (I should probably gitignore it)

stackotter (changes requested):
Just requested a few small changes this time. I'll also just have a quick look over all of the changes again because it's been a while since I last looked at this PR (and for this review I just looked at the 'changes since last viewed')

Sources/SwiftCrossUI/Views/ForEach.swift
```diff
@@ -93,6 +63,170 @@ extension ForEach: TypeSafeView, View where Child: View {
...
+        var requiresOngoingReinsertion = false
+
+        // Forces node recreation when enabled (expensive on large Collections).
+        // Use only when idKeyPath yields non-unique values. Prefer Identifiable
+        // or guaranteed unique, constant identifiers for optimal performance.
+        // Node caching and diffing require unique, stable IDs.
+        var ongoingNodeReusingDisabled = false
+
+        // Avoid reallocation
+        var inserted = false
```
stackotter: Move this into the for loop given that it's scoped to each iteration. Having it out here makes it feel like its value may be persisted across iterations or after the for loop has ended.

Package.swift
MiaKoring: It cannot, 1.3 drops support for Swift 5.10

I suppose the way to approach avoiding conflicts would be to replace the ref to apples repo with a fork owned by moreSwift

MiaKoring:
> But again, I'm happy for that to be left as a future direction in order to get your existing ForEach improvements merged.

Good point, I was not aware of this being that bad on Gtk. I’m going to create an issue for the required update and fix it in a future PR.

stackotter:
Actually, I just realised that we should have the original init with some fallback implementation and a deprecation warning. I'll do that now.

PR #283, opened 2025-12-28
Update WinUIBackend dependency wording for clarity

stackotter:
A user of SwiftCrossUI found the wording of the docs to be a bit ambiguous about which dependencies are required as a developer. I agree that it'd be reasonable to assume that you could get away with only installing the SDK, so I've added an extra sentence to make things a bit more clear.


susandahal:
Looks good. Would same few people(includes me ofc) few headaches

PR #296, opened 2026-01-02
UIKitBackend: Fix text rendering (set NSTextContainer.lineFragmentPadding to 0)

stackotter:
#278 broke UIKitBackend's Text implementation when switching from UILabel to a custom text rendering UIView implementation. It didn't override its NSTextContainer's lineFragmentPadding, so the text content got rendered with a horizontal offset and ended up clipping even when there was available space.


yqhbzvyzbw-cloud (commented):
Count

yqhbzvyzbw-cloud (commented):
[Frame]()

Issue #259, opened 2025-11-28
GtkCodeGen: Add gir files to repo (if allowed), and regenerate bindings in CI to ensure properly generated


stackotter:
Currently, when a contributor adds additional bindings to GtkCodeGen, there's a high likelihood that they'll end up with a bunch of unrelated changes due to the generator using the contributor's own gir files (from their Gtk installation). This isn't ideal, because it can lead to backwards compatibility bugs being introduced when people have newer gir files that I haven't built the generator to handle properly (gtk is hard).

I should add my gir files to the repo (if Gtk's licensing allows it), and add a CI job to regenerate the bindings from source and make sure that they match the contents of the PR. The regeneration serves both to avoid a malicious contributor from hiding stuff in big generated blobs the reviewers may assume have been correctly generated, and to ensure that contributors don't forget to regenerate and format the bindings when they update GtkCodeGen.

MiaKoring:
Mentioned, but never closed. Afaict this has been resolved

PR #315, opened 2026-01-07
Clean up scripts, generate bindings from same Gir files every time

stackotter:
This PR cleans up all of SwiftCrossUI's existing bash scripts. It;

- moves them to the new `Scripts` directory,
- pulls them apart into smaller self contained scripts (e.g. `generate_gtk3.sh`),
- adds extra code that ensures the scripts can be called from anywhere,
- optimises Gtk binding generation by running the format script directly on the generated code (avoiding formatting the whole repo),
- and generates the Gtk bindings from a consistent set of Gir files (hosted in https://github.com/moreSwift/gir-files)

You can now regenerate all generated code by running `./Scripts/generate.sh`; no need to run `./Scripts/format_and_lint.sh` after the generation script anymore. You can also regenerate only the code you changed by using `./Scripts/generate_gyb.sh`, `./Scripts/generate_gtk.sh`, `./Scripts/generate_gtk4.sh`, and `./Scripts/generate_gtk3.sh`.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,3 @@
-[submodule "Sources/GtkCodeGen/gir-files"]
-	path = Sources/GtkCodeGen/GirFiles
-	url = https://github.com/moreSwift/gir-files
 [submodule "Sources/GtkCodeGen/GirFiles"]
 	path = Sources/GtkCodeGen/GirFiles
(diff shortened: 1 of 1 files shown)


stackotter:
Addresses #259

PR #326, opened 2026-01-09
Fix menu button sizing (broken by #278)

stackotter:
We weren't updating the button before measuring its naturalSize, and we were using the button's width instead of its height when positioning its popover.

Tested on macOS and iOS.


stackotter:
Tested on all backends before merging.

PR #329, opened 2026-01-09
ForEach: Remove redundant Range-based inits

stackotter:
@bbrk24 faced an ambiguous init that I think this may fix.


bbrk24 (approved):
Fixes the errors I was seeing previously

PR #340, opened 2026-01-11
Fixes layout on iOS and tvOS when windowResizability is fixed

MiaKoring:
Fixes remaining problems of https://github.com/stackotter/swift-cross-ui/issues/333

The layout issues occured when views resized inside a WindowGroup on a backend/device where the window size cannot be set. The layout system just positions the Content at x = 0, y = 0 under the assumption that the window has got the dictated size.

This PR excludes iOS and tvOS from modifying the window resizability through an early exit at runtime. 

iPadOS is not excluded, as it theoretically supports resizing windows, even though its currently being ignored as far as I know.

Its a targeted workaround at the moment as its entirely transparent to the layout engine. I would propose a way for backends to let the layout engine know if the current device/window supports resizing at the moment, as it can also change on iPadOS and Android during runtime. This proposal would be a better solution and should replace this fix in the future.

### DIFF
--- a/Sources/AppKitBackend/AppKitBackend.swift
+++ b/Sources/AppKitBackend/AppKitBackend.swift
@@ -424,7 +424,7 @@ public final class AppKitBackend: AppBackend {
     public func swap(childAt firstIndex: Int, withChildAt secondIndex: Int, in container: NSView) {
         assert(
             container.subviews.indices.contains(firstIndex)
-            && container.subviews.indices.contains(secondIndex),
+                && container.subviews.indices.contains(secondIndex),
             """
             attempted to swap container child out of bounds; container count \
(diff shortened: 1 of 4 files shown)


MiaKoring:
Closed for being delayed and likely implemented as a proper, general solution then

mba6comma1:
> I would propose a way for backends to let the layout engine know if the current device/window supports resizing at the moment, as it can also change on iPadOS and Android during runtime.

Correct me if I'm wrong, but wouldn't the existing `isWindowProgrammaticallyResizable(_:)` work for this?

MiaKoring:
> Correct me if I'm wrong, but wouldn't the existing `isWindowProgrammaticallyResizable(_:)` work for this?

It's set to false on UIKitBackend but currently doesn't get respected and I didn't have the time yet to learn the layout system.

PR #348, opened 2026-01-15
Fix Xcode-based CI jobs

stackotter:
GitHub seems to have changed the macOS 15 CI runner image, and it no longer has the iOS, tvOS, or visionOS SDKs associated with Xcode 16.3. This PR aims to fix that.

### DIFF
--- a/.github/workflows/build-test-and-docs.yml
+++ b/.github/workflows/build-test-and-docs.yml
@@ -73,10 +73,10 @@ jobs:
           - TV
           - Vision
     steps:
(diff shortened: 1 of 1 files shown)


stackotter:
This appears to be the cause: https://github.com/actions/runner-images/issues/12758#issuecomment-3206748945

Due to the size of the visionOS platform component in newer Xcode versions, they can no longer fit all Xcode versions and their platform components in the macOS-15 runner image. They decided to keep all Xcode versions supported by the OS but only include the platform components for the latest 3 Xcode versions.

PR #352, opened 2026-01-16
Android backend

stackotter:
This is still WIP. I'll update the description as it becomes more complete.


Sources/AndroidBackend/Kotlin/dev/swiftcrossui/androidbackend/AndroidBackendHelpers.kt
```diff
@@ -0,0 +1,21 @@
+package dev.swiftcrossui.androidbackend
+
+import android.view.WindowInsets
+import android.app.Activity
+
+/// Static helper functions used by AndroidBackend
+class AndroidBackendHelpers {
```
bbrk24: I believe you can see these as Java static functions if you mark them as `@JvmStatic` and either put them in a `companion object`, or change the `class` here to an `object`.

Sources/AndroidBackend/Kotlin/dev/swiftcrossui/androidbackend/CustomEditText.kt
```diff
@@ -0,0 +1,38 @@
...
+
+class CustomEditText(
+    val activity: Activity,
+    var onChange: SwiftAction? = null,
+    var onSubmit: SwiftAction? = null
+): EditText(activity) {
+    init {
+        setOnEditorActionListener { v, actionId, event ->
+            Log.i("CustomEditText", "Editor action!")
+            if (actionId == EditorInfo.IME_ACTION_SEND) {
```
bbrk24: Do we want to restrict this to "send"? There's also "go" and "search".

Sources/AndroidBackend/Kotlin/dev/swiftcrossui/androidbackend/SwiftObject.kt
```diff
@@ -0,0 +1,17 @@
...
+/// Swift object retained by JVM. We have to support both 32-bit and 64-bit
+/// systems, so we store the pointers as longs.
+class SwiftObject(val pointerValue: Long) {
+    override fun toString(): String {
+        return toStringSwift()
+    }
+
+    external fun toStringSwift(): String
+
+    fun finalize() {
```
bbrk24: nitpick: `finalize` is usually `protected`.

Sources/AndroidBackend/AndroidBackend.swift
```diff
@@ -0,0 +1,369 @@
...
+    android_log(Int32(ANDROID_LOG_DEBUG.rawValue), "swift", message)
+}
+
+/// A valid AndroidBackend shim must call this to begin execution of the app.
+/// Once initial setup and rendering is done, this function returns control
+/// back to the JVM (by returning).
+@MainActor
+@_cdecl("AndroidBackend_entrypoint")
+public func entrypoint(_ env: UnsafeMutablePointer<JNIEnv?>, _ object: jobject) {
+    let env = JNIEnvWrapper(env: env)
```
bbrk24: The fact that these have the same name is a little confusing further down (e.g. `env.env` threw me off)

Sources/AndroidBackend/AndroidBackend.swift
```diff
@@ -0,0 +1,369 @@
...
+extension App {
+    public typealias Backend = AndroidBackend
+
+    public var backend: AndroidBackend {
+        AndroidBackend()
+    }
+}
+
+public final class AndroidBackend: AppBackend {
+    public typealias Window = Void
```
bbrk24: Huh, not [`android.view.Window`](https://developer.android.com/reference/android/view/Window)?

Sources/AndroidBackend/AndroidBackend.swift
```diff
@@ -0,0 +1,369 @@
...
+    public typealias Widget = AndroidKit.View
+    public typealias Menu = Never
+    public typealias Alert = Never
+    public typealias Path = Never
+    public typealias Sheet = Never
+
+    static let stdoutPipe = Pipe()
+    static let stderrPipe = Pipe()
+
+    public let deviceClass = DeviceClass.phone
```
bbrk24, line 87: Android also supports tablets and TVs, so in the fullness of time this should reflect that. I've never made an Android TV app, but typically a device is considered a tablet if the smaller of screen width and screen height is at least 600dp. This does mean some devices are considered phones that you wouldn't expect, like the Lenovo Tab M8 which has an 8"/20cm screen diagonal (about the same as an iPad mini), but that does have a SIM card slot.

Sources/AndroidBackend/AndroidBackend.swift
```diff
@@ -0,0 +1,369 @@
...
+    public let defaultTableRowContentHeight = 0
+    public let defaultTableCellVerticalPadding = 0
+    public let defaultPaddingAmount = 10
+    public let scrollBarWidth = 0
+    public let requiresToggleSwitchSpacer = false
+    public let defaultToggleStyle = ToggleStyle.switch
```
bbrk24: Nitpick: This is completely arbitrary, but based on my experience with Material, I would have put "checkbox" here. (Material does have switches but they're more finnicky to style than I like, and not as common as checkboxes.)

Sources/AndroidBackend/AndroidBackend.swift
```diff
@@ -0,0 +1,369 @@
...
+        tickler.start()
+        self.tickler = tickler
+
+        // We just fall through to return control to Java when we're done
+        // setting up the initial view hierarchy.
+        callback()
+    }
+
+    public func createWindow(withDefaultSize defaultSize: SIMD2<Int>?) -> Window {
+        // TODO: Find out whether Android has a window abstraction like UIKit does.
```
bbrk24: It does: [`android.view.Window`](https://developer.android.com/reference/android/view/Window). One thing to note is that alerts (typically [`android.app.DialogFragment`](https://developer.android.com/reference/android/app/DialogFragment) or subclasses thereof) are considered to be in their own window.

Sources/AndroidBackend/AndroidBackend.swift
```diff
@@ -0,0 +1,369 @@
...
+    public func setResizability(ofWindow window: Window, to resizable: Bool) {}
+
+    public func setChild(ofWindow window: Window, to child: Widget) {
+        Self.activity.setContentView(child)
+    }
+
+    public func size(ofWindow window: Window) -> SIMD2<Int> {
+        let width = Int(helpers.getWindowWidth(Self.activity))
+        let height = Int(helpers.getWindowHeight(Self.activity))
+        return SIMD2(Int(width), Int(height))
```
bbrk24, line 154: Those methods return numbers in `px`, which is screen pixels. This is probably not what we want for high-density screens (Android's equivalent of Retina displays) -- we probably want `dp`/`dip`, which are equivalent to iOS's `pt`. One very important difference is that Android APIs work with `px` unless specified otherwise, while UIKit APIs work with `pt` by default.

(`dp` and `dip` are two different spellings of the same thing.)

Sources/AndroidBackend/AndroidBackend.swift
```diff
@@ -0,0 +1,369 @@
...
+    }
+
+    public func setSizeLimits(ofWindow window: Void, minimum minimumSize: SIMD2<Int>, maximum maximumSize: SIMD2<Int>?) {}
+
+    public func setBehaviors(ofWindow window: Void, closable: Bool, minimizable: Bool, resizable: Bool) {}
+
+    public func setResizeHandler(
+        ofWindow window: Window,
+        to action: @escaping (_ newSize: SIMD2<Int>) -> Void
+    ) {}
```
bbrk24: Since I don't see a TODO comment here, this needs to be wired up to the activity's `onConfigurationChanged`.

Sources/AndroidBackend/AndroidBackend.swift
```diff
@@ -0,0 +1,369 @@
...
+        environment: EnvironmentValues
+    ) -> SIMD2<Int> {
+        let widget = createTextView()
+        updateTextView(widget, content: text, environment: environment)
+        widget.measure(
+            proposedWidth.map(Int32.init) ?? Int32.max,
+            proposedHeight.map(Int32.init) ?? Int32.max
+        )
+        let width = widget.getMeasuredWidth()
+        let height = widget.getMeasuredHeight()
```
bbrk24, line 393: That's one way to do that. (I don't have any other suggestions besides maybe reusing the same widget.)

Sources/AndroidBackend/JNI.swift
```diff
@@ -0,0 +1,89 @@
+import Android
+
+struct JNIError: Error {
+    var message: String
+}
+
+struct JNIEnvWrapper {
+    var env: UnsafeMutablePointer<JNIEnv?>
```
bbrk24: Huh, is the env ptr guaranteed to be stable past when the native method returns? TIL.

Sources/AndroidBackend/MainRunLoopTickler.swift
```diff
@@ -0,0 +1,36 @@
...
+class MainRunLoopTickler: JavaObject {
+    @JavaMethod
+    @_nonoverride convenience init(environment: JNIEnvironment? = nil)
+
+    @JavaMethod
+    func start()
+}
+
+@JavaImplementation("dev.swiftcrossui.androidbackend.MainRunLoopTickler")
+extension ViewOnClickListener {
```
bbrk24: huh?

Sources/AndroidBackendShim/include/header.h
```diff
@@ -0,0 +1,7 @@
+#include <android/log.h>
+
+// We use this forward declaration to call the app's main function without ever
+// being given a handle to it.
+int main(void);
```
bbrk24: `(void)`, really? I suppose there's no good way to get an argv on Android but you could technically `main(0, nil)` just to be safe. But if this works I'm not opposed to leaving it as `(void)`.

Sources/Gtk3Backend/Gtk3WidgetRepresentable.swift
```diff
@@ -219,7 +220,9 @@ final class RepresentingWidget<Representable: Gtk3WidgetRepresentable>: Gtk3.Fix
 
     deinit {
         if let context, let child {
-            Representable.dismantleGtk3Widget(child, coordinator: context.coordinator)
+            Task { @MainActor in
```
bbrk24, line 233: Nitpick: Since GTK 3 isn't supported on any Apple platforms, couldn't this use `Task.immediate`?

Sources/SwiftCrossUI/Views/Shapes/Shape.swift
```diff
@@ -2,7 +2,7 @@
 ///
 /// If no stroke color or fill color is specified, the default is no stroke and a fill of the
 /// current foreground color.
-public protocol Shape: View, Sendable where Content == EmptyView {
+public protocol Shape: View, Sendable, _RemoveGlobalActorIsolation where Content == EmptyView {
```
bbrk24, line 5: Why is `_RemoveGlobalActorIsolation` needed here? Is `nonisolated` not sufficient?

Package.swift
```diff
@@ -170,6 +179,9 @@ let package = Package(
                 "SwiftCrossUI",
                 "DummyBackend",
                 .target(name: "AppKitBackend", condition: .when(platforms: [.macOS])),
+                .product(
+                    name: "ImageFormats", package: "swift-image-formats",
+                    moduleAliases: ["ImageFormats": "Asdf"], condition: .when(platforms: [.macOS])),
```
bbrk24: what

Sources/AndroidBackend/Kotlin/dev/swiftcrossui/androidbackend/AndroidBackendHelpers.kt
stackotter: Ah that `@JvmStatic` annotation would be what I was missing. I originally had them as methods on a companion object but was getting runtime issues due to the Swift code trying to access a non-existent method. Makes sense, thanks!

Sources/AndroidBackend/AndroidBackend.swift
stackotter: Yeah, that env wrapper type is from when I was rolling my own JNI code. I've been meaning to rip that out and simplify things a bit (now that we're fully using AndroidKit).

Sources/SwiftCrossUI/Views/Shapes/Shape.swift
stackotter: It's a trick that SwiftUI uses to remove the default actor isolation of members conforming to a protocol that inherits from a protocol with a default actor isolation. This just makes it a bit more ergonomic to write shape implementations and matches SwiftUI (otherwise every shape property ends up getting manually marked nonisolated I think? can't quite remember the specifics)

Package.swift
stackotter: Lol, must've been from some earlier testing, no clue what for though, good spot

Sources/AndroidBackend/AndroidBackend.swift
stackotter: Yeah I agree, we should be using density adaptive measurements

Sources/AndroidBackend/JNI.swift
stackotter: I guess so 😅

Sources/AndroidBackend/AndroidBackend.swift
stackotter: Yeah... would prefer a nicer method. We can at least reuse the widget like we do on Windows. I think I didn't touch this implementation again after it worked

Sources/SwiftCrossUI/Views/Shapes/Shape.swift
bbrk24: Right, but SwiftUI did that trick before `nonisolated protocol` was valid syntax. I'm wondering if, now that that's valid, that would be a cleaner solution.

stackotter: Ah right yeah, definitely once we bump to Swift 6.1 properly. I think that Swift 6.1 has enough compelling features that it's probably worth us bumping to it soon, but will have to think through the consequences first (I think it would technically be possible to drop the requirement again before merging this PR).

Proposal is [SE-0449](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0449-nonisolated-for-global-actor-cutoff.md) (for anyone else coming across this).

Sources/Gtk3Backend/Gtk3WidgetRepresentable.swift
stackotter: Does `Task.immediate` make a difference when given a `@MainActor` closure? I guess it could when fired off the `@MainActor`, if Swift does the runtime checks required to identify that the first part of the `@MainActor` closure can run synchronously up to the first suspension point. Not sure whether that's how it would actually compile though

Sources/SwiftCrossUI/Views/Shapes/Shape.swift
stackotter: For now I've dropped the Swift requirement back to 5.9, but we should keep this in mind when we want to update to >=6.1

Sources/Gtk3Backend/Gtk3WidgetRepresentable.swift
stackotter: Anyway, I do occasionally compile Gtk3Backend on macOS to make sure that it doesn't have build errors, so it'd be nice to not rule that out.

Sources/AndroidBackend/MainRunLoopTickler.swift
stackotter: I have no clue how that hasn't caused any problems...

Sources/AndroidBackend/AndroidBackend.swift
stackotter: Yeah, once we start testing with tablets we should update that property to dynamically determine the device class. I'll leave a TODO for now

Sources/AndroidBackend/Kotlin/dev/swiftcrossui/androidbackend/AndroidBackendHelpers.kt
stackotter: If I make them static the experience of using the methods from Swift turns out to be almost identical so I'll leave them as is for now; (you can't use them as static members directly in Swift, you first have to create a `JavaClass<AndroidBackendHelpers>` instance and then access the static members as instance members)

Sources/AndroidBackend/AndroidBackend.swift
stackotter: Going to leave that as a TODO for now, just trying to get a solid base for people to work off while I work on some other sbun and scui things for a bit

Sources/AndroidBackend/JNI.swift
stackotter: Seems like we're allowed to store it however long we need it, but we can't share it between threads (which makes sense), so we'll have to be careful of that.

stackotter (commented):
Skimmed through my changes and nothing obvious stood out. I think once the CI passes this should be ready to merge. Not in a stable capacity, but given the rudimentary nature of the current AndroidBackend implementation I think it's pretty clear that it's experimental.

The reason that I'm merging this in while it's still so basic is to avoid having to rebase it any more, and to make it easier for other contributors to collaborate on this feature now that all of the ground work is there.

I won't update the readme or docs to mention AndroidBackend until it supports a majority of the example apps.

PR #393, opened 2026-01-26
Implement windowScaleFactor for AppKitBackend

ulrikdamm:
Just a very small change, implementing windowScaleFactor in the AppKit backend. Not sure if there was any reason it was left as a TODO, since it was very quick to do, please let me know if I've missed something.

Has been tested to correctly update the value when moving between a retina and non-retina display. Didn't see any change in UI rendering as a result of the implementation (doesn't seem like the scale factor is in real use anywhere yet).

I couldn't run the tests though, the test.sh mentioned in the contribution guidelines doesn't exist and `swift test` couldn't compile the WinUI backend because of missing headers.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,7 +8,7 @@
 
 ## Running tests
 
-Due to limitations of Swift Package Manager and the way this project is structured, running tests is a little more annoying than just running `swift test`. Luckily, there's a handy [test.sh](test.sh) script which performs the required workaround. To run the tests, just run `./test.sh` in the root of the repository.
+Due to limitations of Swift Package Manager and the way this project is structured, running tests is a little more annoying than just running `swift test`. Luckily, there's a handy [test.sh](Scripts/test.sh) script which performs the required workaround. To run the tests, just run `./Scripts/test.sh` in the root of the repository.
 
 ## How to do something useful
 


Sources/AppKitBackend/AppKitBackend.swift
```diff
@@ -411,15 +411,21 @@ public final class AppKitBackend: AppBackend {
...
     ) {
-        // TODO: Notify when window scale factor changes
+        // For updating window scale factor
+        NotificationCenter.default.addObserver(
+            forName: NSWindow.didChangeBackingPropertiesNotification,
+            object: window,
+            queue: .main
+        ) { notification in
+            action()
+        }
```
stackotter, line 434: I think that we should make this a bit more granular so that it only triggers when the scale factor changes, as it seems like this might get called a bit more often than that. We use a custom window type, so you could update that to store 'last known scale factor'.

ulrikdamm:
Alright, environment now only updates when backing scale factor actually changes, and I've run the tests and updated the contribution guide.

It seems like right now only the GTK+4 backend supports scale factor. The iOS backend should be pretty easy to implement based on the AppKit implementation (not sure how to test scale factor changing on iOS though, on iPadOS moving a window to an external non-retina display maybe? 🤔).

stackotter:
Works locally 👍

Issue #279, opened 2025-12-13
Support environment insertion of arbitrary ObservableObjects (or other observable types)


MiaKoring:
SwiftUI allows inserting Observable (@Observable) directly into the environment without EnvironmentKeys (ObservableObjects through .environmentObject).
You would access them through @Environment(XYZ.self) and @EnvironmentObject(ZYX.self) respectively.

I believe we should add this ourselves, because its very useful for app state for example.

Once the pr allowing environment insertions through EnvironmentKey is merged, I’m going to work on the implementation. I’m not entirely sure why SwiftUI separates @Observable and ObservableObject in this case, both should just be references internally. It might be due to the way they propagate view updates. With SCUIs ObservableObject being compatible with @State I believe we don’t need this separation.

svanimpe:
Are you referring to `@Observable`, the new macro? If so, that replaces the need for `ObservableObject`, `@Published`, `@StateObject`, and `@EnvironmentObject`.

That would be really great to have. SwiftUI's state wrappers/macros can be quite confusing, and even more so because SwiftCrossUI doesn't match them exactly 😅

svanimpe:
As per Apple's recommendations, most SwiftUI apps rely heavily on observable environment objects to pass around view models / stores:

```swift
view
    .environment(comicStore)

@Environment(ComicStore.self) private var comicStore
```

Without these, we have to resort back to manual dependency injection through initializers, which gets tedious quite fast.

https://developer.apple.com/documentation/swiftui/scene/environment(_:)

stackotter:
This was resolved by #395

PR #395, opened 2026-01-29
Makes usage of ObservableObject in environment possible and adds @Observable and @ObservationIgnored macros

MiaKoring:
Implements suggestion from https://github.com/moreSwift/swift-cross-ui/issues/279

## Features added:
- `@Environment` now supports `ObservableObject`s, which can be inserted into the environment via `.environment([instance of observable object])`
- `@Observable` macro generating `@Published` on supported class members
- `@ObservationIgnored` macro, used as a flag to opt out of observation for certain properties

## Macro Logic
The `@Observable` macro
- generates conformance to `ObservableObject` via an extension macros
- applies `@Published` to stored, non-static single variable declarions with publicly accessible getters
- excludes members with `@ObservationIgnored` applied
- throws an error if the class contains declarations like `var a, b: String` without `@ObservationIgnored`

## QA
- Added unit test validating the macro code generation logic

## Other
- added `ObservableEnvironmentExample` showcasing the changes (can be removed at any time, I just didn’t come up with an idea for a logical use in one of the existing Examples)

## Note
Due to `@Observable` just generating the otherwise manually added boilerplate the changes are fully backwards compatible to existing ObservableObjects

### DIFF
--- a/Sources/SwiftCrossUI/Environment/Environment.swift
+++ b/Sources/SwiftCrossUI/Environment/Environment.swift
@@ -55,18 +55,9 @@ public struct Environment<Value>: DynamicProperty {
 
     public var wrappedValue: Value {
         guard let value = value.value else {
-            let pathDescription = {
-                return switch mode {
-                    case .keyPath(let keyPath):
-                        "\(keyPath)"
-                    case .observableObject:
-                        "\(Value.self).self"
-                }
-            }()
-
             fatalError(
                 """
-                Environment value at \(pathDescription) used before initialization. Don't \
+                Environment value at \(mode.pathDescription) used before initialization. Don't \
                 use @Environment properties before SwiftCrossUI requests the \
                 view's body.
                 """
@@ -88,5 +79,14 @@ public struct Environment<Value>: DynamicProperty {
     private enum Mode {
         case keyPath(KeyPath<EnvironmentValues, Value>)
         case observableObject
+
+        var pathDescription: String {
+            switch self {
+                case .keyPath(let keyPath):
+                    "\(keyPath)"
+                case .observableObject:
+                    "\(Value.self).self"
+            }
+        }
     }
 }


MiaKoring:
(sorry, format_and_lint.sh generated more changes than originally made by me again. I hope its not too distracting)

Sources/SwiftCrossUI/Environment/EnvironmentValues.swift
```diff
@@ -125,6 +125,15 @@ public struct EnvironmentValues {
         }
     }
 
+    public subscript<T: ObservableObject>(observable key: T.Type) -> T? {
+        get {
+            extraValues[ObjectIdentifier(T.self)] as! T?
```
stackotter: We should `log.critical` a proper error message when the types don't match (and then fatally error ourselves).

Sources/SwiftCrossUI/Environment/EnvironmentValues.swift
```diff
@@ -125,6 +125,15 @@ public struct EnvironmentValues {
         }
     }
 
+    public subscript<T: ObservableObject>(observable key: T.Type) -> T? {
```
stackotter, line 149: I'm leaning towards saying that we should store the observable objects in a separate dictionary, because someone could technically make something an observable object and an environment key. I'm not sure whether it matters, but environment value keys and environment object keys do feel conceptually distinct.

Sources/SCUIMacrosPlugin/CompilerPlugin.swift
```diff
@@ -0,0 +1,9 @@
+import SwiftCompilerPlugin
+import SwiftSyntaxMacros
+
+@main
+struct SCUIMacrosPlugin: CompilerPlugin {
```
stackotter: We should move our hot reloading macros into this target as well, because I don't really see a reason to have them live in a separate macro target.

stackotter: We should call the containing target `SwiftCrossUIMacrosPlugin` for consistency with our naming of other targets (and rename this type accordingly).

Sources/SCUIMacrosPlugin/ObservableMacro.swift
```diff
@@ -0,0 +1,101 @@
...
+                let trimmedDescription = attr.as(AttributeSyntax.self)?
+                    .attributeName
+                    .trimmedDescription
+                return
+                    trimmedDescription == "Published"
+                    || trimmedDescription == "SwiftCrossUI.Published"
+                    || trimmedDescription == "ObservationIgnored"
+                    || trimmedDescription == "SwiftCrossUI.ObservationIgnored"
+            }),
+            // only non computed properties
```
stackotter: This check also excludes stored properties with `didSet`. I think that's reasonable, but the comment should be updated to reflect that;

`// Only include properties without accessors`

Sources/SwiftCrossUI/State/ObservableObject.swift
```diff
@@ -104,3 +104,21 @@ extension Optional: OptionalObservableObject where Wrapped: ObservableObject {
...
 public typealias Observable = ObservableObject
+
+/// Automatically observes all public noncomputed variables with public getter and setter
+@attached(memberAttribute)
+@attached(extension, conformances: ObservableObject)
+public macro Observable() =
+    #externalMacro(
+        module: "SCUIMacrosPlugin",
+        type: "ObservableMacro"
+    )
```
stackotter, line 115: We should rename this macro in order to avoid breaking changes when we support the `@Observable` macro's proper behaviour. SwiftUI's `Observable` macro provides a conformance to the `Observable` protocol rather than the `ObservableObject` protocol, so if we keep the macro naming now and make the change in future, then people could be relying on the `didChange` property of their `@Observable` classes, which would disappear because SwiftUI's `@Observable` doesn't use publishers (it uses a much more efficient approach).

`ObservableObject` is probably the best name to use for this macro.

The `ObservationIgnored` macro doesn't have to be renamed because it's just a marker attribute. Additionally, keeping its name consistent with SwiftUI's corresponding macro will make it easier for people to transition once we support `@Observable` properly.

Sources/SwiftCrossUI/Views/Modifiers/EnvironmentModifier.swift
```diff
@@ -61,4 +61,13 @@ extension View {
             environment.with(keyPath, newValue)
         }
     }
+
+    /// Modifies the environment of the View its applied to
```
stackotter:
```swift
/// Adds an observable object to the environment of the enclosed View.
/// You are responsible for ensuring that the object is being observed
/// by a parent view, as this modifier does not perform any observation.
```

Examples/Package.swift
```diff
@@ -94,6 +94,10 @@ let package = Package(
             name: "AdvancedCustomizationExample",
             dependencies: exampleDependencies,
             resources: [.copy("Banner.png")]
+        ),
+        .executableTarget (
+            name: "ObservableEnvironmentExample",
```
stackotter: I think it'd be best to combine this into an existing example, as it's not showing off much functionality. It could be more compelling to use it in RandomNumberGenerator (just not CounterExample because that should be kept as simple as possible).

Sources/SCUIMacrosPlugin/ObservableMacro.swift
```diff
@@ -0,0 +1,101 @@
...
+import SwiftSyntaxMacros
+
+public struct ObservableMacro: MemberAttributeMacro, ExtensionMacro {
+    public static func expansion(
+        of node: AttributeSyntax,
+        attachedTo declaration: some DeclGroupSyntax,
+        providingAttributesFor member: some DeclSyntaxProtocol,
+        in context: some MacroExpansionContext
+    ) throws -> [AttributeSyntax] {
+        guard let classDecl = ClassDeclSyntax(declaration) else {
```
stackotter: `classDecl` is unused according to CI

Sources/SCUIMacrosPlugin/ObservableMacro.swift
```diff
@@ -0,0 +1,101 @@
+import MacroToolkit
+import SwiftSyntax
+import SwiftSyntaxMacroExpansion
+import SwiftSyntaxMacros
+
+public struct ObservableMacro: MemberAttributeMacro, ExtensionMacro {
```
stackotter: If you're comfortable giving it a go, I think it'd be worthwhile updating this implementation to use more of MacroToolkit. This example code should help with this specific use case: https://github.com/stackotter/swift-macro-toolkit/blob/77d4a6d6215a0709d169fa1c2be4df8d8e5f0796/Sources/MacroToolkitExamplePlugin/CustomCodableMacro.swift

I'm not sure how SwiftUI's `@Observable` modifier handles things, but using MacroToolkit would allow the macro to support variable declarations that declare multiple variables at once, and whatever other weird stuff users like doing. And it'll hopefully also make the code shorter/easier to read.

Sources/SwiftCrossUI/Environment/Environment.swift
```diff
@@ -36,14 +36,19 @@
...
 
     public func update(
         with environment: EnvironmentValues,
         previousValue: Self?
     ) {
-        value.value = environment[keyPath: keyPath]
+        if let keyPath {
+            value.value = environment[keyPath: keyPath]
+        } else if let observableType = Value.self as? any ObservableObject.Type {
+            value.value = environment[observable: observableType] as! Value?
```
stackotter: Is the `as! Value?` required here? I don't think it should be.

Sources/SwiftCrossUI/Environment/Environment.swift
```diff
@@ -36,14 +36,19 @@
 /// ```
 @propertyWrapper
 public struct Environment<Value>: DynamicProperty {
-    var keyPath: KeyPath<EnvironmentValues, Value>
+    var keyPath: KeyPath<EnvironmentValues, Value>?
+    var observableType: Value.Type?
```
stackotter: It's unnecessary for us to store `Value.Type` here because we can just use `Value.self` here. I know that it could technically be a subclass of the parent class, but I think we just want it to be `Value.self` anyway.

I think we should replace `keyPath` and `observableType` with a `Mode` enum that has a `keyPath(KeyPath<...>)` case and an `object` case (without an associated value. That we we're also making sure it's never in an unexpected state (e.g. both nil or both present).

Tests/SwiftCrossUITests/ObservableTests.swift
```diff
@@ -0,0 +1,160 @@
...
+            macros: testMacros
+        )
+    }
+    
+    @Test("Private and static  properties are ignored")
+    func testPrivateAndStaticAreIgnored() {
+        assertMacroExpansion(
+            """
+            @Observable
+            class ViewModel {
```
stackotter, line 89: If the model is public, should we ignore internal properties as well as private properties? That would of course be a different unit test, but this one brought that idea to mind.

Sources/SCUIMacrosPlugin/ObservableMacro.swift
MiaKoring: It doesn’t exclude didSet, but didSet isn’t called it looks like. So it would probably be better to exclude it.
Screenshot 2026-02-01 at 12 22 42

Sources/SwiftCrossUI/Environment/Environment.swift
MiaKoring: It is required, because I need to cast it to any ObservableObject.Type, making it look to the compiler as not being Value anymore.
I could remove this by making the observableObject store in EnvironmentValues and the subscript not constrained to any ObservableObject anymore, but in my opinion its better like this, to make sure the observableObject store only can contain ObservableObjects.

The alternative would be @EnvironmentObject as a separate wrapper with a constrained generic

Sources/SCUIMacrosPlugin/ObservableMacro.swift
MiaKoring: I don’t know how you would support multi variable declarations. I don’t even know how it would be done in plain code.
```swift
@Published
var a, b, c, d: Int
```
doesn’t work and neither does
```swift
var @Published a, @Published b, @Published c, @Published d: Int
```

I changed the parts supported by macro toolkit over. Do you have an example for me with the property wrapper on multi binding declarations?

Tests/SwiftCrossUITests/ObservableTests.swift
MiaKoring: I don’t think we can do that, as a package may use internal properties in Views.

Sources/SCUIMacrosPlugin/ObservableMacro.swift
stackotter: Ah yeah, I guess if you can't change the variable declarations themselves then you can't support multi-binding variable declarations (you'd have to split the declarations into separate declarations which may not always be possible). For proper `@Observable` we may be able to get that to work, but not sure.

Sources/SCUIMacrosPlugin/ObservableMacro.swift
stackotter: I guess the didSet then refers to setting the value of `_text: Published<String>` instead? Interesting that it's allowed, but yeah we should probably block it if it doesn't act intuitively

Examples/Sources/RandomNumberGeneratorExample/RandomNumberGeneratorApp.swift
```diff
@@ -25,50 +25,62 @@ enum ColorOption: String, CaseIterable {
...
-                    Slider(
-                        value: $minNum.onChange { newValue in
-                            if newValue > maxNum {
-                                minNum = maxNum
-                            }
-                        },
-                        in: 0...100
-                    )
+struct ContentView: View {
+    @Environment(ViewModel.self) var vm
```
stackotter: Call this `viewModel`, and put a blank line below for consistency

Examples/Sources/RandomNumberGeneratorExample/RandomNumberGeneratorApp.swift
```diff
@@ -25,50 +25,62 @@ enum ColorOption: String, CaseIterable {
...
-                        value: $maxNum.onChange { newValue in
-                            if newValue < minNum {
-                                maxNum = minNum
-                            }
-                        },
-                        in: 0...100
-                    )
+            Text("Minimum: \(vm.minNum)")
+            Slider(
+                value: vm.$minNum.onChange { newValue in
```
stackotter: Does this work with `$vm.minNum`? I think that's the way people generally do it in SwiftUI, and we should probably encourage that too because it makes it easier to transition from an ObservableObject to a struct. It's possible that SwiftUI also doesn't implement a projected value for `Published`, in which case we should probably deprecate that in future

Sources/SwiftCrossUIMacrosPlugin/SwiftCrossUIMacrosPlugin.swift
```diff
@@ -0,0 +1,12 @@
+import SwiftCompilerPlugin
+import SwiftSyntaxMacros
+
+@main
+struct SwiftCrossUIMacrosPlugin: CompilerPlugin {
```
stackotter, line 5: Rename the file to match the struct name

Sources/SwiftCrossUIMacrosPlugin/Observability/ObservableObjectMacro.swift
```diff
@@ -3,45 +3,41 @@ import SwiftSyntax
...
+                        "SwiftCrossUI.ObservationIgnored",
+                    ].contains(attr.attribute?.name.name)
             }),
-            // only non computed properties
-            !variable.bindings.contains(where: {
-                guard let accessor = $0.accessorBlock else { return false }
-                return indicatesComputed(accessor: accessor)
-            })
+            // Only include properties without accessors
+            variable.isStoredProperty
```
stackotter: This allows `didSet` and `willSet`, but in this case we want to disallow all accessors as you discovered when trying `didSet`.

```swift
// Only support variables with a single bindings.
let binding = destructureSingle(variable.bindings),
// Don't allow any accessors, because even when the property is
// stored (i.e. supports `@Published`), the added property wrapper
// changes the meaning of `didSet` and `willSet` accessors.
binding.accessors.isEmpty
```

That check also allows you to remove the binding check below because it ensures that there's only a single binding.

Sources/SwiftCrossUI/Environment/Environment.swift
stackotter: Ah ok, yes you're right, I didn't read that part of the code carefully enough

Sources/SwiftCrossUI/State/ObservableObject.swift
```diff
@@ -108,17 +108,17 @@ public typealias Observable = ObservableObject
...
-        module: "SCUIMacrosPlugin",
-        type: "ObservableMacro"
+        module: "SwiftCrossUIMacrosPlugin",
+        type: "ObservableObjectMacro"
     )
 
-/// Apply to a member inside your `@Observable` class to  opt out of observation
-// This macro is just used as a flage for `@Observable` to ignore a specific property
+/// Apply to a member inside your `@ObservableObject` class to  opt out of observation
+// This macro is just used as a flage for `@ObservableObject` to ignore a specific property
```
stackotter: flag is typoed as flage

Sources/SwiftCrossUI/Environment/Environment.swift
```diff
@@ -36,26 +36,28 @@
...
+                }
         }
     }
 
     public var wrappedValue: Value {
         guard let value = value.value else {
             fatalError(
                 """
-                Environment value at \(keyPath) used before initialization. Don't \
+                Environment value at \(mode) used before initialization. Don't \
```
stackotter: Display the keyPath case as `keyPath.description`, and the observableObject case as `"\(Value.self).self"`

s-k:
@MiaKoring @stackotter Just a question from someone very interested in this project: Why would you reimplement `@Observable` in this project? Doesn't the standard `@Observable` implementation come with Swift? I think, the standard implementation could be used relatively easily by wrapping all calls of `body` properties inside `withObservationTracking()` like this:

```swift
var body: Content! = nil
withObservationTracking {
  body = someView.body
} onChange: {
  // Mark `someView` as 'dirty'. Its `body` property needs to be re-evaluated on the next run loop iteration.
}
// Use `body` for rendering
```

If you are interested, I would be happy to try writing an implementation after familiarizing myself with the project.

MiaKoring:
> Doesn't the standard `@Observable` implementation come with Swift?
Oh, you are right 👀 
I always thought it was apple platform bound. 

That would be cool to have! Let us know if you have any questions.

Are you already on the discord server? Its usually a bit easier to go back and forth there in my opinion.

MiaKoring:
A problem might be the compatibility tho. Currently we support OS versions back to iOS 13,...

stackotter:
Yeah, unfortunately the Observation module hasn’t been back ported to earlier macOS and iOS versions, and likely won’t if the forum threads I’ve read are the be believed. It’s only available from macOS 14 onwards and iOS 17 onwards which is too restrictive.

Luckily for us though, the code for the Observation module is open source, so we should be able to use the module as long as we have a copy of the implementation within SwiftCrossUI itself that we can use on earlier platforms that don’t have stdlib observation support. That would be cool to have if you’d be interested in working on that @s-k !

mba6comma1:
> Yeah, unfortunately the Observation module hasn’t been back ported to earlier macOS and iOS versions, and likely won’t if the forum threads I’ve read are the be believed.

You can always count on pointfree to do some magic: https://github.com/pointfreeco/swift-perception

stackotter:
Sweet! We probably want to copy the built-in APIs when back porting them rather than renaming them, but we might be able to borrow code from there if the stdlib implementation has things we can’t copy

s-k:
I'd be happy to try an implementation. However, as I said, I need to familiarize myself with the project first. Is there any deadline or preferred date when the implementation should be finished?

I would start by making the default Swift implementation of `Observable` work and could then add an `Observable` implementation to the project. Is that OK? And are there any licensing issues when copying code into the project? swift-perception seems to have the same license as this project.

I have just joined the Discord (my name is `_stephen_`). However, I will not be very responsive there because I have disabled all notifications on my phone an don't regularly check Discord. However, you can reach me via github or email.

stackotter:
No rush at all. Your approach sounds sensible. In terms of licensing, I believe we can use any MIT or Apache 2 licensed code, but we should reference anything copied verbatim and should copy original licenses alongside copied files (even if they match ours). For Apache and MIT we may not have to copy the licenses, but I would like to anyway.

s-k:
I have some news, but first of all: Is this a good place to discuss this? Maybe an issue would be a better place? Until a better place is found, here is the news:

I have `Observable` support pretty much running with a minimum target of macOS 14. (`View`s automatically update themselves. `App` object not yet.) I also see a path to make it work on older versions of macOS and other platforms. However, I have found an issue: `body` properties of `View`s are executed many times for a single update. And the problem are not my changes. Take this code for example:

```swift
@main
struct CounterApp: App {
    @State var count = 0
    var body: some Scene {
        WindowGroup {
            HStack {
                Button("-") {
                    count -= 1
                }
                Text("Count: \(count)")
                Button("+") {
                    count += 1
                }
            }
            ContentView()
        }
    }
}

struct ContentView: View {
    @State var count = 0
    var body: some View {
        HStack {
            Button("-") {
                count -= 1
            }
            Text("Count: \(count)")
            Button("+") {
                count += 1
            }
        }
        .padding()
    }
}
```

When I run this on macOS with standard SwiftUI, each `body` property is run exactly once on app start. When I click a button in `App`, only `App.body` is executed and only once. When I click a button in `ContentView`, only `ContentView.body` is executed and also only once.

Using SwiftCrossUI, when I click a button in `App`, `App.body` is executed once and `ContentView.body` is executed 5 times. When I click a button in `ContentView`, `App.body` is executed once and `ContentView.body` approximately 7 times.

I have not yet tested deeper view hierarchies, but can imagine that this would exacerbate the problem.

Do you think, this should be addressed? If so, would it be OK for me to attempt a solution? I believe such a solution could mean extensive changes to `ViewGraphNode` and `View`.

Optimally, such body re-executions would run at most once per display frame. However, I don't know if getting this info for each backend (or any backend for that matter) is possible.

Examples/Sources/RandomNumberGeneratorExample/RandomNumberGeneratorApp.swift
MiaKoring: SwiftUI does have a projected value on `@Published`, but its a `Publisher<Value>` instead of a `Binding<Value>`.
And $vm.minNum currently is an error, because environment doesn’t have a wrapped value.

let me try if its an easy change to mimick that

MiaKoring:
```swift
/// A projection of the environment object that creates bindings to its
    /// properties using dynamic member lookup.
    ///
    /// Use the projected value to pass an environment object down a view
    /// hierarchy.
    @MainActor @preconcurrency public var projectedValue: EnvironmentObject<ObjectType>.Wrapper { get }
```
  Dynamic member lookup for that? the heck?
  I honestly prefer `@Publisher` providing a Binding over that.
  
  It feels more clear to me (and is theoretically faster at runtime). But your call

Sources/SwiftCrossUI/Environment/Environment.swift
MiaKoring: keyPath.debugDescription is not available on our older platforms, so I will do interpolation instead

Examples/Sources/RandomNumberGeneratorExample/RandomNumberGeneratorApp.swift
MiaKoring: If the decision falls for dynamic member lookup, Binding also needs to get it, so it keeps working with `@State`

Noted here so its not forgotten

stackotter:
> I have some news, but first of all: Is this a good place to discuss this? Maybe an issue would be a better place? Until a better place is found, here is the news:

Yeah I think an issue would be better because this PR will get merged soon and then the discussion will be kind of hidden for anyone trying to find it back. Feel free to open an issue about supporting the `@Observable` macro.

> However, I have found an issue: body properties of Views are executed many times for a single update.

Yeah that's a known issue. I have ideas for addressing it, but pursued other optimisations first because computing view bodies wasn't the bottleneck for performance at that time (and it probably still isn't). Since my restructure of the layout system at the end of last year, the layout system should be able to properly benefit from state dependency tracking once we implement it, so it'll definitely be more worthwhile.

The way that I believe we should address the issue is by introducing a 'dirty' flag on view graph nodes that we propagate when we receive a state update (flagging any views that depend on the updated state), and then only recomputing 'dirty' views. While recomputing the bodies of those dirty views we'd also clear their layout caches. Currently we clear the layout cache of each view on each update, because we don't have any way of knowing whether the cache should be invalidated or not.

I have some experimental code from a while ago (sitting on my laptop) that tracks the state properties accessed by each view (quite precisely). I can share that if you'd like to give the state dependency tracking issue a go, but I will warn that it'd be quite an involved first contribution.

Examples/Sources/RandomNumberGeneratorExample/RandomNumberGeneratorApp.swift
stackotter: Ah ok, yeah didn't realise that Environment didn't have a projectValue implementation, but it makes sense that it doesn't because most environment non-objects are read only. I agree with you that we should keep it the way it is

Sources/SwiftCrossUI/Environment/Environment.swift
```diff
@@ -57,7 +57,16 @@ public struct Environment<Value>: DynamicProperty {
         guard let value = value.value else {
             fatalError(
                 """
-                Environment value at \(mode) used before initialization. Don't \
+                Environment value at \(
+                    {
```
stackotter: Did you mean for this to be an immediately invoked closure? I believe that this would print `Environment value at (Function) used before ...`.

I think it'd be best to compute this property description as a `let` variable before the `fatalError` statement just to make things a bit cleaner.

MiaKoring: Oh, yeah, good catch haha
I meant to add () at the end, but variable before is indeed cleaner

Sources/SwiftCrossUI/Environment/Environment.swift
```diff
@@ -55,18 +55,18 @@ public struct Environment<Value>: DynamicProperty {
...
     public var wrappedValue: Value {
         guard let value = value.value else {
+            let pathDescription = {
+                return switch mode {
+                    case .keyPath(let keyPath):
+                        "\(keyPath)"
+                    case .observableObject:
+                        "\(Value.self).self"
+                }
+            }()
```
stackotter: Use a switch expression or assign to the variable in each case (I'm generally not a fan of immediately-invoked closures)

stackotter:
Will wait for CI to finish, so that probably means that I'll merge it tomorrow.

PR #396, opened 2026-01-31
Add MusicPlayerExample

stackotter:
This PR introduces MusicPlayerExample which is a basic MP3 player. Users can create playlists and load mp3/wav files off their file system. Everything gets persisted to disk. Audio playback is supported by [swift-miniaudio](https://github.com/stackotter/swift-miniaudio).


mba6comma1:
Once you've merged in #392, do you suppose you'd be able to use that for persistence instead? Or is this persisting too much data for that to be appropriate?

stackotter:
I’ll have to look into it, but I feel like it’s probably too much data for AppStorage to be appropriate. But I’m not sure, the threshold might be different to SwiftUI due to your Codable support

MiaKoring:
If you are interested we could try Vein ^^

although that might be less than optimal due to version compatibility 🥲

I really need to expand it at some point. Might have to give up traits

mba6comma1:
@MiaKoring If you _really_ wanted, you could write an `AppStorageProvider` that interfaces with Vein!

PR #430, opened 2026-02-08
[FEAT] @Entry

MiaKoring:
### Summary
Adds the `@Entry` macro to streamline defining custom values for EnvironmentValues and AppStorageValues, replacing the need for developer written boilerplate of manual Key protocol implementation and accessor logic.

### Usage:
```swift
extension EnvironmentValues {
    @Entry
    var foo = Foo()
}

extension AppStorageValues {
    @Entry
    var foo = Foo()
}
```

### Changes
- New `@Entry` Macro: Automates key generation and storage accessors.
- AppStorage Update: Added `init(_: WritableKeyPath<AppStorageValues, Value>)` to support the new key-path based lookup.
- Internal Refactor: Migrated most SCUI EnvironmentValues to `@Entry`.
- Performance Optimization: Kept high-frequency values (e.g., `allowLayoutCaching`) as stored properties to avoid dictionary lookup overhead.
- Examples & Tests: Updated GreetingGeneratorExample and ControlsExample; added comprehensive unit tests.

### Notes
Due to `AppStorageValues` not being self-contained, it uses different accessors than `EnvironmentValues` and also has an additional value in its Key.

Stored and `@Entry` properties are fully interchangeable, so in case a certain value causes performance issues it can just be moved back to being stored.

### DIFF
--- a/Sources/SwiftCrossUI/Environment/EnvironmentValues.swift
+++ b/Sources/SwiftCrossUI/Environment/EnvironmentValues.swift
@@ -287,6 +287,8 @@ extension EnvironmentValues {
     /// The menu ordering to use.
     @Entry public var menuOrder: MenuOrder = .automatic
 
+    /// Backing store for ``EnvironmentValues/openWindowFunctionsByID``.
+    /// Used to resolve "non-sendable type" warnings in Swift 5 and errors in Swift 6 language mode.
     @Entry private var openWindowFunctionsByIDStore = UncheckedSendable(
         wrappedValue: Box<[String: @MainActor () -> Void]>([:]))
 
@@ -300,6 +302,8 @@ extension EnvironmentValues {
         }
     }
 
+    /// Backing store for ``EnvironmentValues/window``.
+    /// Used to resolve "non-sendable type" warnings in Swift 5 and errors in Swift 6 language mode.
     @Entry private var windowStore = UncheckedSendable<Any?>(wrappedValue: nil)
 
     /// The backend's representation of the window that the current view is
@@ -314,6 +318,8 @@ extension EnvironmentValues {
         }
     }
 
+    /// Backing store for ``EnvironmentValues/sheet``.
+    /// Used to resolve "non-sendable type" warnings in Swift 5 and errors in Swift 6 language mode.
     @Entry private var sheetStore = UncheckedSendable<Any?>(wrappedValue: nil)
 
     /// The backend's representation of the sheet that the current view is
--- a/Sources/SwiftCrossUI/State/AppStorage/AppStorage.swift
+++ b/Sources/SwiftCrossUI/State/AppStorage/AppStorage.swift
@@ -1,7 +1,7 @@
 import Foundation
 import Mutex
 
-private let appStorageCache: Mutex<[String: any Codable & Sendable]> = Mutex([:])
+let appStorageCache: Mutex<[String: any Codable & Sendable]> = Mutex([:])
 private let appStoragePublisherCache: Mutex<[String: Publisher]> = Mutex([:])
 
 /// Like ``State``, but persists its value to disk so that it survives betweeen
@@ -42,7 +42,7 @@ public struct AppStorage<Value: Codable & Sendable>: ObservableProperty {
                         }
                         return provider.getValue(key: key, defaultValue: defaultValue)
                     case .path(let keyPath):
-                        return AppStorageValues(__provider: provider)[keyPath: keyPath]
+                        return AppStorageValues(provider: provider)[keyPath: keyPath]
                 }
             }
 
@@ -60,7 +60,7 @@ public struct AppStorage<Value: Codable & Sendable>: ObservableProperty {
                     case .key(let key, _):
                         provider.setValue(key: key, newValue: newValue)
                     case .path(let keyPath):
-                        var values = AppStorageValues(__provider: provider)
+                        var values = AppStorageValues(provider: provider)
                         values[keyPath: keyPath] = newValue
                 }
 
@@ -139,55 +139,5 @@ extension AppStorage {
 
     public init(_ keyPath: WritableKeyPath<AppStorageValues, Value>) {
         implementation = StateImpl(initialStorage: Storage(mode: .path(keyPath)))
-    }
-}
-
-extension AppStorageProvider {
-    public func getValue<T: Codable & Sendable>(key: String, defaultValue: T) -> T {
-        return appStorageCache.withLock { cache in
-            // If this is the very first time we're reading from this key, it won't
-            // be in the cache yet. In that case, we return the already-persisted value
-            // if it exists, or the default value otherwise; either way, we add it to the
-            // cache so subsequent accesses of `value` won't have to read from disk again.
-            guard let cachedValue = cache[key] else {
-                let value =
-                    self.retrieveValue(ofType: T.self, forKey: key) ?? defaultValue
-                cache[key] = value
-                return value
-            }
-
-            // Make sure that we have the right type.
-            guard let cachedValue = cachedValue as? T else {
-                logger.warning(
-                    "'@AppStorage' property is of the wrong type; using default value",
-                    metadata: [
-                        "key": "\(key)",
-                        "providedType": "\(T.self)",
-                        "actualType": "\(type(of: cachedValue))",
-                    ]
-                )
-                return defaultValue
-            }
-
-            return cachedValue
-        }
-    }
-
-    public func setValue<T: Codable & Sendable>(key: String, newValue: T) {
-        appStorageCache.withLock { cache in
-            cache[key] = newValue
-            do {
-                logger.trace("persisting '\(newValue)' for '\(key)'")
-                try self.persistValue(newValue, forKey: key)
-            } catch {
-                logger.warning(
-                    "failed to encode '@AppStorage' data",
-                    metadata: [
-                        "value": "\(newValue)",
-                        "error": "\(error.localizedDescription)",
-                    ]
-                )
-            }
-        }
     }
 }
--- a/Sources/SwiftCrossUI/State/AppStorage/AppStorageKey.swift
+++ b/Sources/SwiftCrossUI/State/AppStorage/AppStorageKey.swift
@@ -1,5 +1,6 @@
-/// A type safe key for ``AppStorage`` properties, similar in spirit
+/// A type safe key for ``AppStorageValues`` properties, similar in spirit
 /// to ``EnvironmentKey``.
+/// Properties can be accessed using the ``AppStorage`` property wrapper.
 public protocol AppStorageKey<Value> {
     associatedtype Value: Codable
 
--- a/Sources/SwiftCrossUI/State/AppStorage/AppStorageValues.swift
+++ b/Sources/SwiftCrossUI/State/AppStorage/AppStorageValues.swift
@@ -1,17 +1,18 @@
+/// A container for typed application-defined app storage values
 public struct AppStorageValues {
-    private let __provider: AppStorageProvider?
+    private let provider: AppStorageProvider?
 
     /// Only to be used by AppStorage
-    internal init(__provider: AppStorageProvider?) {
-        self.__provider = __provider
+    internal init(provider: AppStorageProvider?) {
+        self.provider = provider
     }
 
-    public func __getValue<T: Codable & Sendable>(_ key: any AppStorageKey<T>.Type) -> T {
-        guard let __provider else { return key.defaultValue }
-        return __provider.getValue(key: key.name, defaultValue: key.defaultValue)
+    public func getValue<T: Codable & Sendable>(_ key: any AppStorageKey<T>.Type) -> T {
+        guard let provider else { return key.defaultValue }
+        return provider.getValue(key: key.name, defaultValue: key.defaultValue)
     }
 
-    public func __setValue<T: Codable & Sendable>(_ key: any AppStorageKey<T>.Type, newValue: T) {
-        __provider?.setValue(key: key.name, newValue: newValue)
+    public func setValue<T: Codable & Sendable>(_ key: any AppStorageKey<T>.Type, newValue: T) {
+        provider?.setValue(key: key.name, newValue: newValue)
     }
 }
--- a/Sources/SwiftCrossUI/State/Entry.swift
+++ b/Sources/SwiftCrossUI/State/Entry.swift
@@ -1,6 +1,6 @@
-/// Creates an environment values, or appstorage values entry.
+/// Creates an ``EnvironmentValues``, or ``AppStorageValues`` entry.
 ///
-/// Create EnvironmentValues entries by extending the EnvironmentValues structure with new properties and attaching the @Entry macro to the variable declarations:
+/// You can create ``EnvironmentValues`` entries by extending the ``EnvironmentValues`` struct with `@Entry`-annotated properties:
 /// ```swift
 /// extension EnvironmentValues {
 ///     @Entry var myCustomValue: String = "Default value"
@@ -8,7 +8,7 @@
 /// }
 /// ```
 ///
-/// Create AppStorage entries by extending the AppStorageValues structure with new properties and attaching the @Entry macro to the variable declarations:
+/// You can create ``AppStorageValues`` entries by extending the ``AppStorageValues`` struct with `@Entry`-annotated properties:
 /// ```swift
 /// extension AppStorageValues {
 ///     @Entry var myCustomValue: String = "Default value"
--- a/Sources/SwiftCrossUIMacrosPlugin/EntryMacro.swift
+++ b/Sources/SwiftCrossUIMacrosPlugin/EntryMacro.swift
@@ -21,13 +21,15 @@ public struct EntryMacro: AccessorMacro, PeerMacro {
         let getterContent: String
         let setterContent: String
 
+        let trimmedIdentifier = identifier.trimmingCharacters(in: ["`"])
+
         switch enclosingType {
             case .environment:
-                getterContent = "self[__Key_\(identifier).self]"
-                setterContent = "self[__Key_\(identifier).self] = newValue"
+                getterContent = "self[`__Key_\(trimmedIdentifier)`.self]"
+                setterContent = "self[`__Key_\(trimmedIdentifier)`.self] = newValue"
             case .appStorage:
-                getterContent = "__getValue(__Key_\(identifier).self)"
-                setterContent = "__setValue(__Key_\(identifier).self, newValue: newValue)"
+                getterContent = "getValue(`__Key_\(trimmedIdentifier)`.self)"
+                setterContent = "setValue(`__Key_\(trimmedIdentifier)`.self, newValue: newValue)"
         }
 
         return [
@@ -64,19 +66,23 @@ public struct EntryMacro: AccessorMacro, PeerMacro {
             return []
         }
 
+        let escapedIdentifier = identifier.replacingOccurrences(of: "\"", with: "\\\"")
+
         // AppStorage has got a special requirement to know the key name as string
         let nameDeclaration: String
         switch enclosingType {
             case .environment:
                 nameDeclaration = ""
             case .appStorage:
-                nameDeclaration = "\nstatic let name = \"\(identifier)\""
+                nameDeclaration = "\nstatic let name = \"\(escapedIdentifier)\""
         }
+
+        let trimmedIdentifier = identifier.trimmingCharacters(in: ["`"])
 
         return [
             DeclSyntax(
                 stringLiteral: """
-                    private struct __Key_\(identifier): \(enclosingType.keyName) {
+                    private struct `__Key_\(trimmedIdentifier)`: \(enclosingType.keyName) {
                         \(defaultValueDeclaration)\(nameDeclaration)
                     } 
                     """)
@@ -98,7 +104,7 @@ public struct EntryMacro: AccessorMacro, PeerMacro {
                 rawValue: extensionDecl.extendedType.trimmedDescription)
         else {
             throw MacroError(
-                "@Entry-annotated properties must be direct children of an EnvironmentValues or AppStorageValues extension."
+                "@Entry-annotated properties must be direct children of EnvironmentValues or AppStorageValues extensions."
             )
         }
 
@@ -122,7 +128,7 @@ public struct EntryMacro: AccessorMacro, PeerMacro {
         }
 
         // Verify defaultValue
-        var defaultValueDeclaration = ""
+        let defaultValueDeclaration: String
         if patternBinding.initialValue == nil,
             patternBinding.type?.isOptional == true
         {
--- a/Sources/SwiftCrossUIMacrosPlugin/Utils/VariableExtension.swift
+++ b/Sources/SwiftCrossUIMacrosPlugin/Utils/VariableExtension.swift
@@ -7,7 +7,7 @@ extension Variable {
     // Portions of this code are derived from the Swift Open Source Project.
     // Copyright (c) 2014 - 2024 Apple Inc. and the Swift project authors
     // Licensed under Apache License v2.0 with Runtime Library Exception
-    // Source https://github.com/swiftlang/swift/blob/2e8977f
+    // Source https://github.com/swiftlang/swift/blob/2e8977f/lib/Macros/Sources/ObservationMacros/Extensions.swift#L111
     func hasMacroApplication(_ name: String) -> Bool {
         for attribute in _syntax.attributes {
             switch attribute {
--- a/Tests/SwiftCrossUITests/EntryMacroTests.swift
+++ b/Tests/SwiftCrossUITests/EntryMacroTests.swift
@@ -22,14 +22,14 @@ struct EntryMacroTests {
             extension EnvironmentValues {
                 var test {
                     get {
-                        self[__Key_test.self]
+                        self[`__Key_test`.self]
                     }
                     set {
-                        self[__Key_test.self] = newValue
+                        self[`__Key_test`.self] = newValue
                     }
                 }
             
-                private struct __Key_test: SwiftCrossUI.EnvironmentKey {
+                private struct `__Key_test`: SwiftCrossUI.EnvironmentKey {
                     static let defaultValue = 22
                 }
             }
@@ -53,14 +53,14 @@ struct EntryMacroTests {
             extension EnvironmentValues {
                 var test: UInt64 {
                     get {
-                        self[__Key_test.self]
+                        self[`__Key_test`.self]
                     }
                     set {
-                        self[__Key_test.self] = newValue
+                        self[`__Key_test`.self] = newValue
                     }
                 }
 
-                private struct __Key_test: SwiftCrossUI.EnvironmentKey {
+                private struct `__Key_test`: SwiftCrossUI.EnvironmentKey {
                     static let defaultValue: UInt64 = 22
                 }
             }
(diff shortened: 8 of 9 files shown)


Examples/Sources/GreetingGeneratorExample/GreetingGeneratorApp.swift
```diff
@@ -68,8 +68,6 @@ struct LatestGreetingKey: EnvironmentKey {
 }
```
stackotter: I believe that this can be removed

Sources/SwiftCrossUI/Environment/EnvironmentValues.swift
```diff
@@ -44,89 +24,27 @@ public struct EnvironmentValues {
...
-    private var extraValues: [ObjectIdentifier: Any]
-
-    /// A mapping of window IDs to functions that open the corresponding windows.
-    var openWindowFunctionsByID: Box<[String: @MainActor () -> Void]>
+    private var values: [ObjectIdentifier: Any]
 
-    /// Backing storage for observable subscript
-    private var observableObjects: [ObjectIdentifier: any ObservableObject]
+    /// Whether user interaction is enabled. Set by ``View/disabled(_:)``.
+    public var isEnabled: Bool = true
```
stackotter: Why isn't this one an entry? It only changes when the `disabled` modifier is used right?

Sources/SwiftCrossUIMacrosPlugin/Utils/VariableExtension.swift
```diff
@@ -0,0 +1,22 @@
...
+                    ] {
+                        return true
+                    }
+                default:
+                    break
+            }
+        }
+        return false
+    }
+}
```
stackotter, line 26: In the `CustomCodableMacro` from MacroToolkit's example plugin, we perform a similar check simply by doing `guard let attribute = variable.attributes.first(called: name) else { ... }`

Sources/SwiftCrossUI/State/StateImpl.swift
```diff
@@ -1,4 +1,4 @@
-struct StateImpl<Storage: StateStorageProtocol> {
+public struct StateImpl<Storage: StateStorageProtocol> {
```
stackotter: Does this have to be made public? It's an internal implementation detail that I'd rather not expose if we don't have to. To fully support user-defined dynamic stateful properties we'll probably do something similar to SwiftUI.

Sources/SwiftCrossUI/State/StateImpl.swift
```diff
@@ -58,7 +58,7 @@ struct StateImpl<Storage: StateStorageProtocol> {
     }
 }
 
-protocol StateStorageProtocol: AnyObject {
+public protocol StateStorageProtocol: AnyObject {
```
stackotter: Same as for StateImpl, does this have to be made public?

Sources/SwiftCrossUI/State/AppStorage.swift
```diff
@@ -176,3 +203,21 @@ public protocol AppStorageKey<Value> {
     /// The default value for the key.
     static var defaultValue: Value { get }
 }
+
+public struct AppStorageValues {
```
stackotter: Move this to a separate file given that it's public and not completely trivial. We should probably do that with AppStorageKey as well, but that's a previous PR's fault.

Sources/SwiftCrossUI/State/AppStorage.swift
```diff
@@ -164,6 +136,61 @@ extension AppStorage {
     public init<Key: AppStorageKey<Value>>(_: Key.Type) {
         self.init(wrappedValue: Key.defaultValue, Key.name)
     }
+
+    public init(_ keyPath: WritableKeyPath<AppStorageValues, Value>) {
+        implementation = StateImpl(initialStorage: Storage(mode: .path(keyPath)))
+    }
+}
+
+// MARK: - AppStorageProviderExtension
```
stackotter: Remove this mark

Sources/SwiftCrossUI/State/AppStorage.swift
```diff
@@ -164,6 +136,61 @@ extension AppStorage {
...
+extension AppStorageProvider {
+    public func getValue<T: Codable>(key: String, defaultValue: T) -> T {
```
stackotter: Require T to be Sendable to silence the warnings.

Sources/SwiftCrossUI/State/AppStorage.swift
```diff
@@ -164,6 +136,61 @@ extension AppStorage {
...
+                    ]
+                )
+                return defaultValue
+            }
+
+            return cachedValue
+        }
+    }
+
+    public func setValue<T: Codable>(key: String, newValue: T) {
```
stackotter: Require T to be Sendable to silence the warnings.

Sources/SwiftCrossUI/State/AppStorage.swift
```diff
@@ -176,3 +203,21 @@ public protocol AppStorageKey<Value> {
...
+
+public struct AppStorageValues {
+    private let __provider: AppStorageProvider?
+
+    /// Only to be used by AppStorage
+    internal init(__provider: AppStorageProvider?) {
+        self.__provider = __provider
+    }
+
+    public func __getValue<T: Codable>(_ key: AppStorageKey<T>.Type) -> T {
```
stackotter: Fix these warnings with `any`

Sources/SwiftCrossUI/State/AppStorage.swift
```diff
@@ -176,3 +203,21 @@ public protocol AppStorageKey<Value> {
...
+        guard let __provider else { return key.defaultValue }
+        return __provider.getValue(key: key.name, defaultValue: key.defaultValue)
+    }
+
+    public func __setValue<T: Codable>(_ key: AppStorageKey<T>.Type, newValue: T) {
```
stackotter: Add `any` to fix the warnings

Sources/SwiftCrossUI/State/AppStorage/AppStorage.swift
```diff
@@ -164,6 +136,61 @@ extension AppStorage {
...
         self.init(wrappedValue: Key.defaultValue, Key.name)
     }
+
+    public init(_ keyPath: WritableKeyPath<AppStorageValues, Value>) {
+        implementation = StateImpl(initialStorage: Storage(mode: .path(keyPath)))
+    }
+}
+
+// MARK: - AppStorageProviderExtension
+extension AppStorageProvider {
```
stackotter: Can this be in the AppStorageProvider file?

Tests/SwiftCrossUITests/EntryMacroTests.swift
```diff
@@ -0,0 +1,198 @@
...
+            """,
+            expandedSource: """
+            extension EnvironmentValues {
+                var test: UInt64
+            
+                private struct __Key_test: SwiftCrossUI.EnvironmentKey {
+            
+                }
+            }
+            """,
```
stackotter, line 87: Can we omit the `expandedSource` here given that we don't actually care what it is?

Tests/SwiftCrossUITests/Utils/TestFailureSpecExtension.swift
```diff
@@ -0,0 +1,8 @@
+import SwiftSyntaxMacrosGenericTestSupport
+import Testing
+
+extension TestFailureSpec {
+    public var issueComment: Comment {
+        "\(message) \nat: \(location.filePath) \(location.line):\(location.column)"
```
stackotter, line 6: Might be best to keep the comment to a single line; e.g. `\(message) (at: ...)`

Sources/SwiftCrossUIMacrosPlugin/EntryMacro.swift
```diff
@@ -0,0 +1,136 @@
...
+                    private struct __Key_\(identifier): SwiftCrossUI.\(keyName.rawValue)Key {
+                        \(defaultValueDeclaration)\(nameDeclaration)
+                    } 
+                    """)
+        ]
+
+        return []
+    }
+
+    enum KeyName: String {
```
stackotter: I don't really understand the name of this enum. `EnclosingType` might be better? (because these aren't key names)

Sources/SwiftCrossUIMacrosPlugin/EntryMacro.swift
```diff
@@ -0,0 +1,136 @@
...
+    public static func expansion(
+        of node: SwiftSyntax.AttributeSyntax,
+        providingAccessorsOf declaration: some SwiftSyntax.DeclSyntaxProtocol,
+        in context: some SwiftSyntaxMacros.MacroExpansionContext
+    ) throws -> [SwiftSyntax.AccessorDeclSyntax] {
+        guard
+            let extensionDecl = context.lexicalContext.compactMap({ syntax in
+                syntax.as(ExtensionDeclSyntax.self)
+            }).first
+        else { return [] }
```
stackotter: We want this macro to fail if the extension isn't the direct parent of the `@Entry`-annotated property, so we should require that here.

E.g. you current implementation might accept this and then fail in a confusing way later on during compilation;

```swift
extension EnvironmentValues {
    struct Asdf {
        @Entry var name: String?
    }

    // or

    var asdf: String? {
        @Entry var hello: String?
    }
}
```

stackotter: We should also throw a `MacroError` here with a description like "`@Entry`-annotated properties must be direct children of an EnvironmentValues or AppStorageValues extension."

Sources/SwiftCrossUIMacrosPlugin/EntryMacro.swift
```diff
@@ -0,0 +1,136 @@
...
+        let extendedType = extensionDecl.extendedType.trimmedDescription
+
+        guard let keyName = KeyName(rawValue: extendedType) else { return [] }
```
stackotter: Throw an error

Sources/SwiftCrossUIMacrosPlugin/EntryMacro.swift
```diff
@@ -0,0 +1,136 @@
...
+
+        guard
+            let variable = Decl(declaration).asVariable,
+            let patternBinding = destructureSingle(variable.bindings),
+            let identifier = patternBinding.identifier
+        else { return [] }
```
stackotter: Throw an error

Sources/SwiftCrossUIMacrosPlugin/EntryMacro.swift
```diff
@@ -0,0 +1,136 @@
...
+
+        if patternBinding.initialValue == nil,
+            patternBinding.type?.isOptional != true
+        {
+            throw MacroError("@Entry requires an initial value to be set.")
```
stackotter: `@Entry requires an initial value for non-optional properties`

Sources/SwiftCrossUIMacrosPlugin/EntryMacro.swift
```diff
@@ -0,0 +1,136 @@
...
+        of node: SwiftSyntax.AttributeSyntax,
+        providingPeersOf declaration: some SwiftSyntax.DeclSyntaxProtocol,
+        in context: some SwiftSyntaxMacros.MacroExpansionContext
+    ) throws -> [SwiftSyntax.DeclSyntax] {
+        // Information about extension context
+        guard
+            let extensionDecl = context.lexicalContext.compactMap({ syntax in
+                syntax.as(ExtensionDeclSyntax.self)
+            }).first
+        else { return [] }
```
stackotter: Same as comment at top of previous method. We want direct parent to be an extension declaration.

Sources/SwiftCrossUIMacrosPlugin/EntryMacro.swift
```diff
@@ -0,0 +1,136 @@
...
+        let extendedType = extensionDecl.extendedType.trimmedDescription
+
+        guard let keyName = KeyName(rawValue: extendedType) else { return [] }
```
stackotter: Throw error (same for other silent early exits below)

Sources/SwiftCrossUIMacrosPlugin/EntryMacro.swift
```diff
@@ -0,0 +1,136 @@
...
+
+        // Optional types get nil as default value if no initial value is provided
+        var defaultValueDeclaration = ""
+        if patternBinding.initialValue == nil,
+            patternBinding.type?.isOptional == true
+        {
+            defaultValueDeclaration = "static let defaultValue\(typeDeclaration) = nil"
+        } else if let initialValue = patternBinding.initialValue?._syntax.trimmedDescription {
+            defaultValueDeclaration = "static let defaultValue\(typeDeclaration) = \(initialValue)"
+        }  // No else case here, as it already throws at the accessor macro
```
stackotter: It'd be good to throw here so that the `defaultValueDeclaration` variable doesn't have to have the theoretically never-used `""` initial value (and can be a `let`). We also can't assume the ordering of these macro expansions, so this one might get expanded first and then cause type checking errors before the other one gets expanded.

Sources/SwiftCrossUIMacrosPlugin/EntryMacro.swift
```diff
@@ -0,0 +1,136 @@
...
+            case .environment:
+                nameDeclaration = ""
+            case .appStorage:
+                nameDeclaration = "\nstatic let name = \"\(identifier)\""
+        }
+
+        return [
+            DeclSyntax(
+                stringLiteral: """
+                    private struct __Key_\(identifier): SwiftCrossUI.\(keyName.rawValue)Key {
```
stackotter: Add a `keyName` computed property to `KeyName` (which I've suggested renaming to `EnclosingType`, and return `"SwiftCrossUI.\(self.rawValue)Key"` from there to centralize things nicely.

Sources/SwiftCrossUIMacrosPlugin/EntryMacro.swift
```diff
@@ -0,0 +1,136 @@
...
+                        \(defaultValueDeclaration)\(nameDeclaration)
+                    } 
+                    """)
+        ]
+
+        return []
```
stackotter: Remove unused return

Sources/SwiftCrossUI/Environment/EnvironmentValues.swift
```diff
@@ -328,10 +207,123 @@ public struct EnvironmentValues {
...
 /// A key that can be used to extend the environment with new properties.
 public protocol EnvironmentKey<Value> {
     /// The type of value the key can hold.
     associatedtype Value
     /// The default value for the key.
     static var defaultValue: Value { get }
 }
+
+@attached(accessor) @attached(peer, names: prefixed(__Key_))
+public macro Entry() =
```
stackotter: Please move this macro declaration to a separate file as well (Entry.swift), especially given that it's not just an EnvironmentValues thing. And add a documentation comment with example usage code for both EnvironmentValues and AppStorageValues.

Sources/SwiftCrossUI/Environment/EnvironmentValues.swift
MiaKoring: I had concerns about the hashing overhead, as its one of the more often accessed properties, I can change it and we see if it becomes a bottleneck at some point?

Sources/SwiftCrossUI/State/StateImpl.swift
MiaKoring: I just thought it may be useful for custom implementations, but as it isn't the final idea I will make it internal again

Tests/SwiftCrossUITests/Utils/TestFailureSpecExtension.swift
MiaKoring: the message is really long with code and diff, I originally had it in the same line, but the message would usually be a long multiline string, so I felt like another line was better. you can try by removing a character in the expanded source, making it fail

Sources/SwiftCrossUIMacrosPlugin/EntryMacro.swift
MiaKoring: You are right, that is better. The thought was that of xyzKey xyz would be the name (Environment would be the name, appended by Key becoming EnvironmentKey)

Sources/SwiftCrossUI/Environment/EnvironmentValues.swift
stackotter: Lots of other properties (such as the font) are accessed a similar amount, so yeah we should probably leave this particular optimisation for future performance work

stackotter:
Oh and UIKitBackend has a custom environment value at around `Sources/UIKitBackend/KeyboardToolbar.swift:300`

Sources/SwiftCrossUIMacrosPlugin/Utils/VariableExtension.swift
MiaKoring:
```
guard let targetAttribute = attributes.first(called: name) else { return false }
return targetAttribute.asMacroAttribute != nil
```
doesn’t work. I’m not entirely sure WHY it doesn’t. But the code I stole from https://github.com/swiftlang/swift/blob/main/lib/Macros/Sources/ObservationMacros/Extensions.swift does.
(In my next push I’ll have attribution, I forgot about the Apache V 2 License)

Sources/SwiftCrossUI/State/AppStorage/AppStorage.swift
MiaKoring: It cannot, unless 
```swift
private let appStorageCache: Mutex<[String: any Codable & Sendable]> = Mutex([:])
```
isn’t private anymore

Tests/SwiftCrossUITests/EntryMacroTests.swift
MiaKoring: We cannot, expandedSource is a non-optional parameter

Sources/SwiftCrossUIMacrosPlugin/EntryMacro.swift
MiaKoring: I’m not convinced it makes sense to throw the same error twice. 
We would need to specify it 2 times in the tests for the error and the user would also get the same message twice one the same macro application.

Acknowledgements/swift/LICENSE.txt
```diff
@@ -0,0 +1,211 @@
...
+                        http://www.apache.org/licenses/
+
+    TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+    1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
```
stackotter, line 12: Did you mean for this to be in this PR?

Sources/SwiftCrossUI/Environment/EnvironmentValues.swift
```diff
@@ -290,18 +287,46 @@ extension EnvironmentValues {
     /// The menu ordering to use.
     @Entry public var menuOrder: MenuOrder = .automatic
 
+    @Entry private var openWindowFunctionsByIDStore = UncheckedSendable(
+        wrappedValue: Box<[String: @MainActor () -> Void]>([:]))
```
stackotter, line 293: Is this pattern of a private `blahStore` and a public computed `blah` property to address compiler warnings? If so, please document each of these private `blahStore` properties to make their purpose clear.

Sources/SwiftCrossUIMacrosPlugin/Utils/VariableExtension.swift
```diff
@@ -4,6 +4,10 @@ import SwiftSyntaxMacroExpansion
 import SwiftSyntaxMacros
 
 extension Variable {
+    // Portions of this code are derived from the Swift Open Source Project.
+    // Copyright (c) 2014 - 2024 Apple Inc. and the Swift project authors
+    // Licensed under Apache License v2.0 with Runtime Library Exception
+    // Source https://github.com/swiftlang/swift/blob/2e8977f
```
stackotter: Link to the actual source of this algorithm

Sources/SwiftCrossUI/State/AppStorage/AppStorageValues.swift
```diff
@@ -0,0 +1,17 @@
+public struct AppStorageValues {
```
stackotter, line 2: Document this type. Something like `A container for typed application-defined app storage values`.

Sources/SwiftCrossUI/State/AppStorage/AppStorageKey.swift
```diff
@@ -0,0 +1,10 @@
+/// A type safe key for ``AppStorage`` properties, similar in spirit
```
stackotter: Should probably say AppStorageValues instead of AppStorage given the properties are on AppStorageValues. You could then add a sentence to explain that they are accessed using the AppStorage property wrapper.

Sources/SwiftCrossUI/State/AppStorage/AppStorageValues.swift
```diff
@@ -0,0 +1,17 @@
...
+    private let __provider: AppStorageProvider?
```
stackotter: This only just came to mind, but why the double leading underscore instead of a single leading underscore?

I've also realised that given that `__provider` is private, user-defined properties can't clash with it anyway, so it probably doesn't have to be underscored at all.

Also, from a quick local test, Swift handles properties that have the same name as a method just fine, so the methods can probably have their double underscores removed as well.

Sources/SwiftCrossUI/State/AppStorage/AppStorage.swift
stackotter: I'd be ok with removing the `private` on that cache to allow this extension to move into the AppStorageProvider file

Sources/SwiftCrossUI/State/Entry.swift
```diff
@@ -0,0 +1,23 @@
+/// Creates an environment values, or appstorage values entry.
```
stackotter: Link these to EnvironmentValues and AppStorageValues properly

Sources/SwiftCrossUI/State/Entry.swift
```diff
@@ -0,0 +1,23 @@
...
+///
+/// Create EnvironmentValues entries by extending the EnvironmentValues structure with new properties and attaching the @Entry macro to the variable declarations:
```
stackotter:
```
You can create ``EnvironmentValues`` entries by extending the ``EnvironmentValues`` struct with `@Entry`-annotated properties:
```

Use that same wording for the appstorage as well (replacing EnvironmentValues with AppStorageValues)

Sources/SwiftCrossUIMacrosPlugin/EntryMacro.swift
```diff
@@ -63,22 +51,68 @@ public struct EntryMacro: AccessorMacro, PeerMacro {
...
+        defaultValueDeclaration: String
+    ) {
+        // Verify extension context
+        guard
+            let extensionDecl = context.lexicalContext.first?.as(ExtensionDeclSyntax.self),
+            let enclosingValueType = EnclosingType(
+                rawValue: extensionDecl.extendedType.trimmedDescription)
+        else {
+            throw MacroError(
+                "@Entry-annotated properties must be direct children of an EnvironmentValues or AppStorageValues extension."
```
stackotter: of EnvironmentValues or AppStorageValues extensions

Sources/SwiftCrossUIMacrosPlugin/EntryMacro.swift
```diff
@@ -63,22 +51,68 @@ public struct EntryMacro: AccessorMacro, PeerMacro {
...
+                nameDeclaration = ""
+            case .appStorage:
+                nameDeclaration = "\nstatic let name = \"\(identifier)\""
+        }
 
-        // Information about variable
+        return [
+            DeclSyntax(
+                stringLiteral: """
+                    private struct __Key_\(identifier): \(enclosingType.keyName) {
```
stackotter: We should probably surround our key struct identifier with backticks, because from Swift 6.2 developers can have spaces in variable identifiers (when surrounded by backticks).

Sources/SwiftCrossUIMacrosPlugin/EntryMacro.swift
```diff
@@ -63,22 +51,68 @@ public struct EntryMacro: AccessorMacro, PeerMacro {
...
+        }
 
-        guard let keyName = KeyName(rawValue: extendedType) else { return [] }
+        // AppStorage has got a special requirement to know the key name as string
+        let nameDeclaration: String
+        switch enclosingType {
+            case .environment:
+                nameDeclaration = ""
+            case .appStorage:
+                nameDeclaration = "\nstatic let name = \"\(identifier)\""
```
stackotter: We should escape `"` characters in `identifier`, because with [raw identifiers](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0451-escaped-identifiers.md#permitted-characters) developers can have double quotes in variable names.

`identifier.replacingOccurrences(of: "\"", with: "\\\"")`

Sources/SwiftCrossUIMacrosPlugin/EntryMacro.swift
```diff
@@ -87,38 +121,25 @@ public struct EntryMacro: AccessorMacro, PeerMacro {
             typeDeclaration = ""
         }
 
-        // Optional types get nil as default value if no initial value is provided
+        // Verify defaultValue
         var defaultValueDeclaration = ""
```
stackotter: I think that this can be a let without a default value

Tests/SwiftCrossUITests/EntryMacroTests.swift
```diff
@@ -194,5 +190,63 @@ struct EntryMacroTests {
...
+                    line: 2,
+                    column: 5
+                )
+            ],
+            macroSpecs: testMacros,
+            failureHandler: { spec in
+                Issue.record(spec.issueComment)
+            }
+        )
+    }
```
stackotter, line 250: Add a test for backticked variable names, because I feel like they could be an interesting edge case.

stackotter (changes requested):

Sources/SwiftCrossUIMacrosPlugin/EntryMacro.swift
MiaKoring: I'm pretty sure those wouldn't be picked up with the current SwiftSyntax version (they didn't in my test function parsing)

can I update it?

stackotter: We should leave it at the current version for now because it has to be the same as Swift Bundler. I realise that it doesn’t support actually parsing raw identifiers, but when we update our Swift Syntax version there won’t be any API breaking change to remind us of this issue, and that issue will silently get introduced. So if we handle it now (cause we can without the new API, using string replacement), it’s future proof, if I’m understanding things correctly

MiaKoring: It should be, we just can't add a test, which sucks a little.

Acknowledgements/swift/LICENSE.txt
MiaKoring: Yes, I used code from the swift repo, which is Apache licensed and afaik a copy of the license is an Apache license requirement.

stackotter: Oh, sorry, I meant to remove this comment when I saw the source of the attribution 🤦‍♂️

Sources/SwiftCrossUI/State/Entry.swift
MiaKoring: done, I wonder why the Apple documentation doesn’t link to the struct declarations. (I took the comment from there)

Tests/SwiftCrossUITests/EntryMacroTests.swift
MiaKoring: Just adding the comment here again, so its not forgotten, won’t work without a swift-syntax version bump to 602.x.x

Sources/SwiftCrossUI/State/Entry.swift
stackotter: Interesting...

PR #431, opened 2026-02-08
Support SCUI_DEFAULT_BACKEND on all platforms

### LINKED ISSUE #272: Update SCUI_DEFAULT_BACKEND handling in Package.swift to allow it to override non-desktop backends
At the moment SCUI_DEFAULT_BACKEND is only respected when targeting desktop platforms, because on other platforms we have to use SwiftPM conditionals. To allow SCUI_DEFAULT_BACKEND to override the backend used when building for a non-desktop platform, we should update our Package.swift logic to unconditionally use defaultBackend when it gets set by the user.

Should be a simple fix, just wanted to keep it separate from my current changes cause it feels possible to mess up.

bbrk24:
Fixes #272

I also got rid of the `#if`s in the backend selection logic in favor of SwiftPM-level conditionals.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -27,17 +27,16 @@ if let version = getGtk4MinorVersion(), version >= 10 {
 }
 
 let env = ProcessInfo.processInfo.environment
-let defaultBackend: String
+let defaultBackendDependencies: [Target.Dependency]
 if let backend = env["SCUI_DEFAULT_BACKEND"] {
-    defaultBackend = backend
+    defaultBackendDependencies = [.target(name: backend)]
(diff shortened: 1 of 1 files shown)


bbrk24:
Okay I don't understand why UIKit CI is suddenly trying to compile WinUI and GTK code

stackotter:
I agree that this would be a better way of doing things. I have no clue why it's not working though. Gonna change to a draft PR now so that I know it's not ready.

stackotter (approved):
Looks good, thanks!

Issue #438, opened 2026-02-09
Text() doesn’t wrap inside ScrollView() on GtkBackend


MiaKoring:
Text() doesn’t wrap inside ScrollView() on GtkBackend, in a normal VStack outside of a scrollview it wraps perfectly, inside its just a single line, always. 

This might be an issue with `GtkBackend/CustomLabel/setSizeRequest` not perfectly working inside ScrollView:

> Override the label's layout height. We do this so that the label grows
    vertically to fill available space even though we have ellipsizing
    enabled (which generally causes labels to limit themselves to a single line).

LineWrapReproducer in Examples shows the issue
https://github.com/MiaKoring/swift-cross-ui/tree/linewrap-reproducer

stackotter:
Can confirm that this is a bug

GtkBackend (bugged):

<img width="548" height="353" alt="Image" src="https://github.com/user-attachments/assets/5a615a9a-222b-418a-b743-e1ebecf3920d" />

AppKitBackend (correct):

<img width="512" height="340" alt="Image" src="https://github.com/user-attachments/assets/6bcef11c-f884-4eba-91a1-f94c79ea2531" />

```swift
import DefaultBackend
import SwiftCrossUI

@main
struct BugApp: App {
    var body: some Scene {
        WindowGroup("BugApp") {
            ScrollView {
                Text("Lorem ipsum dolor sit amet. Lorem ipsum dolor sit amet. Lorem ipsum dolor sit amet. Lorem ipsum dolor sit amet. Lorem ipsum dolor sit amet. Lorem ipsum dolor sit amet. Lorem ipsum dolor sit amet. Lorem ipsum dolor sit amet. Lorem ipsum dolor sit amet. Lorem ipsum dolor sit amet. Lorem ipsum dolor sit amet. Lorem ipsum dolor sit amet. Lorem ipsum dolor sit amet. Lorem ipsum dolor sit amet. Lorem ipsum dolor sit amet. Lorem ipsum dolor sit amet. Lorem ipsum dolor sit amet. ")
            }
        }
    }
}
```

MiaKoring:
This might be resolvable by using the new measurement-only widget not mounted in hierarchy I added in https://github.com/moreSwift/swift-cross-ui/pull/440
I will try and fix it in the same PR, I don’t think a separate PR makes sense for that small bug, especially as its related to the idea of the PR

MiaKoring:
[...]

PR #440, opened 2026-02-09
[FEAT] View/lineLimit modifier

### LINKED ISSUE #432: [FEAT] Control over squishing behavior of Text
In SwiftUI you can specify the maximum amount of lines a Text can have, therefore controlling its behavior when squished below its ideal size
https://developer.apple.com/documentation/swiftui/environmentvalues/linelimit

MiaKoring:
Resolves https://github.com/moreSwift/swift-cross-ui/issues/432

### Summary
Adds the `View/lineLimit` modifier, allowing to limit the vertical space a Text is allowed to occupy based on the amount of lines instead of fixed pixel values.

### Support:
- AppKitBackend
- UIKitBackend
- WinUIBackend
- GtkBackend

### Changes
- added `View/lineLimit` modifier
- added `Environment/lineLimitSettings` containing the lineLimit number and wether space should be reserved
- added `Environment/lineLimit` computed `Int?` to match SwiftUI’s counterpart for usage in `@Environment`, setting is not supported, as I see clearness issues when it gets set without `LineLimit/reservesSpace`

### Improvements
WinUI doesn’t require the single line height measurement anymore, as it now follows the SCUI dictated lineHeight of resolvedFont

### Problems
On GtkBackend there is the issue that we can’t set the lineHeight, therefore its not immediately possible to calculate the height of Text if it had n lines.

I took the approach WinUI used before of just measuring another widget:
I create a multiline string with as many lines as lineLimit specifies and measure it inside a dedicated measurement widget that gets created at the start of the apps main loop. 

Updating the widget and measuring it only happens when lineLimit is set in the environment, therefore not affecting performance unless used. lineLimit should also mostly get used with smaller numbers, so generating the string shouldn’t have too much of a performance impact

There are further issues when `Text` is rendered inside a `ScrollView`, detailed in https://github.com/moreSwift/swift-cross-ui/issues/438

### Notes
The changes can be tested using `Examples/LineLimitTest`, I see no point in keeping it around, so it should be removed before merging.

This PR is built on top of https://github.com/moreSwift/swift-cross-ui/pull/430 
It should be merged first. This one is marked as draft until then.


Examples/Bundler.toml
```diff
@@ -79,3 +79,8 @@ version = '0.1.0'
 identifier = 'dev.swiftcrossui.ColorsExample'
 product = 'ColorsExample'
 version = '0.1.0'
+
+[apps.LineLimitTest]
```
stackotter: As you mentioned in the PR description, this should be removed before merging

Sources/AppKitBackend/AppKitBackend.swift
```diff
@@ -635,9 +635,21 @@ public final class AppKitBackend: AppBackend {
             options: [.usesLineFragmentOrigin, .truncatesLastVisibleLine],
             attributes: Self.attributes(forTextIn: environment)
         )
+
+        var usedHeight = rect.size.height
```
stackotter: Rename this to `height` (as you later set it to not be the used height, but rather the height imposed by SCUI).

Sources/AppKitBackend/AppKitBackend.swift
```diff
@@ -635,9 +635,21 @@ public final class AppKitBackend: AppBackend {
...
+
+        if let lineLimitSettings = environment.lineLimitSettings {
+            let height =
```
stackotter: Rename this to `limitedHeight` to not clash with the other variable rename

Sources/SwiftCrossUI/Environment/EnvironmentValues.swift
```diff
@@ -345,6 +345,13 @@ extension EnvironmentValues {
 
     /// Whether user interaction is enabled. Set by ``View/disabled(_:)``.
     @Entry public var isEnabled: Bool = true
+
+    @Entry public var lineLimitSettings: LineLimit?
```
stackotter, line 359: Document this

Sources/SwiftCrossUI/State/StateImpl.swift
```diff
@@ -1,4 +1,4 @@
-struct StateImpl<Storage: StateStorageProtocol> {
+public struct StateImpl<Storage: StateStorageProtocol> {
```
stackotter: Make these internal again unless necessary (I think this may just be from a merging issue given that you changed this in an earlier pr?)

Sources/SwiftCrossUI/Values/LineLimit.swift
```diff
@@ -0,0 +1,4 @@
+public struct LineLimit: Sendable {
```
stackotter: Document this and its two properties. Might be best to make the type `Hashable` as well. And could be good to make the two properties `var` as well (not much reason to prevent people from mutating the properties individually afaict)

Sources/UIKitBackend/UIKitBackend+Passive.swift
```diff
@@ -70,9 +70,21 @@ extension UIKitBackend {
             options: [.usesLineFragmentOrigin, .truncatesLastVisibleLine],
             context: nil
         )
+
+        var usedHeight = size.height
```
stackotter: Similar to AppKitBackend, rename this var to `height`, and rename `height` (in the if) to `limitedHeight`

Sources/GtkBackend/GtkBackend.swift
```diff
@@ -730,7 +734,32 @@ public final class GtkBackend: AppBackend {
...
+            let multilineString = [String](repeating: "a", count: lineLimitSettings.limit)
+                .joined(separator: "\n")
+            updateTextView(
+                measurementCustomLabel,
+                content: "",
+                environment: environment
+            )
+
+            let pango = Pango(for: measurementCustomLabel)
+            let (_, potentialHeight) = pango.getTextSize(
```
stackotter: Rename `potentialHeight` to `heightLimit`

Sources/GtkBackend/GtkBackend.swift
```diff
@@ -730,7 +734,32 @@ public final class GtkBackend: AppBackend {
             proposedWidth: proposedWidth.map(Double.init),
             proposedHeight: proposedHeight.map(Double.init)
         )
-        return SIMD2(width, height)
+
+        var usedHeight = height
+
+        if let lineLimitSettings = environment.lineLimitSettings {
+            let multilineString = [String](repeating: "a", count: lineLimitSettings.limit)
```
stackotter, line 743: Is it possible for us to instead just measure a single `"a"` and then multiply the resulting height by the line limit? It'd be worth trying if you haven't already because it should be more efficient (don't need to construct a string and don't have to lay out as much).

Sources/GtkBackend/GtkBackend.swift
```diff
@@ -730,7 +734,32 @@ public final class GtkBackend: AppBackend {
...
+                content: "",
+                environment: environment
+            )
+
+            let pango = Pango(for: measurementCustomLabel)
+            let (_, potentialHeight) = pango.getTextSize(
+                multilineString,
+                ellipsize: (widget as! CustomLabel).ellipsize,
+                proposedWidth: proposedWidth.map(Double.init),
+                proposedHeight: proposedHeight.map(Double.init)
```
stackotter: Have you tested this fully on Linux? It seems that `reservesSpace = true` would not function correctly due to the text getting ellipsized. You should set `proposedWidth` and `proposedHeight` to `nil` instead because we just want to know the ideal size of the measurement string (we don't want line wrapping or ellipsizing to happen).

Sources/GtkBackend/GtkBackend.swift
```diff
@@ -730,7 +734,32 @@ public final class GtkBackend: AppBackend {
             proposedWidth: proposedWidth.map(Double.init),
             proposedHeight: proposedHeight.map(Double.init)
         )
-        return SIMD2(width, height)
+
+        var usedHeight = height
```
stackotter: Maybe rename to `imposedHeight` or something similar. To me `usedHeight` sounds like the height that the text used, rather than the height that we are going to use as our return value (which I assume is what you meant?)

stackotter (changes requested):

Sources/GtkBackend/GtkBackend.swift
MiaKoring: It is not, as gtk currently doesn’t respect the lineHeight. 
the height of a single line of “a”  * 2 is not the same as “a\na” due to some spacing(or smth like that) in between lines. 
Once lineHeight is properly respected it would be a yes.

stackotter: > Note that the actual baseline-to-baseline distance between lines of text is influenced by other factors, such as [pango_layout_set_spacing()](https://docs.gtk.org/Pango/method.Layout.set_spacing.html) and [pango_layout_set_line_spacing()](https://docs.gtk.org/Pango/method.Layout.set_line_spacing.html).

Just found that in the [get_line_height](https://docs.gtk.org/Pango/method.LayoutLine.get_height.html) docs. Might help resolve that

Sources/GtkBackend/GtkBackend.swift
MiaKoring: Screenshot 2026-02-23 at 15 36 19
True, with proposedWidth/Height it can get squished. Setting to nil prevents it like you expected. And yes, I forgot test this case on GtkBackend, I’m sorry.

Still something wrong with the alignment, as it shouldn't be centered. Any idea how to prevent it? My brain isn’t really functioning well today

Sources/GtkBackend/GtkBackend.swift
MiaKoring: doesn’t work
```swift
public func getLineHeight(_ text: String) -> Double {
    let layout = pango_layout_new(pangoContext)!
    pango_layout_set_text(layout, text, Int32(text.utf8.count))
    
    let line = pango_layout_get_line(layout, 0)
    
    var height: Int32 = 0
    pango_layout_line_get_height(line, &height)
    
    g_object_unref(UnsafeMutableRawPointer(layout))
    
    return pango_units_to_double(height)
}
```
using this * lineLimit leads to  
Screenshot 2026-02-23 at 16 31 00

(expected would be 1, 2, 3, 4)

Sources/GtkBackend/GtkBackend.swift
MiaKoring: vertical alignment now works after settign textView.yalign = 0.0 in the create function

Sources/GtkBackend/GtkBackend.swift
MiaKoring: yes

Sources/SwiftCrossUI/State/StateImpl.swift
MiaKoring: yes removing the public keyword again worked without any warnings, curious that it was still here tho, I’m pretty sure the commit removing it was newer. anyway, its gone again

Sources/GtkBackend/GtkBackend.swift
stackotter: Ah I didn’t mean to use that function, because that would just return what the line height was set to. But the docs mention ‘baseline to baseline distance’ which is what SwiftCrossUI refers to as line height. They say that you need to incorporate the pango spacing and line spacing in order to compute that distance we want.

Sources/GtkBackend/GtkBackend.swift
stackotter: No worries about not testing it, it’s easy to miss stuff when you have to test it lots of times during development

Sources/GtkBackend/GtkBackend.swift
MiaKoring: should we just pango_layout_set_spacing for text in general to 0 to match the other platforms? And later allow custom line spacing via a https://developer.apple.com/documentation/swiftui/view/linespacing(_:)

MiaKoring: gtk documentation says linespacing and spacing are 0 by default. huh? 

why doesn't it match then?

stackotter: Weird… might be worth double checking just in case

MiaKoring: Can confirm, both are 0

Sources/SwiftCrossUI/Values/LineLimit.swift
```diff
@@ -1,4 +1,8 @@
-public struct LineLimit: Sendable {
-    public let limit: Int
-    public let reservesSpace: Bool
+/// Configuration for text height limits, propagated vi ``EnvironmentValues/lineLimitSettings``.
```
stackotter: `propagated via ...`

stackotter (dismissed):
Thanks for implementing this! Super useful to have

stackotter:
Just one small typo to fix then happy to merge

stackotter (approved):

Issue #444, opened 2026-02-11
Unnecessary writes to AppStorage on other state changes


bbrk24:
The broad structure of the code is this:

```swift
@AppStorage var strs: [String]
@State var index: Int

var body: some View {
  TextField($strs[index])
 
  // control that changes index
}
```

If the content of the text field changes because the index changes, even when the array itself does not change, the app storage values are re-saved. This unnecessary save should be eliminated.

I first noticed this on GtkBackend but it may be happening with others too

stackotter:
I believe that this issue has been resolved. This is the code I used to verify that;

```swift
import DefaultBackend
import SwiftCrossUI

struct ContentView: View {
    @AppStorage("strs") var strs: [String] = ["First", "Second", "Third"]
    @State var index: Int = 0

    var body: some View {
        TextField("field", text: $strs[index])

        Button("Decrement") { index -= 1 }
        Button("Increment") { index += 1 }
    }
}

@main
struct Repro: App {
    var body: some Scene {
        WindowGroup("Repro") {
            ContentView()
        }
    }
}
```

PR #446, opened 2026-02-13
[FIX] Unnecessary writes to AppStorage

MiaKoring:
### Summary 
Resolves https://github.com/moreSwift/swift-cross-ui/issues/444 on WinUIBackend, improves on GtkBackend, others unaffected.

### How
Added a flag on widgets via an extension to disable the first changed handler call after an internal programmatic update to the content.

On Gtk those functions are:
- setState(ofToggle:_, _)
- setValue(ofSlider:_, _)
- setContent(ofTextField:_, _)
- setContent(ofTextEditor:_, _)

On WinUI:
- setValue(ofSlider:_, _)
- setContent(ofTextField:_, _)
- setContent(ofTextEditor:_, _)

### Result
The WinUI changes entirely removed unnecessary writes in my test, including `TextField`, `TextEditor`, `Slider` and `Toggle`.

On GtkBackend `Slider` and `Toggle` don’t cause unnecessary writes anymore, TextField and TextEditor only set once instead of previously 2 times.

### Notes 
On GtkBackend startup the signal gets called once. This is now eliminated. 

At runtime it gets called twice, but I'm not sure if this is true for every layout. Blocking two events is too unstable in my opinion, especially since there already is a case where it would fail (startup).

Only one instead of 2 writes on button press is an improvement, but to fully prevent it I would recommend just using State and persisting it debounced to AppStorage.

### DIFF
--- a/Sources/Gtk/GObject.swift
+++ b/Sources/Gtk/GObject.swift
@@ -23,11 +23,21 @@ open class GObject: GObjectRepresentable {
 
     private var signals: [(UInt, Any)] = []
 
-    private static let blockableSignalNames: Set<String> = [
-        "changed"
+    /// GObject signals sometimes get invoked when you programmatically set something.
+    /// If you don't want them to, you can temporarily disable them, by adding the signal name here
+    /// and wrapping the set operation in ``GObject/withBlockedSignal(named:block:)``.
+    ///
+    /// We made blocking support opt in to save memory.
+    public static let blockableSignalNames: Set<String> = [
+        "changed", "notify::active", "toggled", "value-changed",
     ]
 
-    private var blockableSignalIDs = [String: UInt]()
+    /// Stores the signal handler ID for a  signal name.
+    /// Signal handler IDs need to be stored, as `g_signal_handler_block` and `g_signal_handler_unblock`
+    /// need it to block/unblock it.
+    ///
+    /// We have no other way to retrieve them later.
+    private var blockableSignalIDs: [String: UInt] = [:]
 
     open func registerSignals() {}
 
@@ -147,7 +157,6 @@ open class GObject: GObjectRepresentable {
         storeHandler(handlerId, box: box, for: name)
     }
 
-    @inline(__always)
     private func storeHandler(_ id: UInt, box: Any, for signalName: String) {
         signals.append((id, box))
         if Self.blockableSignalNames.contains(signalName) {
@@ -156,6 +165,7 @@ open class GObject: GObjectRepresentable {
     }
 
     /// Executes a closure while temporarily suppressing a specific signal handler.
+    /// You can only block signals included in ``GObject/blockableSignalNames``.
     ///
     /// - Parameters:
     ///   - named: The name of the GObject signal to block (e.g., "changed").
@@ -166,11 +176,12 @@ open class GObject: GObjectRepresentable {
         block: @escaping () -> Void
     ) {
         guard let signalID = blockableSignalIDs[signalName] else {
-            print(
-                """
-                Notice: Skipping signal block for '\(signalName)' (ID missing). \
-                Proceeding with block execution anyway.
-                """)
+            if !Self.blockableSignalNames.contains(signalName) {
+                print(
+                    """
+                    Warning: Could not block signal '\(signalName)' because it is not included in GObject.blockableSignalNames.
+                    """)
+            }
             block()
             return
         }
--- a/Sources/GtkBackend/GtkBackend.swift
+++ b/Sources/GtkBackend/GtkBackend.swift
@@ -890,7 +890,10 @@ public final class GtkBackend: AppBackend {
 
     public func setState(ofToggle toggle: Widget, to state: Bool) {
         let toggle = toggle as! Gtk.ToggleButton
-        toggle.active = state
+
+        toggle.withBlockedSignal(named: "toggled") {
+            toggle.active = state
+        }
     }
 
     public func createSwitch() -> Widget {
@@ -910,7 +913,11 @@ public final class GtkBackend: AppBackend {
     }
 
     public func setState(ofSwitch switchWidget: Widget, to state: Bool) {
-        (switchWidget as! Gtk.Switch).active = state
+        let switchWidget = switchWidget as! Gtk.Switch
+
+        switchWidget.withBlockedSignal(named: "notify::active") {
+            switchWidget.active = state
+        }
     }
 
     public func createCheckbox() -> Widget {
@@ -930,7 +937,11 @@ public final class GtkBackend: AppBackend {
     }
 
     public func setState(ofCheckbox checkboxWidget: Widget, to state: Bool) {
-        (checkboxWidget as! Gtk.CheckButton).active = state
+        let checkboxWidget = checkboxWidget as! Gtk.CheckButton
+
+        checkboxWidget.withBlockedSignal(named: "notify::active") {
+            checkboxWidget.active = state
+        }
     }
 
     public func createSlider() -> Widget {
@@ -959,7 +970,10 @@ public final class GtkBackend: AppBackend {
 
     public func setValue(ofSlider slider: Widget, to value: Double) {
         let slider = slider as! Scale
-        slider.value = value
+
+        slider.withBlockedSignal(named: "value-changed") {
+            slider.value = value
+        }
     }
 
     public func createTextField() -> Widget {
--- a/Sources/SwiftCrossUI/Views/Checkbox.swift
+++ b/Sources/SwiftCrossUI/Views/Checkbox.swift
@@ -30,7 +30,19 @@ struct Checkbox: ElementaryView, View {
         backend: Backend
     ) {
         backend.updateCheckbox(widget, environment: environment) { newActiveState in
-            active.wrappedValue = newActiveState
+            if active.wrappedValue != newActiveState {
+                active.wrappedValue = newActiveState
+            } else {
+                #if DEBUG
+                    logger.warning(
+                        """
+                        Unnecessary write to wrappedValue binding of Checkbox detected, \
+                        please open an issue on the SwiftCrossUI GitHub repository \
+                        so we can fix it on \(type(of: backend)).
+                        """
+                    )
+                #endif
+            }
         }
         backend.setState(ofCheckbox: widget, to: active.wrappedValue)
     }
--- a/Sources/SwiftCrossUI/Views/Slider.swift
+++ b/Sources/SwiftCrossUI/Views/Slider.swift
@@ -96,7 +96,7 @@ public struct Slider: ElementaryView, View {
                 #if DEBUG
                     logger.warning(
                         """
-                        Unnecessary write to text Binding of Slider detected, \
+                        Unnecessary write to wrappedValue Binding of Slider, \
                         please open an issue on the SwiftCrossUI GitHub repository \
                         so we can fix it on \(type(of: backend)).
                         """
--- a/Sources/SwiftCrossUI/Views/ToggleButton.swift
+++ b/Sources/SwiftCrossUI/Views/ToggleButton.swift
@@ -30,7 +30,7 @@ struct ToggleButton: ElementaryView, View {
                 #if DEBUG
                     logger.warning(
                         """
-                        Unnecessary write to text Binding of TextEditor detected, \
+                        Unnecessary write to wrappedValue binding of ToggleButton detected, \
                         please open an issue on the SwiftCrossUI GitHub repository \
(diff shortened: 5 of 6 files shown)


Sources/GtkBackend/GtkBackend.swift
```diff
@@ -889,7 +893,9 @@ public final class GtkBackend: AppBackend {
     }
 
     public func setState(ofToggle toggle: Widget, to state: Bool) {
-        (toggle as! Gtk.ToggleButton).active = state
+        let toggle = toggle as! Gtk.ToggleButton
+        toggle.shouldBlockNextChangedSignal
```
bbrk24: On mobile so maybe this is a GitHub issue but this looks like an unused read, not a set?

Sources/GtkBackend/GtkBackend.swift
```diff
@@ -2028,3 +2056,29 @@ final class TimePicker: Box {
...
+                    widgetPointer.cast(),
+                    "shouldBlockNextChangedSignal",
+                    nil
+                )
+                return
+            }
+            g_object_set_data(
+                widgetPointer.cast(),
+                "shouldBlockNextChangedSignal",
+                UnsafeMutableRawPointer(bitPattern: 1)
```
bbrk24: that looks scary

MiaKoring: it really isn't, we only care if gtk gives us the pointer so its completely safe

Sources/GtkBackend/GtkBackend.swift
MiaKoring: Oh, yeah, thanks

MiaKoring:
no idea whats going on with the windows Ci

Sources/GtkBackend/GtkBackend.swift
```diff
@@ -2028,3 +2056,29 @@ final class TimePicker: Box {
...
+        set {
+            guard newValue else {
+                g_object_set_data(
+                    widgetPointer.cast(),
+                    "shouldBlockNextChangedSignal",
+                    nil
+                )
+                return
+            }
+            g_object_set_data(
```
stackotter: You might be better off using [`g_object_set`](https://docs.gtk.org/gobject/method.Object.set.html). Using `g_object_set_data` to set a fake dummy pointer feels like a hack

MiaKoring: "g_object_set is not directly available to language bindings”, which is also the case here.
And yes it is a hack, but a completely safe one

MiaKoring:
> We could avoid setting the value of the provided binding if its existing value matches the new value from the underlying widget. 

i.e. adding
```swift
getState(ofToggle: _) -> Bool 
getValue(ofSlider: _) -> Double 
```
to AppBackend and checking those before setting in `commit`?

(TextField and Editor already have get functions)

stackotter:
I meant updating the change handler passed to the backend with an additional equality check against the existing value of the binding, given that the approach of avoiding sets doesn’t seem to be working for text fields on some backends

MiaKoring:
Ah, yes, I'll do that later, sounds good

MiaKoring:
Newest commit moved the checks for TextField, Slider and Toggle to SCUI. 
GtkBackend keeps the flag for TextField & TextEditor, TextField is now fully resolved.
WinUI Backend only keeps TextEditor.

I’m not sure if the debug warning is optimal, as it still gets called on GtkBackend, and I currently see no way to suppress it without risking insability (2 skipped signal calls) or checking the actual content.

Sources/SwiftCrossUI/Views/TextEditor.swift
```diff
@@ -56,6 +56,13 @@ public struct TextEditor: ElementaryView {
         let content = self.text
 
         backend.updateTextEditor(widget, environment: environment) { newValue in
+            #if DEBUG
+                if text == newValue {
+                    logger.warning(
+                        "Unnecessary write to text Binding of TextEditor detected, please open an issue on the SwiftCrossUI GitHub repository so we can fix it on the backend. Known issue on GtkBackend."
+                    )
```
stackotter, line 67: We shouldn't be logging this for any backends where this is a known issue, because that'd just be annoying and we'll already have an issue for it if this PR doesn't end up addressing the backend-specific issues. Ideally we should figure out why GtkBackend is acting so weird before I merge these changes anyway, and that might lead us to a solution

stackotter (changes requested):

Sources/SwiftCrossUI/Views/TextEditor.swift
MiaKoring: I added the warning in DEBUG to the other widgets we noticed it on. 
The current commit now fully eliminates any unnecessary writes on GtkBackend by temporarily blocking the signal handler. 

If we need to do so on more widget we just add the signal name to the static `GObject.blockableSignalNames` property. Once its added the new storeHandler function adds the ID to a dictionary with the signal name as key.

To execute a block you would use 
```swift
widget.withBlockedSignal(named: "signalName") { 
    // Do stuff
}
```

I’m happy with this solution. 

I ommited the g_object_set() idea as it wasn’t compatible with how signals currently are registered.

Sources/Gtk/GObject.swift
```diff
@@ -23,6 +23,12 @@ open class GObject: GObjectRepresentable {
 
     private var signals: [(UInt, Any)] = []
 
+    private static let blockableSignalNames: Set<String> = [
+        "changed"
+    ]
+
+    private var blockableSignalIDs = [String: UInt]()
```
stackotter: Add short documentation comments to both of these for future contributors, because it's not directly a concept from Gtk and might be unclear

stackotter: As a code style thing, I generally use `var blah: [String: UInt] = [:]`

Sources/Gtk/GObject.swift
```diff
@@ -137,6 +144,38 @@ open class GObject: GObjectRepresentable {
             handler: handler
         )
 
-        signals.append((handlerId, box))
+        storeHandler(handlerId, box: box, for: name)
+    }
+
+    @inline(__always)
```
stackotter: I prefer to avoid these annotations unless profiling says they're necessary. The compiler will generally make pretty good decisions, and this code is complicated enough (internally within the stdlib for append, contains, and dictionary insertion), that the function call overhead should be minimal compared to the body of the function

Sources/Gtk/GObject.swift
```diff
@@ -137,6 +144,38 @@ open class GObject: GObjectRepresentable {
...
+    private func storeHandler(_ id: UInt, box: Any, for signalName: String) {
+        signals.append((id, box))
+        if Self.blockableSignalNames.contains(signalName) {
+            blockableSignalIDs[signalName] = id
+        }
+    }
+
+    /// Executes a closure while temporarily suppressing a specific signal handler.
```
stackotter, line 167: Mention that this only works for signals that are included in the `blockableSignalNames` array. It may also be best to make that array public so that people can read its contents from their own code if necessary.

Sources/Gtk/GObject.swift
```diff
@@ -137,6 +144,38 @@ open class GObject: GObjectRepresentable {
...
+    /// - Note: If no signal ID is stored for the given name, the block executes normally without suppression.
+    public func withBlockedSignal(
+        named signalName: String,
+        block: @escaping () -> Void
+    ) {
+        guard let signalID = blockableSignalIDs[signalName] else {
+            print(
+                """
+                Notice: Skipping signal block for '\(signalName)' (ID missing). \
+                Proceeding with block execution anyway.
```
stackotter: `Warning: Could not block signal '\(signalName)' because it is not included in GObject.blockableSignalNames`

Sources/Gtk/GObject.swift
```diff
@@ -137,6 +144,38 @@ open class GObject: GObjectRepresentable {
...
+    /// - Parameters:
+    ///   - named: The name of the GObject signal to block (e.g., "changed").
+    ///   - block: The closure to execute while the signal is suppressed.
+    /// - Note: If no signal ID is stored for the given name, the block executes normally without suppression.
+    public func withBlockedSignal(
+        named signalName: String,
+        block: @escaping () -> Void
+    ) {
+        guard let signalID = blockableSignalIDs[signalName] else {
+            print(
```
stackotter: If the signal name is in `blockableSignalNames` but not in `blockableSignalIDs` then that just means that there's no registered handler, and nothing for us to do, so we could probably avoid printing the warning in that case?

Sources/SwiftCrossUI/Views/ToggleButton.swift
```diff
@@ -24,7 +24,19 @@ struct ToggleButton: ElementaryView, View {
...
         //   can delay updating the underlying widget until `commit`.
         backend.updateToggle(widget, label: label, environment: environment) { newActiveState in
-            active.wrappedValue = newActiveState
+            if active.wrappedValue != newActiveState {
+                active.wrappedValue = newActiveState
+            } else {
+                #if DEBUG
+                    logger.warning(
+                        """
+                        Unnecessary write to text Binding of TextEditor detected, \
```
stackotter: `to wrappedValue binding of ToggleButton`

Sources/SwiftCrossUI/Views/Slider.swift
```diff
@@ -88,8 +88,20 @@ public struct Slider: ElementaryView, View {
...
-            if let value {
+            if let value,
+                value.wrappedValue != newValue
+            {
                 value.wrappedValue = newValue
+            } else {
+                #if DEBUG
+                    logger.warning(
+                        """
+                        Unnecessary write to text Binding of Slider detected, \
```
stackotter: `to wrappedValue Binding of Slider`

MiaKoring:
Oh, yeah, I somehow forgot about ToggleSwitch and Checkbox. I resolved it on SCUI Level. Also I added the signal blockers to all of them (and the Slider) so the warning doesn’t get logged anymore

Sources/Gtk/GObject.swift
```diff
@@ -156,6 +165,7 @@ open class GObject: GObjectRepresentable {
     }
 
     /// Executes a closure while temporarily suppressing a specific signal handler.
+    /// You can only block signals included in ``GObject/blockableSignalNames``.
```
stackotter, line 168: Add a blank line before this (to separate the overview line and the description paragraph.

Sources/Gtk/GObject.swift
```diff
@@ -166,11 +176,12 @@ open class GObject: GObjectRepresentable {
         block: @escaping () -> Void
     ) {
         guard let signalID = blockableSignalIDs[signalName] else {
-            print(
-                """
-                Notice: Skipping signal block for '\(signalName)' (ID missing). \
-                Proceeding with block execution anyway.
-                """)
+            if !Self.blockableSignalNames.contains(signalName) {
+                print(
```
stackotter, line 180: Wrap this message

Sources/SwiftCrossUI/Views/Checkbox.swift
```diff
@@ -30,7 +30,19 @@ struct Checkbox: ElementaryView, View {
...
-            active.wrappedValue = newActiveState
+            if active.wrappedValue != newActiveState {
+                active.wrappedValue = newActiveState
+            } else {
+                #if DEBUG
+                    logger.warning(
+                        """
+                        Unnecessary write to wrappedValue binding of Checkbox detected, \
+                        please open an issue on the SwiftCrossUI GitHub repository \
+                        so we can fix it on \(type(of: backend)).
```
stackotter, line 41: `fix it for`

Sources/SwiftCrossUI/Views/Slider.swift
```diff
@@ -96,7 +96,7 @@ public struct Slider: ElementaryView, View {
                 #if DEBUG
                     logger.warning(
                         """
-                        Unnecessary write to text Binding of Slider detected, \
+                        Unnecessary write to wrappedValue Binding of Slider, \
```
stackotter, line 99: `detected`

stackotter, line 99: Test whether this is triggered when changing a slider back to its original value

stackotter (changes requested):
Just a few small things this time. I'm going to merge this into a branch and address them myself

Sources/SwiftCrossUI/Views/Slider.swift
stackotter: This was indeed the case. I've removed the warnings from all the views where this check is trivial (i.e. all but TextField and TextEditor)

Issue #179, opened 2025-06-07
Gtk3Backend: Implement `Picker` support again


stackotter:
The `Picker` implementation in Gtk3Backend is currently commented out and outdated. It should be re-enabled and fixed to ensure feature parity across platforms.

PR #447, opened 2026-02-15
Picker styles

### LINKED ISSUE #162: Update UIKitBackend Picker implementation to match .menu style
Currently UIKitBackend's Picker implementation produces pickers that look like so:

Image

This style of picker doesn't act anything like the pickers rendered by other backends and makes it tricky to build layouts that work on both desktop platforms and iOS. For this reason, I believe that we should match our picker style to SwiftUI's default `.menu` picker style:

Image

See [Sarunw's picker style article](https://sarunw.com/posts/swiftui-form-picker-styles/) to see what the other available picker styles look like.

The current UIKitBackend Picker implementation uses [UIPickerView](https://developer.apple.com/documentation/uikit/uipickerview) which is explicitly only for wheel-style pickers. We'll probably have to try to replicate SwiftUI's `.menu` picker style with something similar to our Menu button implementation.

### LINKED ISSUE #261: Support pickerStyle modifier
Currently each backend only support a single picker style. We should support SwiftUI's `pickerStyle` modifier so that we can have more than just each platform's default picker style. Ideally each picker style should be supported on as many platforms as practical.

bbrk24:
Fixes #162
Fixes #261

With the way `NavigationLink` is currently implemented, I was unable to figure out how to implement the `navigationLink` picker style -- in particular, I have no idea how I'm supposed to get a binding to the current navigation path from an arbitrary view.

Support table:
| | `menu` | `radioGroup` | `segmented` | `wheel` |
|-|-|-|-|-|
| AppKitBackend | :ballot_box_with_check: | :ballot_box_with_check: | :ballot_box_with_check: | |
| GtkBackend | :ballot_box_with_check: | | | |
| Gtk3Backend | | | | |
| UIKitBackend | iOS 14+<br />tvOS 17+<br />visionOS | | :ballot_box_with_check: | iOS<br />visionOS |
| WinUIBackend | :ballot_box_with_check: | :ballot_box_with_check: | | |

I attempted to add radioGroup support to GTK, but ran into some issues. It seems like only GTK3 has the `RadioButton` class, not GTK4. I got something that *almost* worked in GTK3, which would allow me to close issue #179 as well, but I ran into an issue: GTK3 seemingly selects the first option by default, so if SCUI thought that nothing was selected, there's no way to select the first option without selecting another option first.

<details>
<summary>Screenshots</summary>

missing tvOS 17, visionOS 1-2, visionOS 26, macOS 26

menu:
macOS 15  
GTK 4  
iOS 26  
iOS 18  
iOS 14-17  
tvOS 26  
tvOS 18  
WinUI  

radioGroup:
macOS 15 
WinUI 

segmented:
macOS 15 
iOS 26 
iOS 13-18 
tvOS 26 (focused)  (unfocused) 
tvOS 18 (focused)  (unfocused) 

wheel:
iOS 26 
iOS 13-18 
</details>


Sources/SwiftCrossUI/Views/Picker.swift
```diff
@@ -59,10 +126,153 @@ public struct Picker<Value: Equatable>: ElementaryView, View {
...
+}
+
+extension PickerStyle where Self == WheelPickerStyle {
+    public static nonisolated var wheel: Self { .init() }
+}
+
+extension PickerStyle where Self == SegmentedPickerStyle {
+    public static nonisolated var segmented: Self { .init() }
+    public static nonisolated var palette: Self { .init() }
+}
```
bbrk24: I would love to include screenshots in the documentation for these, but I don't actually know how to add images to doc comments.

stackotter: We don’t have a good solution for images in the docs yet, because including the images in the repo would blow things up pretty quickly. For now it’d be super handy if you could just include the images in the PR description, and then we can transfer them to the docs once we have a way to do so. It’d also be really handy for when I make the next monthly update video, because I spent quite a while collecting screenshots and screen recordings for the last one.

bbrk24: Got it, I'll get screenshots and attach them to the PR description later tonight! And I should probably write up doc comments anyways.

stackotter:
Thanks for the incredibly detailed set of screenshots! They will be very useful

Sources/SwiftCrossUI/Backend/AppBackend.swift
stackotter, line 611: I feel like we should consider passing the picker style to `updatePicker` so that we don't require backends to store it themselves when using the same widget type for multiple different styles. To make it clear that the picker style passed in is meant to match the style that the picker was created with we could change the first argument to `_ picker: (Widget, BackendPickerStyle)`. What do you reckon?

Sources/SwiftCrossUI/Backend/AppBackend.swift
```diff
@@ -105,6 +105,11 @@ public protocol AppBackend: Sendable {
     /// are supported at all.
     nonisolated var supportedDatePickerStyles: [DatePickerStyle] { get }
 
+    /// The supported picker styles.
+    nonisolated static var supportedPickerStyles: [BackendPickerStyle] { get }
+    /// The picker style used by `automatic`.
+    nonisolated static var defaultPickerStyle: BackendPickerStyle { get }
```
stackotter: Can we make these non-static to match `supportedDatePickerStyles`? Or alternatively, could we make `supportedDatePickerStyles` static? I feel like we should be consistent with the staticness

Sources/SwiftCrossUI/Environment/EnvironmentValues.swift
```diff
@@ -274,6 +274,8 @@ public struct EnvironmentValues {
     /// The display styles supported by ``DatePicker``. ``datePickerStyle`` must be one of these.
     public let supportedDatePickerStyles: [DatePickerStyle]
 
+    public var pickerStyle: any PickerStyle
```
stackotter, line 179: Add documentation comment

Sources/SwiftCrossUI/Views/Modifiers/PickerStyleModifier.swift
```diff
@@ -0,0 +1,11 @@
+extension View {
+    public func pickerStyle<S: PickerStyle>(_ style: S) -> some View {
+        EnvironmentModifier(self) { environment in
+            if !style.isSupported(backend: environment.backend) {
+                assertionFailure("Unsupported picker style: \(style)")
```
stackotter: `"Picker style \(style) not supported by backend \(type(of: environment.backend))"`

Sources/SwiftCrossUI/Views/Picker.swift
```diff
@@ -1,53 +1,132 @@
+public enum BackendPickerStyle: Hashable, Sendable, BitwiseCopyable {
```
stackotter: Document this and move it to a separate file.

Sources/SwiftCrossUI/Views/Picker.swift
```diff
@@ -1,53 +1,132 @@
...
+                selection: value,
+                environment: environment
+            )
+        )
+    }
+}
+
+/// A type that specifies the appearance and interaction of all pickers within a view hierarchy.
+@MainActor
+public protocol PickerStyle {
```
stackotter: Move this to its own file

Sources/SwiftCrossUI/Views/Picker.swift
```diff
@@ -1,53 +1,132 @@
...
         environment: EnvironmentValues,
         backend: Backend
     ) -> ViewLayoutResult {
+        var pickerWidget: Backend.Widget
+
+        if let picker = children.picker, children.style == self.style {
+            pickerWidget = picker.widget as! Backend.Widget
+        } else {
+            let containerWidget = children.container.widget as! Backend.Widget
+            backend.removeAllChildren(of: containerWidget)
```
stackotter: I prefer to use `remove(childAt: 0, in: containerWidget)` when we know that the container has exactly one child.

Sources/SwiftCrossUI/Views/Picker.swift
```diff
@@ -59,10 +138,163 @@ public struct Picker<Value: Equatable>: ElementaryView, View {
...
+    private func asBackendPickerStyle<Backend: AppBackend>(backend: Backend) -> BackendPickerStyle {
+        _asBackendPickerStyle(Backend.self)
+    }
+
+    public func makeView<Value: Equatable>(
+        options: [Value],
+        selection: Binding<Value?>,
+        environment: EnvironmentValues
+    ) -> _PickerImplementation {
+        .init(
```
stackotter: Spell out the init

Sources/SwiftCrossUI/Views/Picker.swift
```diff
@@ -1,53 +1,132 @@
...
+        isSupported(Backend.self)
+    }
+
+    // Default implementation for custom picker styles
+    public func isSupported<Backend: AppBackend>(_: Backend.Type) -> Bool {
+        true
+    }
+}
+
+public struct _PickerImplementation: TypeSafeView {
```
stackotter: Rename to `_BackendPickerImplementation` or `_BuiltinPickerImplementation`

Sources/SwiftCrossUI/Views/Picker.swift
```diff
@@ -59,10 +138,163 @@ public struct Picker<Value: Equatable>: ElementaryView, View {
...
         layout: ViewLayoutResult,
         environment: EnvironmentValues,
         backend: Backend
     ) {
         backend.setSize(of: widget, to: layout.size.vector)
+        backend.setSize(of: children.picker!.widget as! Backend.Widget, to: layout.size.vector)
+    }
+}
+
+final class PickerChildren: ViewGraphNodeChildren {
```
stackotter: Rename to match `_PickerImplementation`'s naming (whatever you decide to rename it to).

Sources/SwiftCrossUI/Views/Picker.swift
```diff
@@ -59,10 +138,163 @@ public struct Picker<Value: Equatable>: ElementaryView, View {
...
+            }
+        )
+    }
+
+    public func isSupported<Backend: AppBackend>(_ backendType: Backend.Type) -> Bool {
+        backendType.supportedPickerStyles.contains(_asBackendPickerStyle(backendType))
+    }
+}
+
+public struct MenuPickerStyle: PickerStyle, _BuiltinPickerStyle {
```
stackotter: Add a struct-level doc comment to each of these style types

Sources/SwiftCrossUI/Views/Picker.swift
```diff
@@ -59,10 +138,163 @@ public struct Picker<Value: Equatable>: ElementaryView, View {
...
+            .wheel
+        } else {
+            .segmented
+        }
+    }
+}
+
+extension PickerStyle where Self == RadioGroupPickerStyle {
+    /// A picker style that presents the options as a group of radio buttons.
+    public static nonisolated var radioGroup: Self { .init() }
```
stackotter: Spell out each of these inits as `Self()`

Sources/UIKitBackend/UIKitBackend+Menu.swift
```diff
@@ -85,7 +85,7 @@ extension UIKitBackend {
             setButtonTitle(buttonWidget, label, environment: environment)
             buttonWidget.child.menu = menu.uiMenu
             buttonWidget.child.showsMenuAsPrimaryAction = true
-            if #available(iOS 16, tvOS 17, macCatalyst 16, *) {
+            if #available(iOS 16, macCatalyst 16, *) {
```
stackotter, line 88: tvOS 17 does seem to support `preferredMenuElementOrder`; is there a reason for removing it from the acceptable platforms here?

Sources/WinUIBackend/WinUIBackend.swift
```diff
@@ -915,26 +916,33 @@ public final class WinUIBackend: AppBackend {
...
+                    brush.color = picker.actualForegroundColor
+                    picker.foreground = brush
+                }
 
-        return picker
+                return picker
+            case .radioGroup:
+                return RadioButtons()
+            default:
+                fatalError("unsupported picker style")
```
stackotter: Put the unsupported style in the message and log with logger.critical in addition to fatalError (I still need to standardise that last part better throughout the whole codebase).

Sources/WinUIBackend/WinUIBackend.swift
```diff
@@ -943,58 +951,80 @@ public final class WinUIBackend: AppBackend {
...
+                }
+            } else {
+                for option in options[picker.items.count...] {
+                    picker.items.append(option)
+                }
             }
-            for i in options.count..<picker.items.count {
-                picker.items.removeAt(UInt32(i))
+
+            picker.selectionChanged.addHandler { [weak picker] _, _ in
```
stackotter: Set this up in `createPicker` instead so that we don't end up with an ever-increasing set of selection handlers on each picker. Copy what I did with the original custom picker class to attach the handler implementation to the widget as a stored property.

Sources/GtkBackend/GtkBackend.swift
```diff
@@ -1024,7 +1025,11 @@ public final class GtkBackend: AppBackend {
         return textEditor.buffer.text
     }
 
-    public func createPicker() -> Widget {
+    public func createPicker(style: BackendPickerStyle) -> Widget {
+        if style != .menu {
+            assertionFailure("unsupported picker style")
```
stackotter: Do a fatalError here for consistency and include the picker style in the message. Also `logger.critical` the same message

Sources/AppKitBackend/AppKitBackend.swift
```diff
@@ -789,8 +792,13 @@ public final class AppKitBackend: AppBackend {
...
     }
 
-    public func createPicker() -> Widget {
-        return NSPopUpButton()
+    public func createPicker(style: BackendPickerStyle) -> Widget {
+        switch style {
+            case .menu: NSPopUpButton()
+            case .segmented: NSSegmentedControl()
+            case .radioGroup: RadioGroup()
+            default: preconditionFailure("unsupported picker type")
```
stackotter: Use fatalError for consistency and include the unsupported picker style in the message. Also `logger.critical` the error

Examples/Sources/ControlsExample/ControlsApp.swift
```diff
@@ -77,37 +109,55 @@ struct ControlsApp: App {
...
                                 HStack {
                                     Text("Flavor: ")
+
                                     Picker(
                                         of: ["Vanilla", "Chocolate", "Strawberry"],
                                         selection: $flavor
                                     )
+                                    .existentialPickerStyle(
+                                        pickerStyle?.asPickerStyle ?? AutomaticPickerStyle()
+                                    )
```
stackotter, line 147: Is this required to do the same thing in regular SwiftUI?

stackotter (changes requested):
Thanks for implementing all this!

Sources/SwiftCrossUI/Backend/AppBackend.swift
bbrk24: Yeah the date picker one can absolutely be static. I wanted this to be static so that the caller of the `isSupported` method didn't need to find the backend instance.

Sources/SwiftCrossUI/Backend/AppBackend.swift
bbrk24: I can do that, but I'm not sure how much it matters. UIKit literally doesn't care with the current implementation, and for AppKit and WinUI that would just change an `if let` chain to a switch with a force-cast in each branch.

Sources/SwiftCrossUI/Views/Picker.swift
bbrk24: But we don't know in this case, because it could have zero children (if this is the first layout pass).

Sources/UIKitBackend/UIKitBackend+Menu.swift
bbrk24: There's already an `if #available(iOS 14, tvOS 17, ...)` further up. The check needs to be done again for iOS because this is more restrictive than the previous check, but not so for tvOS; when compiling for tvOS, this was actually generating a warning for being a redundant check.

Sources/WinUIBackend/WinUIBackend.swift
bbrk24: Should they be fatalError or preconditionFailure? I don't think I was consistent about that between backends.

Sources/SwiftCrossUI/Backend/AppBackend.swift
stackotter: Yeah after finishing the review I think it’s fine the way it is. I meant to go back and amend this comment

Sources/WinUIBackend/WinUIBackend.swift
bbrk24: Oops, good point

Sources/SwiftCrossUI/Backend/AppBackend.swift
stackotter: Does the isSupported method get called without an instance anywhere outside of your example code? I see you have a wrapper method that takes an instance and just turns it into the type. I kind of prefer the way you did things for DatePicker with being able to query the environment rather than a specific backend (cause it’s guaranteed to be using the correct backend if you query the environment). I think having both as instance properties makes more sense to me

Examples/Sources/ControlsExample/ControlsApp.swift
bbrk24: Yes. The problem is that when `pickerStyle` has the type signature `<S: PickerStyle>(S) -> some View`, the "some" is allowed to depend on what `S` ends up being. So when `S` depends on opening an existential, as here, "some" becomes "any", and `any View` isn't allowed in a view builder. I could probably put the `AnyView` wrapper directly here instead of using an extension method, but diagnostics aren't great inside of resultBuilders, so this made it easier to figure out that `AnyView` was necessary in the first place.

stackotter: I wonder if we can improve this, because conditionally changing the picker style seems like a common enough thing to need to do. I guess we would just take an any PickerStyle instead?

Sources/SwiftCrossUI/Backend/AppBackend.swift
bbrk24: The only reason I added the overload that takes an instance was so that I could pass it `environment.backend` (since there's no way to spell the type that way). The problem with exposing the backend's array again on the environment is that that doesn't include `inline` (or, if I got around to implementing it, `navigationLink`).

Sources/WinUIBackend/WinUIBackend.swift
stackotter: I think fatalError is best, cause it always gets triggered (whereas precondition may not in unchecked builds I think?)

Sources/SwiftCrossUI/Backend/AppBackend.swift
stackotter: Maybe you could instead have an environment action that takes a picker style and returns a boolean (e.g. isPickerStyleSupported)

Sources/SwiftCrossUI/Environment/EnvironmentValues.swift
```diff
@@ -262,20 +172,18 @@ public struct EnvironmentValues {
...
-    public var timeZone: TimeZone
-
-    /// The display style used by ``DatePicker``.
-    public var datePickerStyle: DatePickerStyle
-
     /// The display styles supported by ``DatePicker``. ``datePickerStyle`` must be one of these.
     public let supportedDatePickerStyles: [DatePickerStyle]
 
+    /// The display style used by ``Picker``.
     public var pickerStyle: any PickerStyle
```
stackotter, line 179: Make this an `@Entry` (you may have to merge in the latest changes from main if you haven't already)

Sources/SwiftCrossUI/Views/PickerStyle.swift
```diff
@@ -0,0 +1,273 @@
+/// The picker styles that need special backend support.
```
stackotter, line 1: `Built-in picker styles built on top of native widgets.`

Sources/UIKitBackend/UIKitBackend+Menu.swift
stackotter: Ah yep makes sense. I had to expand the diff to see that context

Sources/WinUIBackend/WinUIBackend.swift
```diff
@@ -939,9 +939,19 @@ public final class WinUIBackend: AppBackend {
...
+                    guard let picker else { return }
+                    picker.onChangeSelection?(
+                        picker.selectedIndex == -1 ? nil : Int(picker.selectedIndex)
+                    )
+                }
+
+                return picker
             default:
-                fatalError("unsupported picker style")
+                logger.critical("unsupported picker style \(style)")
```
stackotter, line 953: Compute message once

Sources/GtkBackend/GtkBackend.swift
```diff
@@ -1027,7 +1027,8 @@ public final class GtkBackend: AppBackend {
 
     public func createPicker(style: BackendPickerStyle) -> Widget {
         if style != .menu {
-            assertionFailure("unsupported picker style")
+            logger.critical("unsupported picker style \(style)")
```
stackotter, line 1030: Compute message once

Sources/AppKitBackend/AppKitBackend.swift
```diff
@@ -794,10 +794,15 @@ public final class AppKitBackend: AppBackend {
...
-            case .radioGroup: RadioGroup()
-            default: preconditionFailure("unsupported picker type")
+            case .menu:
+                return NSPopUpButton()
+            case .segmented:
+                return NSSegmentedControl()
+            case .radioGroup:
+                return RadioGroup()
+            default:
+                logger.critical("unsupported picker style \(style)")
```
stackotter, line 804: Compute message once

Sources/SwiftCrossUI/Views/PickerStyle.swift
```diff
@@ -0,0 +1,273 @@
...
+            }
+        )
+    }
+
+    public func isSupported<Backend: AppBackend>(backend: Backend) -> Bool {
+        backend.supportedPickerStyles.contains(_asBackendPickerStyle(backend: backend))
+    }
+}
+
+/// A picker style that presents the options in a drop-down menu.
```
stackotter, line 171: Document backend availability

stackotter (changes requested):
These are mostly just style and docs things, so I'll merge into a branch and apply this feedback

Issue #427, opened 2026-02-07
[FEAT] Gradients


MiaKoring:
I just realized wo don't have gradients yet, which are really important to get a not-completely-flat-looking UI

I expect gradients to be relatively simple to implement across platforms, at least the simple ones. 

Gtk might actually be the easiest due to CSS support.

The first gradients I would like to support initially are: LinearGradient, RadialGradient and AngularGradient.

Honorary mention of MeshGradient which would be nice to have as well.

PR #450, opened 2026-02-17
[FEAT] LinearGradient, RadialGradient & AngularGradient

MiaKoring:
### Summary
Added support for 3 types of gradients to be rendered as Views:
- LinearGradient
- RadialGradient
- AngularGradient

#### SwiftCrossUI
- added functions for creation and update of each gradient type to the `AppBackend` protocol.
- added an `ElementaryView`struct for each type of gradient.
- added a struct `Angle` intended for use by `AngularGradient` later
   - gets also used by multiple backends for calculations
   - added an initializer to create an `Angle` from a `UnitPoint`
- added a struct `UnitPoint`, meant to represent a normalized point in a view’s coordinate space
   - added static values for various points on a rectangle
   - AppKitBackend views using it may need to flip the coordinates vertically, due to AppKit using a different coordinate origin point than the other major platforms.
   - `UnitPoint`s  are not clamped, so you can use them with arbitrary values to benefit for example from `Angle(origin:destination:)`
- added a struct `Gradient`, used by all implementations of a Gradient, its a gradient type agnostic representation of a color gradient
   - added nested struct `Gradient.Stop` representing a color and its normalized position in the gradient
   - added initializers taking `[Color]` or `[Gradient.Stop]`

#### AppKitBackend, UIKitBackend
- Added implementations of the new `AppBackend` methods
- Added separate Widget classes when needed (AppKit & UIKit)

#### WinUI, GtkBackend
- Added implementation of the new `AppBackend` methods related to `LinearGradient` and `RadialGradient`

#### SwiftCrossUITests
- Added Tests validating the initializers of `Gradient`
   - even distribution of stops when supplied `[Color]`
   - transparent stops when passed an empty array of `Color` (to make future support of addition possible and remove edgecases to check outside of `Gradient`)
   - start and end stop of the same color when passed `[Color]` with one entry
   - preservation of the order the colors are passed

#### Examples
- Added GradientsExample
   - added tab with examples for `LinearGradient`
   - added tab with examples for `RadialGradient`
   - added tab with examples for `AngularGradient`

### Notes
- WinUI 2 and 3 don’t support conic (angular) gradients yet. I currently don’t know what an alternative could be
- This PR only adds support for rendering them as views directly. A future `ShapeStyle` could easily re-use the added structs tho.

### TODO / Status
- [x] Added `RadialGradient` support to WinUIBackend (pending [RadialGradientBrush](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.media.radialgradientbrush?view=windows-app-sdk-1.5) generation)
- [x] “unavailable on WinUIBackend” documentation comment is removed from `RadialGradient`


MiaKoring:
Screenshot 2026-02-17 at 20 29 46
Screenshot 2026-02-17 at 20 29 57
Screenshot 2026-02-18 at 12 38 59

Screenshot 2026-02-17 at 20 30 46
Screenshot 2026-02-17 at 20 30 54
Screenshot 2026-02-18 at 12 39 07

MiaKoring:
Related issue: https://github.com/moreSwift/swift-cross-ui/issues/427

MiaKoring:
The new commit now offers full API parity to SwiftUI on AngularGradient and correctly renders all 6 test gradients. “Correct” meaning looking like the SwiftUI rendered result. 

All 3 supported frameworks recieved this change and the shared screenshots were updated.

MiaKoring:
Outstanding Tasks have been completed, this can now be reviewed and merged

Examples/Sources/GradientsExample/GradientsApp.swift
```diff
@@ -0,0 +1,211 @@
...
+        }
+    }
+}
+
+struct AngularGradientView: View {
+    var colors: [Color] { GradientsApp.colors }
+    var stops: [Gradient.Stop] { GradientsApp.stops }
+
+    var specialStops: [Gradient.Stop] = [
+        .init(color: .red, location: 1 / 12),
```
stackotter: Please spell out the init; I generally avoid `.init` in SCUI's code for consistency

(same for the other instances of `.init` in this file)

Examples/Sources/GradientsExample/GradientsApp.swift
```diff
@@ -0,0 +1,211 @@
...
+    import SwiftBundlerRuntime
+#endif
+
+@HotReloadable
+@main
+struct GradientsApp: App {
+    static let colors: [Color] = [.red, .orange, .yellow, .green, .blue, .purple]
+
+    static let stops: [Gradient.Stop] = [
+        .init(color: .red, location: 0), .init(color: .blue, location: 0.25),
```
stackotter: Put each on a new line

Examples/Sources/GradientsExample/GradientsApp.swift
```diff
@@ -0,0 +1,211 @@
...
+    var body: some Scene {
+        WindowGroup("Gradients Example") {
+            HStack {
+                ForEach(GradientType.allCases, id: \.rawValue) { type in
+                    Button(type.rawValue) {
+                        gradientType = type
+                    }
+                    .disabled(gradientType == type)
+                }
+            }
```
stackotter: Until I merge @bbrk24's picker styles PR it'd be best to use a `NavigationSplitView` here as that's the most idiomatic way to switch between multiple views in SwiftCrossUI at the moment

Examples/Sources/GradientsExample/GradientsApp.swift
```diff
@@ -0,0 +1,211 @@
...
+                        RadialGradientView()
+                    case .angular:
+                        #if canImport(UIKitBackend)
+                            ScrollView(.horizontal) {
+                                AngularGradientView()
+                            }
+                        #else
+                            AngularGradientView()
+                        #endif
+                }
```
stackotter: Is this to accommodate thinner screens? It might be best to just use the scroll view on every platform (because it generally doesn't have too much of an effect unless the contents are actually bigger than the scroll view's proposed width)

Examples/Sources/GradientsExample/GradientsApp.swift
```diff
@@ -0,0 +1,211 @@
...
+                        .init(color: .white, location: 0.2),
+                        .init(color: .white, location: 0.5),
+                    ],
+                    center: .center,
+                    angle: .degrees(230)
+                )
+                .frame(width: 300)
+            }
+            .frame(height: 300)
+
```
stackotter: Delete this blank line

Tests/SwiftCrossUITests/GradientTests.swift
```diff
@@ -0,0 +1,200 @@
...
+        #expect(result[1].location ~= 0.75)
+        #expect(result[2].location ~= 1)
+    }
+}
+
+fileprivate extension Double {
+    static func ~= (lhs: Self, rhs: Self) -> Bool {
+        Int(lhs * 1_000_000_000) ==
+        Int(rhs * 1_000_000_000)
+    }
```
stackotter, line 200: This threshold seems a bit arbitrary. I'd prefer that we instead just compare `abs(lhs - rhs)` to some explicitly-set threshold (`1e-6` is a common one if I remember correctly)

Tests/SwiftCrossUITests/GradientTests.swift
```diff
@@ -0,0 +1,200 @@
+import Testing
+@testable import SwiftCrossUI
+
+@Suite("Test Gradients")
+@MainActor
+struct GradientTests {
+    @Test("Automatic equal distribution of color")
+    func testAutomaticColorDistribution() async throws {
+        let gradient = Gradient(colors: .init(repeating: .red, count: 12))
+        let gradient1 = Gradient(colors: .init(repeating: .green, count: 3))
```
stackotter: Why do we test two different gradients here? Is there something different in behaviour between 12 and 3?

Sources/SwiftCrossUI/Backend/AppBackend.swift
```diff
@@ -877,6 +877,51 @@ public protocol AppBackend: Sendable {
...
+    // MARK: Gradients
+
+    /// Create a linear gradient.
+    func createLinearGradient() -> Widget
+    /// Update a linear gradient.
+    /// - Parameters:
+    ///   - widget: The widget to update.
+    ///   - gradient: The SwiftCrossUI representation of the widget.
+    ///   - withSize: The size dictated by ``ViewLayoutResult``.
+    ///   - in: The environment of the gradient.
```
stackotter: The keys in the parameter list are meant to be the internal parameter names instead of the labels. So this one should be `environment: ...`, and the one above should be `size: ...`. Same for the other documentation comments that you introduced in this file

Tests/SwiftCrossUITests/GradientTests.swift
```diff
@@ -0,0 +1,200 @@
...
+            ],
+            center: .center,
+            startRadius: 0,
+            endRadius: 300
+        )
+        
+        let result = gradient.adjustedStops
+        
+        #expect(result == gradient.gradient.stops)
+    }
```
stackotter, line 174: I would've expected the first stop to go at `startRadius` and the last stop to go at `endRadius`, but it seems that swapping the two radii results in the same adjustedStops. I will probably find this out when testing locally, but does this result in correct behaviour for backends that do rely on `adjustedStops`?

Sources/WinUIBackend/WinUIBackend+Gradient.swift
```diff
@@ -0,0 +1,87 @@
...
+        _ widget: Widget,
+        gradient: LinearGradient,
+        withSize size: SIMD2<Int>,
+        in environment: EnvironmentValues
+    ) {
+        let widget = widget as! WinUI.Rectangle
+
+        let collection = GradientStopCollection()
+
+        gradient.gradient.stops.forEach {
```
stackotter: Use `for stop in gradient.gradient.stops {` instead of `forEach`

Sources/WinUIBackend/WinUIBackend+Gradient.swift
```diff
@@ -0,0 +1,87 @@
...
+        _ widget: Widget,
+        gradient: RadialGradient,
+        withSize size: SIMD2<Int>,
+        in environment: EnvironmentValues
+    ) {
+        let widget = widget as! WinUI.Rectangle
+
+        let brush = RadialGradientBrush()
+
+        gradient.adjustedStops.forEach {
```
stackotter: Use `for stop in gradient.gradient.adjustedStops {` instead of `forEach`

Sources/WinUIBackend/WinUIBackend+Gradient.swift
```diff
@@ -0,0 +1,87 @@
...
+        in environment: EnvironmentValues
+    ) {
+        let widget = widget as! WinUI.Rectangle
+
+        let collection = GradientStopCollection()
+
+        gradient.gradient.stops.forEach {
+            let color = $0.color.resolve(in: environment)
+            let stop = GradientStop()
+            stop.color = .init(
```
stackotter: Spell out this `.init`

Sources/SwiftCrossUI/Values/Angle.swift
```diff
@@ -0,0 +1,50 @@
...
+        Angle(degrees: lhs.degrees + rhs.degrees)
+    }
+
+    public static func - (lhs: Self, rhs: Self) -> Self {
+        Angle(degrees: lhs.degrees - rhs.degrees)
+    }
+}
+
+extension Angle {
+    public static var zero: Self { Angle(degrees: 0) }
```
stackotter: I think that this can just be a `static let`

Sources/SwiftCrossUI/Values/Angle.swift
```diff
@@ -0,0 +1,50 @@
+import Foundation
+
+/// A geometric angle whose value you access in either radians or degrees.
+public struct Angle {
+    public var degrees: Double
```
stackotter: Document all public APIs in this file

Sources/SwiftCrossUI/Values/Angle.swift
```diff
@@ -0,0 +1,50 @@
...
+    public var radians: Double
```
stackotter, line 16: We should either store this in `radians` or `degrees`, not both, because with this current API someone could set radians and degrees to contradictory values (no point in having competing sources of truth). I believe that we should radians because that seems to generally be the standard for computer programs (and the Swift stdlib).

The `degrees` property would then be a computed property that implements get and set via the `radians` stored property.

Sources/SwiftCrossUI/Values/Angle.swift
```diff
@@ -0,0 +1,50 @@
...
+
+    public static let conversionFactor = Double.pi / 180
+
+    public static func + (lhs: Self, rhs: Self) -> Self {
+        Angle(degrees: lhs.degrees + rhs.degrees)
+    }
+
+    public static func - (lhs: Self, rhs: Self) -> Self {
+        Angle(degrees: lhs.degrees - rhs.degrees)
+    }
```
stackotter: Once `radians` is the underlying source of truth property, these should be implemented on top of the `radians` property instead of `degrees` so that we can avoid unnecessary conversion errors creeping in.

Sources/SwiftCrossUI/Values/UnitPoint.swift
```diff
@@ -0,0 +1,46 @@
+/// A normalized 2D point in a view’s coordinate space.
+@frozen
```
stackotter: I don't think `frozen` makes sense for us here, because we're not a dynamic library shipped with ABI stability. It probably does nothing in our case

Sources/SwiftCrossUI/Values/UnitPoint.swift
```diff
@@ -0,0 +1,46 @@
...
+public struct UnitPoint {
+    /// The normalized distance from the origin to the point in the horizontal direction.
+    public var x: Double
+    /// The normalized distance from the origin to the point in the vertical dimension.
+    public var y: Double
+
+    /// Creates a unit point with the specified horizontal and vertical offsets.
```
stackotter, line 14: If users are allowed to construct these with arbitrary values, then we should have a precondition that ensures that both `x` and `y` lie between `0` and `1` inclusive (with the precondition documented using the following DocC syntax: `/// - Precondition: ...`)

Sources/SwiftCrossUI/Values/UnitPoint.swift
```diff
@@ -0,0 +1,46 @@
+/// A normalized 2D point in a view’s coordinate space.
```
stackotter: This was quite confusing to me originally. I assumed that it was meant to be a unit vector, but it's actually just an arbitrary point within the unit square in the positive quadrant of the cartesian plane (i.e. any point with `0 <= x <= 1` and `0 <= y <= 1`).

Expand this documentation to describe that it is a point within the positive unit square, and mention the range constraints that I explained above.

Sources/SwiftCrossUI/Values/UnitPoint.swift
```diff
@@ -0,0 +1,46 @@
...
+@frozen
+public struct UnitPoint {
+    /// The normalized distance from the origin to the point in the horizontal direction.
```
stackotter, line 9: I don't know why the Apple docs say it this way, I find it a very confusing usage of `normalized`. I would prefer the following description;

```
/// The point's horizontal coordinate within the unit square.
```

Sources/SwiftCrossUI/Values/UnitPoint.swift
```diff
@@ -0,0 +1,46 @@
...
+    public static let top = UnitPoint(x: 0.5, y: 0)
+    /// A point that’s in the top, trailing corner of a view.
+    public static let topTrailing = UnitPoint(x: 1, y: 0)
+
+    /// A point that’s centered vertically on the leading edge of a view.
+    public static let leading = UnitPoint(x: 0, y: 0.5)
+    /// A point that’s centered in a view.
+    public static let trailing = UnitPoint(x: 1, y: 0.5)
+    /// A point that’s centered vertically on the trailing edge of a view.
+    public static let center = UnitPoint(x: 0.5, y: 0.5)
```
stackotter: I believe that you've swapped the documentation for these two members

Sources/SwiftCrossUI/Values/UnitPoint.swift
```diff
@@ -0,0 +1,46 @@
...
+        self.x = 0
+        self.y = 0
+    }
+}
+
+extension UnitPoint {
+    /// The origin of a view.
+    public static let zero = UnitPoint()
+
+    /// A point that’s in the top, leading corner of a view.
```
stackotter: All of this documentation has smart quotes (copied from the Apple docs). Please replace them with `'` (the normal single quote) for consistency with out other documentation comments. DocC will likely convert all of them to smart quotes automatically anyway, but consistency would be best in case other tools (such as LSPs) don't.

Sources/SwiftCrossUI/Values/UnitPoint.swift
```diff
@@ -0,0 +1,46 @@
...
+
+    /// Creates a unit point at the origin.
+    public init() {
+        self.x = 0
+        self.y = 0
+    }
+}
+
+extension UnitPoint {
+    /// The origin of a view.
```
stackotter: Add that this is the `top, leading corner of the view`

Sources/SwiftCrossUI/Values/Gradient.swift
```diff
@@ -0,0 +1,88 @@
+/// A color gradient represented as an array of color stops, each having a parametric location value.
```
stackotter: I know that this is copied from Apple, but I'm not quite sure what `parametric` is meant to imply here. Maybe that `location` is from `0` to `1` inclusive? Might be best to say `each having a location value between `0` and `1` inclusive.

Sources/SwiftCrossUI/Values/Gradient.swift
```diff
@@ -0,0 +1,88 @@
...
+public struct Gradient {
+    /// The array of color stops.
+    public var stops: [Gradient.Stop]
+
+    /// Creates a gradient from an array of color stops.
+    init(stops: [Gradient.Stop]) {
+        guard
+            let first = stops.first
+        else {
```
stackotter: Put the `guard let ... else {` all on one line cause it fits

Sources/SwiftCrossUI/Values/Gradient.swift
```diff
@@ -0,0 +1,88 @@
...
+            self.stops = [
+                Stop(color: invisible, location: 0),
+                Stop(color: invisible, location: 1),
+            ]
+            return
+        }
+
+        #if DEBUG
+            if stops != stops.sorted(by: { $0.location < $1.location }) {
+                logger.warning("Gradient stop locations must be ordered")
```
stackotter, line 22: Document this requirement in the initialiser's documentation comment, because it's not necessarily clear (in some alternate universe we could sort them on the user's behalf and the API would still make sense)

Sources/SwiftCrossUI/Values/Gradient.swift
```diff
@@ -0,0 +1,88 @@
+/// A color gradient represented as an array of color stops, each having a parametric location value.
+public struct Gradient {
+    /// The array of color stops.
```
stackotter: `/// The array of color stops, ordered by location.

Sources/SwiftCrossUI/Values/Gradient.swift
```diff
@@ -0,0 +1,88 @@
...
+        #endif
+
+        if stops.count == 1 {
+            self.stops = [
+                Stop(color: first.color, location: 0),
+                Stop(color: first.color, location: 1),
+            ]
+            return
+        }
+        self.stops = stops
```
stackotter: Instead of an early return I often prefer to use an `if`/`else`. In this case, put the `self.stops = stops` in the else and remove the early return

Sources/SwiftCrossUI/Values/Gradient.swift
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    /// Creates a gradient from an array of colors.
+    ///
+    /// The gradient synthesizes its location values to evenly space the colors along the gradient.
+    init(colors: [Color]) {
+        guard
+            let first = colors.first
+        else {
```
stackotter: Put `guard let ... else {` on one line because it fits

Sources/SwiftCrossUI/Values/Gradient.swift
```diff
@@ -0,0 +1,88 @@
...
+            let invisible = Color.black.opacity(0)
+            self.stops = [
+                Stop(color: invisible, location: 0),
+                Stop(color: invisible, location: 1),
```
stackotter, line 45: Document that the gradient will be fully transparent if no colors/stops are given (for both inits)

Sources/SwiftCrossUI/Values/Gradient.swift
```diff
@@ -0,0 +1,88 @@
...
+            return
+        }
+
+        let locationDifference = 1.0 / Double(colors.count - 1)
+
+        var stops = [Stop(color: first, location: 0)]
+        var currentLocation = 0.0
+
+        for color in colors[1...] {
+            currentLocation += locationDifference
```
stackotter: To avoid floating point error adding up and giving us an end location not quite equal to `1`, we should instead do this;

```swift
for (i, color) in colors[1...].enumerated() {
    let location = Double(i) / Double(colors.count - 1)
    stops.append(Stop(..., location: location))
}
```

I believe that should work as is because `enumerated` will use the original element indices, but if it returns `0` as the index of the first color in that loop, then you'll have to add 1 to `i`.

Sources/SwiftCrossUI/Values/Gradient.swift
```diff
@@ -0,0 +1,88 @@
...
+            stops.append(
+                Stop(color: color, location: currentLocation)
+            )
+        }
+
+        self.stops = stops
+    }
+
+    /// One color stop in the gradient.
```
stackotter: I think that `in a gradient.` would make more sense (because there's not a specific gradient that we're talking about here)

Sources/SwiftCrossUI/Values/Gradient.swift
```diff
@@ -0,0 +1,88 @@
...
+    public struct Stop {
+        /// Creates a color stop with a color and location.
```
stackotter, line 75: Document that `location` must be between `0` and `1` inclusive.

Sources/SwiftCrossUI/Values/Gradient.swift
```diff
@@ -0,0 +1,88 @@
...
+        public init(color: Color, location: Double) {
+            self.color = color
+            self.location = location
+        }
+
+        /// The color for the stop.
+        public var color: Color
+        /// The parametric location of the stop.
```
stackotter, line 86: Document that `location` must be between `0` and `1` inclusive

Sources/SwiftCrossUI/Values/Gradient.swift
```diff
@@ -0,0 +1,88 @@
...
+        public var location: Double
+    }
+}
+
+extension Gradient.Stop: Equatable {}
```
stackotter: Make `Stop` sendable and hashable. Might be cleaner to just put the conformances on the struct decl itself rather than an extension

Sources/SwiftCrossUI/Values/Gradient.swift
```diff
@@ -0,0 +1,88 @@
+/// A color gradient represented as an array of color stops, each having a parametric location value.
+public struct Gradient {
```
stackotter: Make this sendable and hashable

Sources/SwiftCrossUI/Views/Gradients/AngularGradient.swift
```diff
@@ -0,0 +1,158 @@
...
+///
+/// Currently unsupported on WinUIBackend.
+public struct AngularGradient: ElementaryView {
+    /// The gradient represented as an array of color stops, each having a parametric location value.
+    public let gradient: Gradient
+    /// The normalized center point of the gradient in its coordinate space.
+    public let center: UnitPoint
+    /// The angle at which the gradient starts drawing. 0° is trailing center.
+    public let startAngle: Angle
+    /// The angle at which the gradient stops drawing. Everything after is filled with the last used color.
```
stackotter, line 11: Document what happens when this is nil

Sources/SwiftCrossUI/Views/Gradients/AngularGradient.swift
```diff
@@ -0,0 +1,158 @@
...
+    public let endAngle: Angle?
+
+    private static let idealSize = ViewSize(10, 10)
+
+    /// Creates a conic gradient that completes a full turn.
```
stackotter: SwiftUI's documentation isn't consistent, but we should consistently refer to this type of gradient as an angular gradient, with a note in the type-level docs that points out that it's often also referred to as a conic gradient.

Sources/SwiftCrossUI/Views/Gradients/AngularGradient.swift
```diff
@@ -0,0 +1,158 @@
...
+    ) {
+        self.gradient = gradient
+        self.center = center
+        self.startAngle = angle
+        self.endAngle = nil
+    }
+
+    /// Creates a conic gradient that completes a partial rotation.
+    ///
+    /// Stops are expected to be in 360° unit space.
```
stackotter: I assume that this means that they should be between 0 and 360 degrees? Might be best to spell it out because that wording doesn't seem super clear to me. We should also add a `precondition(...)` guarding against values so that people don't get confused when their angular gradient looks different on every platform if each platform handles the edge case differently (perhaps some backends will handle non-normalised angles just fine while others act strangely).

Sources/SwiftCrossUI/Views/Gradients/RadialGradient.swift
```diff
@@ -0,0 +1,121 @@
+/// A radial gradient.
+public struct RadialGradient: ElementaryView {
+    /// The gradient represented as an array of color stops, each having a parametric location value.
+    public let gradient: Gradient
+    /// The radius the gradient starts laying out colors.
+    /// The circle smaller than this radius gets filled with the first color.
+    public let startRadius: Double
+    /// The radius the gradient stops laying out colors.
+    /// The circle bigger than this radius gets filled with the last color.
+    public let endRadius: Double
```
stackotter, line 12:
```
/// The radius at which the last gradient stop will be placed.
///
/// All space outside this radius gets filled with the color of the last gradient stop.
```

Do something similar to the `startRadius` documentation.

Sources/SwiftCrossUI/Views/Gradients/RadialGradient.swift
```diff
@@ -0,0 +1,121 @@
...
+            startRadius: startRadius,
+            endRadius: endRadius
+        )
+    }
+
+    /// Stops adjusted to accomodate startRadius on backends without native support.
+    public var adjustedStops: [Gradient.Stop] {
+        let range = endRadius - startRadius
+
+        guard range != endRadius else { return gradient.stops }
```
stackotter: `guard startRadius != 0` would be more direct (generally better for this sort of code). Can then be moved to the top of the computed property implementation as well (before `range`).

Sources/SwiftCrossUI/Views/Gradients/RadialGradient.swift
```diff
@@ -0,0 +1,121 @@
...
+
+        #if DEBUG
+            if range < 0 {
+                logger.warning(
```
stackotter: SwiftUI supports inverted radius ranges so we should as well. To do so you can reverse the order of the gradient stops and adjust their locations accordingly.

For future reference, if we did have to have this check, I'd prefer that it ran in both debug and release mode, because it's a cheap check, and otherwise the behaviour of the gradient would visually change between debug and release mode.

Sources/SwiftCrossUI/Views/Gradients/RadialGradient.swift
```diff
@@ -0,0 +1,121 @@
...
+                    """
+                )
+                return gradient.stops
+            }
+        #endif
+
+        let dividableRange = range / endRadius
+        let innerCircle = (endRadius - range) / endRadius
+
+        return gradient.stops.map {
```
stackotter: Give the closure parameter a name so that we don't have to use `$0`; `.map { stop in`

I generally try to do so when closures aren't completely trivial closures that fit all on one line

Sources/AppKitBackend/AppKitBackend+Gradient.swift
```diff
@@ -0,0 +1,209 @@
...
+        with environment: EnvironmentValues,
+        frame: SIMD2<Int>
+    ) -> Self {
+        let layer = Self()
+        layer.type = .conic
+
+        let adjustedStops = gradient.adjustedStops
+
+        layer.locations = adjustedStops.map {
+            NSNumber(floatLiteral: $0.location)
```
stackotter: The attached warning says to do `NSNumber(value: $0.location)` instead I think

Sources/AppKitBackend/AppKitBackend+Gradient.swift
```diff
@@ -0,0 +1,209 @@
...
+        for gradient: AngularGradient,
+        with environment: EnvironmentValues,
+        frame: SIMD2<Int>
+    ) -> Self {
+        let layer = Self()
+        layer.type = .conic
+
+        let adjustedStops = gradient.adjustedStops
+
+        layer.locations = adjustedStops.map {
```
stackotter: Give the closure parameter a name (I explained in earlier comment, but I generally just prefer to do so unless the closure is completely trivial and fits on one line such as `{ $0.id == targetId }` or similar). Same for other closures below

Sources/AppKitBackend/AppKitBackend+Gradient.swift
```diff
@@ -0,0 +1,209 @@
...
+
+        layer.startPoint = gradient.center.cgPoint
+        layer.endPoint = (Angle(degrees: 360) - gradient.startAngle).cgPoint
+
+        return layer
+    }
+}
+
+extension Angle {
+    var cgPoint: CGPoint {
```
stackotter: This is quite a non-standard conversion from angle to vector so it should be given a different more descriptive name and some documentation explaining why.

Sources/AppKitBackend/AppKitBackend+Gradient.swift
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    override func viewDidMoveToWindow() {
+        self.wantsLayer = true
+        self.layer?.drawsAsynchronously = true
+    }
+}
+
+class GradientView: NSView {
+    override var isFlipped: Bool { true }
```
stackotter, line 152: Add blank line after this property

Sources/AppKitBackend/AppKitBackend+Gradient.swift
```diff
@@ -0,0 +1,209 @@
...
+            let environment = lastEnvironment
+        else { return }
+
+        let colors = gradient.gradient.stops.map {
+            $0.color.resolve(in: environment).nsColor
+        }
+
+        NSBezierPath(rect: bounds).addClip()
+
+        let nsGradient = NSGradient(
```
stackotter: Apple doesn't document when this returns `nil` vs not, so we should make this a `guard let` and print an error such as `Failed to construct NSGradient; init returned nil`. The other force unwraps we use around the backends are generally for things that we're pretty sure won't fail unless we break our own assumptions. In this case we don't really know what the preconditions are so there may be ones that we're not already designing for.

Sources/AppKitBackend/AppKitBackend+Gradient.swift
```diff
@@ -0,0 +1,209 @@
...
```
stackotter: Make this a `guard let` (as suggested in an earlier comment)

Sources/AppKitBackend/AppKitBackend+Gradient.swift
```diff
@@ -0,0 +1,209 @@
...
+        layer.locations = adjustedStops.map {
+            NSNumber(floatLiteral: $0.location)
+        }
+
+        layer.colors = adjustedStops.map {
+            $0.color.resolve(in: environment).cgColor
+        }
+
+        layer.startPoint = gradient.center.cgPoint
+        layer.endPoint = (Angle(degrees: 360) - gradient.startAngle).cgPoint
```
stackotter: `endAngle` is not used in this implementation afaict. Does AppKitBackend support `endAngle` properly?

Sources/AppKitBackend/AppKitBackend+Gradient.swift
```diff
@@ -0,0 +1,209 @@
...
+
+    public func updateAngularGradient(
+        _ widget: NSView,
+        gradient: AngularGradient,
+        withSize size: SIMD2<Int>,
+        in environment: EnvironmentValues
+    ) {
+        let widget = widget as! GradientView
+        widget.setGradientLayer(
+            to: CAGradientLayer.angularGradientLayer(
```
stackotter, line 47: Is it possible for us to use this gradient layer approach for all three gradient variants? It seems nice and would probably allow us to reduce the amount of code quite a bit

Sources/GtkBackend/GtkBackend+Gradient.swift
```diff
@@ -0,0 +1,119 @@
...
+
+        let angle = Angle(origin: startPoint, destination: endPoint)
+
+        let stops = cssStops(gradient: gradient.gradient, environment: environment)
+
+        widget.css.set(
+            property: .init(
+                key: "background",
+                value: """
+                    linear-gradient(\((angle + Angle(degrees: 90)).radians)rad, \(stops.joined(separator: ", ")))
```
stackotter: Do the angle adjustment and conversion to radians outside of the interpolation to keep things simple inside the interpolation

Sources/GtkBackend/GtkBackend+Gradient.swift
```diff
@@ -0,0 +1,119 @@
...
+        let widget = widget as! Box
+
+        let stops = cssStops(gradient: gradient.gradient, environment: environment)
+
+        widget.css.set(
+            property: .init(
+                key: "background",
+                value: """
+                    radial-gradient(\
+                    circle at \(gradient.center.x * 100)% \(gradient.center.y * 100)%, \
```
stackotter: Do the percentage conversions (and maybe stop joining) outside of the interpolation to keep the interpolation simple

Sources/GtkBackend/GtkBackend+Gradient.swift
```diff
@@ -0,0 +1,119 @@
...
+                rgba(\(red), \(green), \(blue), \
+                \(resolved.opacity)) \($0.location * 360)deg
+                """
+        }
+
+        widget.css.set(
+            property: .init(
+                key: "background",
+                value: """
+                    conic-gradient(from \(gradient.startAngle.degrees + 90)deg \
```
stackotter: Do the angle conversions, percentage conversions, and joining outside of the interpolation to keep the interpolation simple

Sources/GtkBackend/GtkBackend+Gradient.swift
```diff
@@ -0,0 +1,119 @@
...
+                    at \(gradient.center.x * 100)% \(gradient.center.y * 100)%, \
+                    \(stops.joined(separator: ", ")))
+                    """
+            )
+        )
+    }
+
+    private func cssStops(gradient: Gradient, environment: EnvironmentValues) -> [String] {
+        return gradient.stops.map {
```
stackotter: Give the closure parameter a name to make things a bit more readable

Sources/GtkBackend/GtkBackend+Gradient.swift
```diff
@@ -0,0 +1,119 @@
...
+            let resolved = $0.color.resolve(in: environment)
+            let red = resolved.red * 255
+            let green = resolved.green * 255
+            let blue = resolved.blue * 255
+            return
+                """
+                rgba(\(red), \(green), \(blue), \
+                \(resolved.opacity)) \($0.location * 100)%
```
stackotter: If you do the location percentage conversion outside of the interpolation then you might be able to fit the string on a single line.

Sources/AppKitBackend/AppKitBackend+Gradient.swift
```diff
@@ -0,0 +1,209 @@
...
+        let colors = gradient.gradient.stops.map {
+            $0.color.resolve(in: environment).nsColor
+        }
+
+        NSBezierPath(rect: bounds).addClip()
+
+        let nsGradient = NSGradient(
+            colors: colors,
+            atLocations: gradient.gradient.stops.map { CGFloat($0.location) },
+            colorSpace: .extendedSRGB
```
stackotter: We should use `deviceRGB` for consistency with UIKitBackend until we know what the real correct choice is

Sources/UIKitBackend/UIKitBackend+Gradient.swift
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    func setupGradient() {
+        gradientLayer.drawsAsynchronously = true
+    }
+
+    override func didMoveToWindow() {
+        super.didMoveToWindow()
+        // UIView is always layer-backed, no need for wantsLayer
+        setupGradient()
```
stackotter: Just put the implementation of setupGradient in here because it's not used anywhere else

Sources/UIKitBackend/UIKitBackend+Gradient.swift
```diff
@@ -0,0 +1,220 @@
...
+        self.startPoint = gradient.center.cgPoint
+        self.endPoint = (Angle(degrees: 360) - gradient.startAngle).cgPoint
+    }
+}
+
+extension UnitPoint {
+    var cgPoint: CGPoint {
+        CGPoint(x: x, y: y)
+    }
+}
```
stackotter, line 207: Put a blank line after this

stackotter (changes requested):
Thanks for tackling this feature! I've reviewed all of the code now, but I haven't tested anything locally yet. I plan to construct a SwiftUI app with a bunch of edge cases that I have thought of, and then compile that with your PR and ensure that all of the edge cases act the same across platforms. It should be a useful tool to help you address some of my PR comments too. I likely won't get around to doing that today though.

Tests/SwiftCrossUITests/GradientTests.swift
MiaKoring: I wanted to make sure that I didn't miss something with odd count & few colors. I can remove one if you want me to.

Tests/SwiftCrossUITests/GradientTests.swift
MiaKoring: Sorry, I don’t really understand what you are asking

Sources/SwiftCrossUI/Values/UnitPoint.swift
MiaKoring: I honestly don’t think we should. UnitPoint is just normalized to the coordinate system of views. That doesn’t mean it needs to be constrained to be withing that coordinate system. If you have a value < 0 || > 1 its just outside of it.

And adding a precondition makes sense where data would otherwise corrupt. Which is not the case here. Limiting it to be 0…1 would also limit the use of UnitPoint.

"A unit point with one or more components outside the range [0, 1] projects to a point outside of the view.” ~ [Apple Documentation](https://developer.apple.com/documentation/swiftui/unitpoint/)

Sources/SwiftCrossUI/Values/Gradient.swift
MiaKoring: I believe Apple does the purple warning, because it pulls attention to a likely typo, as you would usually spell them out in order.

This in my opinion is good to make debuggin easier. I’m not sure what you meant with the alternate universe in context of what you would like me to do. I believe keeping this warning would be the right choice.

Sources/SwiftCrossUI/Views/Gradients/AngularGradient.swift
MiaKoring: 360° Unit space means 0 is 0° and 1 is 360°. 
backends convert those relative values to what they need. 

I’m not sure what you would like me to check with a precondition. If it looks differently on a backend the backend is wrong and we need to adjust it. Thats not a concern of the SCUI layer imo.

Sources/SwiftCrossUI/Views/Gradients/RadialGradient.swift
MiaKoring: I’m surprised SwiftUI supports it, because I see 0 point in doing so if you want it reversed just change the order you pass the colors in/the order of your stops. There is no directionality to a radial gradient like on an AngularGradient.

Sources/AppKitBackend/AppKitBackend+Gradient.swift
MiaKoring: It does, but only via the use of adjusted stops, which converts the stop locations to have the range between start and end angle as their unit space. the rest is just filled in the last color, like SwiftUI does it.

There is no native support by CAGradientLayer for it.

Sources/AppKitBackend/AppKitBackend+Gradient.swift
MiaKoring: Sadly it isn’t. That was the approach I initially took. But CAGradientLayer stretches, leading to radial gradients become ovals and linear gradients changing their angle with changing width/height, which is not what SwiftUI does.

NSGradient lets us match SwiftUI’s behavior, but id doesn’t support conic gradients. CAGradientLayers stretching is also completely fine, because all lines are drawn from the center out, so effectively the gradient pattern should expand without changing.

Sources/SwiftCrossUI/Values/UnitPoint.swift
MiaKoring: its normalized because its a factor thats independent from the actual bounds of the view.

I find unit square more confusing, because it sounds a bit like it was only for squares, while it really can be anything.

Sources/SwiftCrossUI/Values/Gradient.swift
MiaKoring: I believe normalized would be more fitting, as its not limited to be between 0 and 1, it can be anything really. And 2 is not rendered the same as 1, not everything is visible, but the color interpolation still happens, so if you have white at 0.5 and black at 2 you just get some gray, no black at the end.

Sources/SwiftCrossUI/Values/UnitPoint.swift
MiaKoring: this is not correct, its actually any point, inside and outside the views bounds, but starting centered at the start of the view’s coordinate space and relative to its size

Sources/SwiftCrossUI/Values/Gradient.swift
MiaKoring: this is not the case

Sources/SwiftCrossUI/Values/Gradient.swift
MiaKoring: this is not the case

MiaKoring:
conflicts should be resolved now, same for all requested changes without open discussion

Examples/Sources/GradientsExample/GradientsApp.swift
```diff
@@ -0,0 +1,235 @@
...
+        Gradient.Stop(color: .red, location: 0),
+        Gradient.Stop(color: .blue, location: 0.25),
+        Gradient.Stop(color: .purple, location: 1),
+    ]
+
+    @State var gradientType: GradientType = .linear
+
+    var body: some Scene {
+        WindowGroup("Gradients Example") {
+            #if !canImport(UIKitBackend)
```
stackotter, line 23: Indent the contents of compile-time conditionals

Examples/Sources/GradientsExample/GradientsApp.swift
```diff
@@ -0,0 +1,235 @@
...
+            NavigationSplitView {
+                ForEach(GradientType.allCases, id: \.rawValue) { type in
```
stackotter: I believe that a `List` would look more idiomatic here. The SplitExample should have an example of that

Examples/Sources/GradientsExample/GradientsApp.swift
```diff
@@ -0,0 +1,235 @@
...
+            VStack {
+                HStack {
+                    ForEach(GradientType.allCases, id: \.rawValue) { type in
+                        Button(type.rawValue) {
+                            gradientType = type
+                        }
+                        .disabled(gradientType == type)
+                    }
+                }
+                ScrollView {
```
stackotter: Extract this duplicate gradientType -> scrollView logic into a function

Examples/Sources/GradientsExample/GradientsApp.swift
```diff
@@ -0,0 +1,235 @@
...
+                    switch gradientType {
+                        case .linear:
+                            LinearGradientView()
+                        case .radial:
+                            RadialGradientView()
+                        case .angular:
+                            #if !canImport(WinUIBackend)
+                                ScrollView(.horizontal) {
+                                    AngularGradientView()
+                                }
```
stackotter: Add an `#else` with a message about angular gradients not being supported on Windows

Examples/Package.swift
```diff
@@ -104,10 +104,14 @@ let package = Package(
             dependencies: exampleDependencies
         ),
         .executableTarget(
+            name: "GradientsExample",
+            dependencies: exampleDependencies
+        ),
+.executableTarget(
```
stackotter: This should be indented like the others

Tests/SwiftCrossUITests/GradientTests.swift
stackotter: I don't know why this comment is attached to the code so confusingly, but what I meant was that it seems 'adjustedStops' doesn't change, even if you swap the startRadius and the endRadius (i.e. adjustedStops is the same for both radialStartingAtZeroReturnsOriginalStops and radialNegativeRangeReturnsOriginalStops). That feels weird and I wanted to make sure that's the intended behaviour?

Sources/SwiftCrossUI/Backend/BackendFeatures/Gradients.swift
```diff
@@ -0,0 +1,43 @@
+extension BackendFeatures {
```
stackotter, line 1: Move this into a `Gradients` subdirectory, then split the various protocols into their own separate files and make this `Gradients.swift` files just house a `typealias Gradients = LinearGradients & RadialGradients & AngularGradients` that people can use when implementing backends that support all gradient features

Sources/SwiftCrossUI/Backend/BackendFeatures/Gradients.swift
```diff
@@ -0,0 +1,43 @@
...
+    /// Backend Methods for Linear Gradients
+    ///
+    /// Used by ``LinearGradient``
+    public protocol LinearGradients: Core {
+        func createLinearGradient() -> Widget
```
stackotter: Document each of these backend methods

stackotter: I think these gradient methods should be named `createFooGradientWidget` and `updateFooGradientWidget` to make it clear that they're creating widgets to display gradients, and not creating some sort of gradient object (like a fill or brush), because we may add that kind of stuff in future.

Sources/SwiftCrossUI/Backend/FullAppBackend.swift
```diff
@@ -41,6 +41,9 @@ public typealias FullAppBackend =
     & BackendFeatures.Colors
     & BackendFeatures.DatePickers
     & BackendFeatures.Windowing
+    & BackendFeatures.LinearGradients
+    & BackendFeatures.RadialGradients
+    & BackendFeatures.AngularGradients
```
stackotter: Make this `& BackendFeatures.Gradients` once you've created that typealias

Sources/SwiftCrossUI/Values/Angle.swift
```diff
@@ -0,0 +1,66 @@
+import Foundation
+
+/// A geometric angle whose value you access in either radians or degrees.
+public struct Angle {
```
stackotter: Conform Angle to Hashable and Sendable

Sources/SwiftCrossUI/Values/Angle.swift
```diff
@@ -0,0 +1,66 @@
...
+    ///   - destination: The end point used to calculate the angle from the origin.
+    public init(origin: UnitPoint, destination: UnitPoint) {
+        let deltaX = destination.x - origin.x
+        let deltaY = destination.y - origin.y
+
+        self.init(radians: atan2(deltaY, deltaX))
+    }
+
+    /// The factor for converting an angle in degrees to the same angle in radians.
+    public static let conversionFactor = Double.pi / 180
```
stackotter: I think this should be private to avoid people using it themselves; it's more of an internal detail

Sources/SwiftCrossUI/Values/UnitPoint.swift
stackotter: Ah good point, I didn't think of that; makes total sense

Sources/SwiftCrossUI/Values/UnitPoint.swift
```diff
@@ -0,0 +1,48 @@
+/// A normalized 2D point in a view's coordinate space.
+///
+/// points matching 0 <= x <= 1 and 0 <= y <= 1 are inside the bounds of the view,
```
stackotter:
```
/// The point's coordinates start from the view's top leading corner, and
/// are relative to the views size.
///
/// Coordinates between 0 and 1 are inside the bounds of the view, and coordinates
/// outside of that range are outside of the view.
```

Sources/SwiftCrossUI/Values/UnitPoint.swift
stackotter: Hmm yeah I think I was interpreting 'normalised' in isolation (i.e. in the mathematical sense) rather that reading it as 'normalised with regard to the view that this unit point gets used with'. I think with the documentation changes that I've now suggested on the struct itself, I'm happy to leave the other docs as they are (even though I find the terminology Apple uses here a bit unclear).

Sources/SwiftCrossUI/Values/UnitPoint.swift
```diff
@@ -0,0 +1,48 @@
...
+/// > 1 or < 0 are outside of the view
+public struct UnitPoint {
```
stackotter: Conform this to Hashable and Sendable

Sources/SwiftCrossUI/Values/Gradient.swift
stackotter: I agree that the warning is good, I just meant that we should also update the documentation comment to tell developers that the array must be sorted. It looks like you've already done that

Sources/SwiftCrossUI/Values/Gradient.swift
```diff
@@ -0,0 +1,86 @@
+/// A color gradient represented as an array of color stops, each having a normalized location value.
+public struct Gradient: Sendable, Hashable {
+    /// The array of color stops, ordered by location.
+    public var stops: [Gradient.Stop]
+
+    /// Creates a gradient from an array of color stops ordered by location.
+    ///
+    /// If no stop is passed, the gradient will be fully transparent.
+    init(stops: [Gradient.Stop]) {
```
stackotter, line 10: Document function/init parameters (shows up nicer in the documentation if we do)

Sources/SwiftCrossUI/Views/Gradients/AngularGradient.swift
```diff
@@ -0,0 +1,165 @@
+/// An angular gradient, often also referred to as a conic gradient.
```
stackotter: `, often referred to as a conic gradient.`

Sources/SwiftCrossUI/Views/Gradients/AngularGradient.swift
stackotter: Ah ok, sounds like the code is fine but the documentation clearly confusing. I think just update the docs to spell out that 0 corresponds to 0 degrees, and 1 corresponds to 360 degrees.

Sources/SwiftCrossUI/Views/Gradients/AngularGradient.swift
```diff
@@ -0,0 +1,165 @@
...
+        endAngle: Angle
+    ) {
+        self.gradient = gradient
+        self.center = center
+        self.startAngle = startAngle
+        self.endAngle = endAngle
+    }
+    
+    @CastBackend<BackendFeatures.AngularGradients>(
+        backendGenericName: "NewBackend",
```
stackotter: I believe that you can omit `backendGenericName` (below as well). You only need it when you specifically want to refer to the new backend's type by name.

Sources/SwiftCrossUI/Views/Gradients/AngularGradient.swift
```diff
@@ -0,0 +1,165 @@
...
+    ) {
+        self.init(
+            gradient: Gradient(stops: stops),
+            center: center,
+            startAngle: startAngle,
+            endAngle: endAngle
+        )
+    }
+
+    public var adjustedStops: [Gradient.Stop] {
```
stackotter: Document what this does, and if it's just a convenient helper for built-in backends, we should probably make this a `package var` so that we're free to change its behaviour in future if needed to support more adjustments

Sources/SwiftCrossUI/Views/Gradients/LinearGradient.swift
```diff
@@ -0,0 +1,91 @@
+/// A linear gradient.
+public struct LinearGradient: ElementaryView {
+    /// The gradient represented as an array of color stops, each having a parametric location value.
+    public let gradient: Gradient
+    /// The normalized point where the gradient begins, defined in the views coordinate space.
```
stackotter: `the view's`

Sources/SwiftCrossUI/Views/Gradients/RadialGradient.swift
```diff
@@ -0,0 +1,126 @@
+/// A radial gradient.
+public struct RadialGradient: ElementaryView {
+    /// The gradient represented as an array of color stops, each having a parametric location value.
+    public let gradient: Gradient
+    /// The radius at which the first gradient stop will be placed.
+    ///
+    /// All space outside inside radius gets filled with the color of the first gradient stop.
```
stackotter: `All space inside this radius ...`

Sources/SwiftCrossUI/Views/Gradients/RadialGradient.swift
stackotter: I think it's easy enough for us to support right? Because we can just automatically reverse the gradient. Use cases where it's probably most applicable would be graphical editors where the user is entering the radii with draggable handles or something similar. I assume some backends that don't use adjustedStops may already support having the radii swapped?

Sources/SwiftCrossUI/Views/Gradients/RadialGradient.swift
```diff
@@ -0,0 +1,126 @@
...
+        self.init(
+            gradient: Gradient(colors: colors),
+            center: center,
+            startRadius: startRadius,
+            endRadius: endRadius
+        )
+    }
+
+    /// Stops adjusted to accomodate startRadius on backends without native support.
+    public var adjustedStops: [Gradient.Stop] {
```
stackotter: This should probably be a `package var` so that we can change its behaviour in future (because it's just a helper for built-in backends)

Sources/AppKitBackend/AppKitBackend+Gradient.swift
stackotter: I think that the stretching would also effect the angles of the radial lines of equal colour in conic gradients? My assumption is that SwiftUI doesn't do that, but maybe it does

Sources/GtkBackend/GtkBackend+Gradient.swift
stackotter: I think you missed this one

stackotter (changes requested):
Thanks for applying all that feedback! I've got a few more things, but I think most of them are relatively small things this time (organisation, documentation, etc). I don't have time to test things locally right now, but I will try to do that soon, because I may be able to resolve some of my own questions and comments by testing out the edge cases that I'm worried about.

Tests/SwiftCrossUITests/GradientTests.swift
MiaKoring: Ahh yeah, I’m not sure why I did it that way. I tested it in SwiftUI and they support inverted start and end locations. I fixed it for us 👍

Sources/AppKitBackend/AppKitBackend+Gradient.swift
MiaKoring: It does, yes, but as far as I can tell its consistent with SwiftUI

MiaKoring: but Gtk doesn’t. FUCK

MiaKoring:
GtkBackend will not support the AngularGradient type right now due to rendering inconsistencies with SwiftUI and the Apple Backends which are not easily resolvable.

For future reference if needed:
```swift
public func createAngularGradient() -> Widget {
    Box()
}

public func updateAngularGradient(
    _ widget: Widget,
    gradient: AngularGradient,
    withSize size: SIMD2<Int>,
    in environment: EnvironmentValues
) {
    let widget = widget as! Box

    let adjustedStops = gradient.adjustedStops

    let stops = adjustedStops.map { stop in
        let resolved = stop.color.resolve(in: environment)
        let red = resolved.red * 255
        let green = resolved.green * 255
        let blue = resolved.blue * 255
        
        let location = stop.location * 360
        
        return "rgba(\(red), \(green), \(blue), \(resolved.opacity)) \(location)deg”
    }.joined(separator: ", “)
    
    let startDegrees = gradient.startAngle.degrees + 90
    let centerXPercent = gradient.center.x * 100
    let centerYPercent = gradient.center.y * 100

    widget.css.set(
        property: .init(
            key: “background”,
            value: “””
                conic-gradient(from \(startDegrees)deg \
                at \(centerXPercent)% \(centerYPercent)%, \
                \(stops))
                “””
        )
    )
}
```

Examples/Sources/GradientsExample/GradientsApp.swift
MiaKoring: as discussed on discord delayed until the mentioned GTK bug is resolved

MiaKoring:
I believe I’m caught up. main is merged again too.

Examples/Sources/GradientsExample/GradientsApp.swift
```diff
@@ -2,77 +2,77 @@ import DefaultBackend
...
-    import SwiftBundlerRuntime
+import SwiftBundlerRuntime
 #endif
 
 @HotReloadable
 @main
 struct GradientsApp: App {
     static let colors: [Color] = [.red, .orange, .yellow, .green, .blue, .purple]
-
+
```
stackotter: This file has ended up with a lot of trailing whitespace added. Maybe Xcode did it?

Examples/Sources/GradientsExample/GradientsApp.swift
```diff
@@ -2,77 +2,77 @@ import DefaultBackend
 import SwiftCrossUI
 
 #if canImport(SwiftBundlerRuntime)
-    import SwiftBundlerRuntime
+import SwiftBundlerRuntime
```
stackotter: Indent one level

Sources/SwiftCrossUI/Backend/BackendFeatures/Gradients/AngularGradients.swift
```diff
@@ -0,0 +1,22 @@
...
+    public protocol AngularGradients: Core {
+        /// Creates the widget for an ``AngularGradient``.
+        func createAngularGradientWidget() -> Widget
+        
+        /// Updates the widget of an ``AngularGradient``.
+        /// - Parameters:
+        ///   - widget: The widget to update.
+        ///   - gradient: The SwiftCrossUI struct housing the information for the gradient's rendering.
+        ///   - size: The new size of the widget.
+        ///   - environment: The widgets environment, used to resolve its colors.
```
stackotter: `widget's`

Sources/SwiftCrossUI/Backend/BackendFeatures/Gradients/LinearGradients.swift
```diff
@@ -0,0 +1,22 @@
...
+    public protocol LinearGradients: Core {
+        /// Creates the widget for a ``LinearGradient``.
+        func createLinearGradientWidget() -> Widget
+        
+        /// Updates the widget of a ``LinearGradient``.
+        /// - Parameters:
+        ///   - widget: The widget to update.
+        ///   - gradient: The SwiftCrossUI struct housing the information for the gradient's rendering.
+        ///   - size: The new size of the widget.
+        ///   - environment: The widgets environment, used to resolve its colors.
```
stackotter: `widget's`

Sources/SwiftCrossUI/Backend/BackendFeatures/Gradients/AngularGradients.swift
```diff
@@ -0,0 +1,22 @@
+extension BackendFeatures {
+    /// Backend Methods for Angular (conic) Gradients
```
stackotter: Don't need to capitalise the `Methods` here, or the `Angular` and `Gradients` either

stackotter: End documentation sentences with periods for consistency with rest of our docs

Sources/SwiftCrossUI/Backend/BackendFeatures/Gradients.swift
```diff
@@ -1,43 +1,6 @@
...
-        func createAngularGradient() -> Widget
-        
-        func updateAngularGradient(
-            _ widget: Widget,
-            gradient: AngularGradient,
-            withSize size: SIMD2<Int>,
-            in environment: EnvironmentValues
-        )
-    }
+    public typealias Gradients =
```
stackotter, line 6: Document this

Sources/SwiftCrossUI/Values/UnitPoint.swift
```diff
@@ -1,8 +1,9 @@
-/// A normalized 2D point in a view's coordinate space.
```
stackotter: Sorry, that probably wasn't clear. I meant for those two paragraphs to replace the second paragraph, but to leave the overview sentence there as well;

```swift
/// A normalized 2D point in a view's coordinate space.
///
/// There point's coordinates start...
```

Sources/SwiftCrossUI/Values/Gradient.swift
```diff
@@ -33,10 +34,9 @@ public struct Gradient: Sendable, Hashable {
     }
 
     /// Creates a gradient from an array of colors.
-    ///
-    /// The gradient synthesizes its location values to evenly space the colors along the gradient.
-    ///
-    /// If no color is passed, the gradient will be fully transparent.
+    /// - Parameters:
+    ///   - colors: The colors of the gradient. The gradient synthesizes its location values to evenly
+    ///   space the colors along the gradient. If no color is passed, the gradient will be fully transparent.
```
stackotter: Indent the wrapped line two more spaces (dunno if DocC works without doing so?)

```swift
///   - colors: blah blah ...
///     blah blah blah ...
```

Sources/SwiftCrossUI/Views/Gradients/AngularGradient.swift
```diff
@@ -29,7 +29,7 @@ public struct AngularGradient: ElementaryView {
 
     /// Creates an angular gradient that completes a partial rotation.
     ///
-    /// Stops are expected to be in 360° unit space.
+    /// For ``Gradient.Stop`` location of 0 corresponds to 0° and 1 to 360°.
```
stackotter:
```swift
/// For each ``Gradient.Stop``, a location of 0 corresponds to 0°, and a location of 1 corresponds to 360°.
```

Sources/SwiftCrossUI/Views/Gradients/RadialGradient.swift
```diff
@@ -4,17 +4,17 @@ public struct RadialGradient: ElementaryView {
     public let gradient: Gradient
     /// The radius at which the first gradient stop will be placed.
     ///
-    /// All space outside inside radius gets filled with the color of the first gradient stop.
+    /// All space inside radius gets filled with the color of the first gradient stop.
```
stackotter: `inside this radius`

Sources/SwiftCrossUI/Views/Gradients/RadialGradient.swift
```diff
@@ -4,17 +4,17 @@ public struct RadialGradient: ElementaryView {
...
     public let startRadius: Double
     /// The radius at which the last gradient stop will be placed.
     ///
     /// All space outside this radius gets filled with the color of the last gradient stop.
     public let endRadius: Double
     /// The normalized center point of the gradient in its coordinate space.
     public let center: UnitPoint
-
+
```
stackotter: Trailing whitespace

Sources/WinUIBackend/WinUIBackend+Gradient.swift
```diff
@@ -4,20 +4,20 @@ import WindowsFoundation
 import UWP
 
 extension WinUIBackend {
-    public func createLinearGradient() -> Widget {
+    public func createLinearGradientWidget() -> Widget {
         WinUI.Rectangle()
     }
-
-    public func updateLinearGradient(
+
```
stackotter: Trailing whitespace

Sources/GtkBackend/GtkBackend+Gradient.swift
```diff
@@ -42,19 +42,21 @@ extension GtkBackend {
...
+    public func updateRadialGradientWidget(
         _ widget: Widget,
         gradient: RadialGradient,
         withSize size: SIMD2<Int>,
         in environment: EnvironmentValues
     ) {
         let widget = widget as! Box
-
-        let stops = cssStops(gradient: gradient.gradient, environment: environment)
+        let stops = gradient.startRadius < gradient.endRadius ?
```
stackotter: Wrap the ternary like this;

```swift
let stops = gradient.startRadius < gradient.endRadius
  ? gradient.gradient.stops
  : invertedStops(stops: gradient.gradient.stops)
```

stackotter (changes requested):
Thanks for resolving those comments, the implementation looks great now. My feedback this time is purely formatting and documentation wording, so this should be the home straight.

I believe the @bbrk24 was having issues with Xcode adding additional trailing whitespace indents, so maybe some default setting has changed recently?

Examples/Sources/GradientsExample/GradientsApp.swift
MiaKoring: I’m pretty sure I didn’t do it intentially 😅 

is it possible to have the formatter remove it in the future? thats quite annoying to do manually haha

Examples/Sources/GradientsExample/GradientsApp.swift
MiaKoring: thats also something Xcode often messes up. compiler flags to the start of the line and no indent for the body. 

Perhaps there’s a setting for it? I’m not sure

Sources/SwiftCrossUI/Backend/BackendFeatures/Gradients/AngularGradients.swift
MiaKoring: sorry, german reflex sometimes 💀

Sources/GtkBackend/GtkBackend+Gradient.swift
MiaKoring: thats a way nicer way of doing it haha

Sources/SwiftCrossUI/Backend/BackendFeatures/Gradients/AngularGradients.swift
```diff
@@ -1,7 +1,7 @@
 extension BackendFeatures {
-    /// Backend Methods for Angular (conic) Gradients
+    /// Backend Methods for Angular (conic) Gradients.
```
stackotter: Fix the capitalisation here too

Sources/SwiftCrossUI/Backend/BackendFeatures/Gradients/LinearGradients.swift
```diff
@@ -1,7 +1,7 @@
 extension BackendFeatures {
-    /// Backend Methods for Linear Gradients
+    /// Backend Methods for Linear Gradients.
```
stackotter: Fix the capitalisation here too

stackotter (changes requested):
Just two more docs fixes 😅

stackotter (approved):
Lgtm! 🚀

PR #455, opened 2026-02-21
[GtkBackend+KDE] Dismiss sheet on close button press

### LINKED ISSUE #445: [GtkBackend] Built-in "X" button doesn't close sheet windows
Might be a KDE-specific thing? But I can only close sheets by pressing the Esc key. An "x" button does render in the pop-up window containing the sheet, but clicking it doesn't do anything.

bbrk24:
Fixes #445

Turns out we were already calling the onDismiss handler, and just not actually dismissing the sheet.

### DIFF
--- a/Sources/GtkBackend/GtkBackend.swift
+++ b/Sources/GtkBackend/GtkBackend.swift
@@ -21,7 +21,6 @@ public final class GtkBackend: AppBackend {
         var onDismiss: (() -> Void)? = nil
         var interactiveDismissDisabled = false
         var nestedSheet: Sheet?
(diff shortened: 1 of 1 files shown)


stackotter:
If this is ready then I'd be happy to merge. I tested on my Mac by commenting out the line that removes the sheet window decorations, and I can reproduce your issue without the PR but can't with the PR; so definitely seems to have fixed that issue.

stackotter:
Unfortunately that last change seems to have introduced those weird squared off corners on Ubuntu as well (it doesn't have them without the change). We may to find a way to disable the KDE window decorations more directly.

PR #460, opened 2026-02-24
Extension of PR #447

stackotter:
This extends #447 to address feedback I gave in my last round of review. The changes were minimal enough that I'm just addressing them myself to get this useful feature merged. I've mostly just updated the docs (with platform availability of each picker style), and separated public types into their own files.

### DIFF
--- a/Sources/UIKitBackend/UIKitBackend.swift
+++ b/Sources/UIKitBackend/UIKitBackend.swift
@@ -66,7 +66,11 @@ public final class UIKitBackend: AppBackend {
         #elseif os(visionOS)
             .menu
         #else
-            .wheel
+            if #available(iOS 14, macCatalyst 14, *) {
+                .menu
+            } else {
+                .wheel
+            }
         #endif
     }
 


stackotter:
I'm making this a draft for now, because there's an iOS picker style rendering issue that I'm getting @bbrk24's opinion on (on Discord). This is how `.menu` style pickers render on iOS 18.6.2, and it looks a little strange (especially when no option is selected because the picker ends up with no label on the button);

IMG_8899

stackotter:
## Updated UIKitBackend MenuPickerStyle screenshots

Picker with no selection;
IMG_8905

Picker with no selection, with menu popped over;
IMG_8906

Picker with selection;
IMG_8907

Picker with selection, with menu popped over;
IMG_8908

stackotter:
## SwiftUI screenshots for future reference

Empty Picker;
IMG_8909

Picker with selection;
IMG_8910

Picker with selection, with menu popped over;
IMG_8911

### Differences of note

- SwiftUI Pickers don't react to font size (and perhaps same with other controls?), not sure whether that's something we want to copy or not though
- SwiftUI dims the chevron icon and text equally when the menu is popped over; our implementation dims the label very slightly, and the icon way more than SwiftUI does

stackotter:
In both SwiftUI and SwiftCrossUI on iOS, only the chevrons are shown when using MenuPickerStyle with a `nil` selection. The `Flavor:` label is in both cases just text in a HStack with the Picker. Perhaps SwiftUI only uses a Picker's title when in a form?

PR #461, opened 2026-02-24
[GTK3] Notify settings with "gtk-shell-shows-menubar" on menubar change

gregcotten:
This fixes a regression introduced by https://github.com/moreSwift/swift-cross-ui/pull/325. 

This change caused the GTK3 application menu bar to get set _after_ the window is already shown, and apparently [GTK doesn't like that](https://gitlab.gnome.org/GNOME/gtk/-/issues/2834). 

This is apparently a #wontfix over on GNOME's gitlab:
```
I don't think application windows with menubars are common enough to fix this.
```
???

This is the only bandaid I could think of, and I'm not setup with a GTK4 environment to see if it's needed there.

### DIFF
--- a/Sources/Gtk3/Application.swift
+++ b/Sources/Gtk3/Application.swift
@@ -27,6 +27,13 @@ public class Application: GObject, GActionMap {
                 applicationPointer,
                 (newValue?.pointer).map(UnsafeMutablePointer.init)
             )
+
+            // works around issue with adding a menubar after showing the window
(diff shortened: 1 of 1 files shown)


stackotter:
Thanks for investigating this! That seems like a suitable workaround to me. I’ll test on my own machine this morning and merge if all is well.

> I don't think application windows with menubars are common enough to fix this.

Man, the Gtk devs really confuse me sometimes…

stackotter (approved):
Just tested locally and it worked perfectly. Thanks for this!

PR #463, opened 2026-02-26
[FIX] namespaced observation ignored noop

MiaKoring:
When you import **Foundation** `@ObservationIgnored` is ambiguous. 
To fix it you would use `@SwiftCrossUI.ObservationIgnored`. 
The namespaced variant wasn’t covered by the macro yet.

This PR adds 
- support for namespaced identifiers in `Variable/hasMacroApplication(_)`
- additional check for the namespaced variant of the macro
- a test validating the fix

changes in other files are caused by the format script


stackotter (approved):
Thanks for catching and fixing that!

PR #467, opened 2026-03-03
Add inset(by:) modifier

### LINKED ISSUE #466: inset(by:)
https://developer.apple.com/documentation/swiftui/insettableshape/inset(by:)

I think in theory this can be done without needing InsettableShape

```swift
struct InsetShape<Base: Shape>: Shape {
  var inset: Double
  var base: Base

  func path(in bounds: Path.Rect) -> Path {
    base.path(in: .init(x: bounds.x + inset, y: bounds.y + inset, width: bounds.width - 2 * inset, height: bounds.height - 2 * inset))
  }

  func size(fitting proposal: ProposedViewSize) -> ViewSize {
    let innerProposal = ProposedViewSize(width: proposal.width.map { max(0, $0 - 2 * inset) }, height: proposal.height.map { max(0, $0 - 2 * inset) })
    let innerSize = base.size(fitting: innerProposal)
    return .init(width: innerSize.width + 2 * inset, height: innerSize.height + 2 * inset)
  }
}

extension InsetShape: StyledShape where Base: StyledShape { ... }
```

but we may want to match SwiftUI anyways.

bbrk24:
Fixes #466 

An easy way to test the rounded rectangle one is to comment out the `.inset(by: 2)` in PathsExample on WinUI or GTK. Because those backends clip children to their bounds, the stroke will look thinner along the edges than in the corners if the shape isn't properly inset.

### DIFF
--- a/Sources/SwiftCrossUI/Views/Shapes/InsettableShape.swift
+++ b/Sources/SwiftCrossUI/Views/Shapes/InsettableShape.swift
@@ -8,7 +8,7 @@ public protocol InsettableShape: Shape {
 }
 
 /// The `InsetShape` implementation used by ``Rectangle``, ``Ellipse``, ``Circle``, and ``Capsule``.
-/// 
+///
 /// This implementation only works for convex shapes where insetting the shape is equivalent to
 /// making the shape smaller.
 struct InsettableShapeImpl<Base: Shape>: InsettableShape {
--- a/Sources/SwiftCrossUI/Views/Shapes/RoundedRectangle.swift
+++ b/Sources/SwiftCrossUI/Views/Shapes/RoundedRectangle.swift
@@ -456,20 +456,20 @@ extension RoundedRectangle: Shape {
 
 // MARK: InsettableShape
 extension RoundedRectangle: InsettableShape {
-    public func inset(by amount: Double) -> InsetShape {
-        InsetShape(initialCornerRadius: cornerRadius, insetAmount: amount)
+    public func inset(by amount: Double) -> some InsettableShape {
+        InsetShapeImpl(initialCornerRadius: cornerRadius, insetAmount: amount)
     }
 
-    public struct InsetShape {
+    struct InsetShapeImpl {
         var initialCornerRadius: Double
         var insetAmount: Double
     }
 }
 
-extension RoundedRectangle.InsetShape: InsettableShape {
+extension RoundedRectangle.InsetShapeImpl: InsettableShape {
     private var actualCornerRadius: Double { max(0, initialCornerRadius - insetAmount) }
 
-    public func path(in bounds: Path.Rect) -> Path {
+    func path(in bounds: Path.Rect) -> Path {
         RoundedRectangle(cornerRadius: actualCornerRadius)
             .path(
                 in: .init(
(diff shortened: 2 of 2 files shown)


Sources/SwiftCrossUI/Views/Shapes/InsettableShape.swift
```diff
@@ -0,0 +1,43 @@
+/// A shape type that is able to inset itself to produce another shape.
+public protocol InsettableShape: Shape {
+    /// The type of the inset shape.
+    associatedtype InsetShape: InsettableShape
+
+    /// Returns `self` inset by `amount`.
+    nonisolated func inset(by amount: Double) -> InsetShape
+}
+
+/// The `InsetShape` implementation used by ``Rectangle``, ``Ellipse``, and ``Capsule``.
```
stackotter: Also `Circle`. Mention that this only works for nice enough shapes.

Sources/SwiftCrossUI/Views/Shapes/RoundedRectangle.swift
```diff
@@ -447,3 +447,42 @@ extension RoundedRectangle: Shape {
...
     }
 }
+
+// MARK: InsettableShape
+extension RoundedRectangle: InsettableShape {
+    public func inset(by amount: Double) -> InsetShape {
+        InsetShape(initialCornerRadius: cornerRadius, insetAmount: amount)
+    }
+
+    public struct InsetShape {
```
stackotter: Hide this implementation detail using `some InsettableShape` above if possible.

Sources/SwiftCrossUI/Views/Shapes/RoundedRectangle.swift
```diff
@@ -447,3 +447,42 @@ extension RoundedRectangle: Shape {
...
+
+    public func size(fitting proposal: ProposedViewSize) -> ViewSize {
+        let proposedWidth = proposal.width ?? 10
+        let proposedHeight = proposal.height ?? 10
+
+        return ViewSize(max(proposedWidth, insetAmount * 2), max(proposedHeight, insetAmount * 2))
+    }
+
+    public func inset(by amount: Double) -> RoundedRectangle.InsetShape {
+        .init(initialCornerRadius: initialCornerRadius, insetAmount: insetAmount + amount)
```
stackotter: Can spell this out as `Self(...)`

stackotter (changes requested):
Thanks for implementing this! Just requested a couple of docs/style changes, but impl looks great

Sources/SwiftCrossUI/Views/Shapes/RoundedRectangle.swift
bbrk24: I think to do that, I would have to make it a non-public top-level struct the way `InsettableShapeImpl` already is, and I didn't really see much point in doing that here. The struct already has no public properties or initializers.

stackotter: Why does it have to be top-level?

bbrk24: I'm pretty sure if I just change the return type of `inset(by:)` without also changing this declaration, the compiler will complain about the type mismatch and not satisfying the protocol. (That being said, I've been wrong about the type checker's behavior before.) I suppose I could also nest this inside the definition of the method itself but I assume that's undesirable.

stackotter: Just tried it out locally, and you're correct that it doesn't work with just a return type change (because the nested type matches the associatedtype name). But renaming the nested type to InsetShapeImpl fixed it.
[fix.patch](https://github.com/user-attachments/files/25808775/fix.patch)

PR #484, opened 2026-03-11
Gtk: Set ellipsize to none to measure text with nil height (fixes #438)

### LINKED ISSUE #438: Text() doesn’t wrap inside ScrollView() on GtkBackend

stackotter:
That was easier than I was expected 🎉

### DIFF
--- a/Sources/GtkBackend/GtkBackend.swift
+++ b/Sources/GtkBackend/GtkBackend.swift
@@ -753,7 +753,7 @@ public final class GtkBackend: AppBackend {
         let pango = Pango(for: widget)
         let (width, height) = pango.getTextSize(
             text,
-            ellipsize: ellipsize,
+            ellipsize: proposedHeight == nil ? .none : ellipsize,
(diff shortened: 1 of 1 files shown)


MiaKoring:
should the counter app be reset to its previous state? currently its not counting

stackotter:
Yeah just saw that when going to test out your line limit scenario 😅 whoops

MiaKoring:
Can confirm it works

PR #486, opened 2026-03-11
Mouseover help

### LINKED ISSUE #435: Hover tooltips
Something like HTML `title`, with a fallback for touchscreens. (Touch screen != mobile, e.g. Surface laptops that can fold back into a tablet.)

Perhaps based on https://developer.apple.com/documentation/swiftui/view/help(_:)

bbrk24:
Fixes #435 

Like in SwiftUI, the `.help(_:)` function is technically available on iOS 14, but on that version it only sets the accessibility hint text and does nothing else.

### DIFF
--- a/Sources/AppKitBackend/AppKitBackend.swift
+++ b/Sources/AppKitBackend/AppKitBackend.swift
@@ -481,9 +481,7 @@ public final class AppKitBackend: AppBackend {
     }
     
     public func createTooltipContainer(wrapping child: NSView) -> NSView {
-        let container = createContainer()
-        insert(child, into: container, at: 0)
-        return container
+        child
     }
     
     public func updateTooltipContainer(_ widget: NSView, tooltip: String) {
--- a/Sources/SwiftCrossUI/Backend/AppBackend.swift
+++ b/Sources/SwiftCrossUI/Backend/AppBackend.swift
@@ -633,7 +633,10 @@ public protocol AppBackend: Sendable {
     )
 
     /// Create a container capable of showing a textual tooltip.
-    /// 
+    ///
+    /// If no container is necessary, this method is allowed to return `child`
+    /// unmodified.
+    ///
     /// - Parameters:
     ///   - child: The widget being wrapped to show a tooltip over.
     func createTooltipContainer(wrapping child: Widget) -> Widget
@@ -641,7 +644,7 @@ public protocol AppBackend: Sendable {
     /// 
     /// - Parameters:
     ///   - widget: The widget to update the tooltip for. Will always have been
-    ///     created by ``createTooltipContainer()``.
+    ///     created by ``createTooltipContainer(wrapping:)``.
     ///   - tooltip: The text to be shown on hover.
     func updateTooltipContainer(_ widget: Widget, tooltip: String)
 
--- a/Sources/WinUIBackend/WinUIBackend.swift
+++ b/Sources/WinUIBackend/WinUIBackend.swift
@@ -660,6 +660,7 @@ public final class WinUIBackend: AppBackend {
     }
     
     public func createTooltipContainer(wrapping child: Widget) -> Widget {
+        // TODO(bbrk24): Look into removing the container, like on AppKit
         TooltipContainer(child: child)
     }
 


Sources/SwiftCrossUI/Backend/AppBackend.swift
```diff
@@ -632,6 +632,19 @@ public protocol AppBackend: Sendable {
...
+    /// - Parameters:
+    ///   - child: The widget being wrapped to show a tooltip over.
+    func createTooltipContainer(wrapping child: Widget) -> Widget
+    /// Update the tooltip shown by a widget.
+    /// 
+    /// - Parameters:
+    ///   - widget: The widget to update the tooltip for. Will always have been
+    ///     created by ``createTooltipContainer()``.
+    ///   - tooltip: The text to be shown on hover.
+    func updateTooltipContainer(widget: Widget, tooltip: String)
```
stackotter: Make the first parameter unlabelled to match general AppBackend API pattern (e.g. `updateTextView(_:content:environment:)`

Sources/AppKitBackend/AppKitBackend.swift
```diff
@@ -479,6 +479,16 @@ public final class AppKitBackend: AppBackend {
             widget.heightAnchor.constraint(equalToConstant: proposedHeight).isActive = true
         }
     }
+    
+    public func createTooltipContainer(wrapping child: NSView) -> NSView {
+        let container = createContainer()
+        insert(child, into: container, at: 0)
+        return container
+    }
+
```
stackotter, line 486: Does it work if we just return `child` here?

Sources/Gtk3Backend/Gtk3Backend.swift
```diff
@@ -1675,3 +1684,43 @@ class CustomLabel: Label {
...
 }
+
+final class TooltipContainer: Fixed {
+    private var tooltip: UnsafeMutableBufferPointer<CChar>
+
+    init(_ child: Widget) {
+        self.tooltip = UnsafeMutableBufferPointer(start: nil, count: 0)
+        super.init()
+        self.put(child, x: 0, y: 0)
+        child.parentWidget = self
```
stackotter: `put` already sets the child's parent widget

Sources/GtkBackend/GtkBackend.swift
```diff
@@ -1943,6 +1952,46 @@ class CustomLabel: Label {
...
 }
 
+final class TooltipContainer: Fixed {
+    private var tooltip: UnsafeMutableBufferPointer<CChar>
+
+    init(_ child: Widget) {
+        self.tooltip = UnsafeMutableBufferPointer(start: nil, count: 0)
+        super.init()
+        self.put(child, x: 0, y: 0)
+        child.parentWidget = self
```
stackotter: `put` already sets the child's parent widget

stackotter (changes requested):
Thanks for implementing this 🙏 Just a couple of suggestions and questions. Going to test locally now

stackotter:
All seems to work locally (although I haven't tested things like Mac Catalyst given that I believe that you have tested pretty thoroughly yourself)

Sources/AppKitBackend/AppKitBackend.swift
bbrk24: I wasn't sure how the layout system would handle something being the same widget instance as its parent, so I did this just to be safe.

stackotter: I believe that it should work in this case. There are a few other places (such as tap gesture containers) where we return the child widget without wrapping it on some platforms. If you do update AppKitBackend to do so, make sure to update the AppBackend docs to reflect that some backends may just return the child.

bbrk24: If that's the case, then WinUIBackend might be able to do that too -- but I don't currently have access to a Windows machine to test it. Gtk/Gtk3 almost could, but then deallocating the C string (currently done in deinit) would be challenging. The only backend that categorically can't is UIKitBackend, because it needs the interaction delegate.

stackotter: I'd be happy for it to just be AppKitBackend for now, cause it's the one that stood out as trivial. The others can be left as todos if they seem doable but annoying

PR #517, opened 2026-04-09
Add support for ColorScheme detection on Linux with the GTK backend.

### LINKED ISSUE #386: [GTK] Support dark mode
As you can see in the screenshots below, the text colors are appropriate for light mode, but not for dark mode:

Image
Image

I am using Fedora 43.

tosage05:
Fixes #386

Does so by making use of `org.freedesktop.appearance->color-scheme` through dbus (method supported by most desktop environments), both at launch and while the scui app is running.

### DIFF
--- a/Sources/Gtk/Utility/Freedesktop.swift
+++ b/Sources/Gtk/Utility/Freedesktop.swift
@@ -1,0 +1,855 @@
+import CGtk
+import Foundation
+import GtkCHelpers
+import Mutex
+
+public enum FreedesktopError: Error {
+    public enum BusError: Error {
+        case failedCall
+        case malformedInput
+        case noProxy
+    }
+}
+
+private class CallbackBox<T> {
+    var value: T
+
+    init(_ value: T) {
+        self.value = value
+    }
+}
+
+private typealias DBusSignalHandler =
+    @convention(c) (
+        OpaquePointer?, UnsafePointer<CChar>?, UnsafePointer<CChar>?,
+        UnsafePointer<CChar>?, UnsafePointer<CChar>?, OpaquePointer?,
+        UnsafeMutableRawPointer?
+    ) -> Void
+
+public class Bus {
+    private var proxy: UnsafeMutablePointer<GDBusProxy>
+    private var interfaceName: String
+    private var subscriptions: [(UInt, Any)] = []
+
+    init(interfaceName: String) throws(FreedesktopError.BusError) {
+        self.interfaceName = interfaceName
+        let p = g_dbus_proxy_new_for_bus_sync(
+            G_BUS_TYPE_SESSION,
+            G_DBUS_PROXY_FLAGS_NONE,
+            nil,
+            "org.freedesktop.portal.Desktop",
+            "/org/freedesktop/portal/desktop",
+            "org.freedesktop.portal.\(interfaceName)",
+            nil,
+            nil
+        )
+
+        guard let p else { throw FreedesktopError.BusError.noProxy }
+        self.proxy = p
+    }
+
+    deinit {
+        for (handlerId, _) in subscriptions {
+            g_signal_handler_disconnect(self.proxy, handlerId)
+        }
+        subscriptions = []
+        g_object_unref(proxy)
+    }
+
+    public func call(namespace: String, key: String, method: String)
+        throws(FreedesktopError.BusError) -> GVariant
+    {
+        guard let v = GVariant.parse("('\(namespace)', '\(key)')") else { throw .malformedInput }
+
+        let output = g_dbus_proxy_call_sync(
+            proxy,
+            method,
+            v.pointer,
+            G_DBUS_CALL_FLAGS_NONE,
+            gint(-1),
+            nil,
+            nil
+        )
+
+        guard let output else { throw .failedCall }
+
+        return GVariant(owning: output)
+    }
+
+    public func call(method: String, parameters: GVariant)
+        throws(FreedesktopError.BusError) -> GVariant
+    {
+        let output = g_dbus_proxy_call_sync(
+            proxy,
+            method,
+            parameters.pointer,
+            G_DBUS_CALL_FLAGS_NONE,
+            gint(-1),
+            nil,
+            nil
+        )
+
+        guard let output else { throw .failedCall }
+
+        return GVariant(owning: output)
+    }
+
+    public func readProperty(_ name: String) -> GVariant? {
+        guard let prop = g_dbus_proxy_get_cached_property(proxy, name) else { return nil }
+        return GVariant(owning: prop)
+    }
+
+    public func connection() -> OpaquePointer? {
+        g_dbus_proxy_get_connection(proxy)
+    }
+
+    public func subscribe(
+        signal: String, namespace: String, key: String, handler: @escaping (OpaquePointer) -> Void
+    ) -> UInt {
+        let callback:
+            @convention(c) (
+                OpaquePointer?, UnsafePointer<CChar>?, UnsafePointer<CChar>?,
+                OpaquePointer?, UnsafeMutableRawPointer?
+            ) -> Void = { _, _, sig, params, data in
+                let dataTyped =
+                    Unmanaged<CallbackBox<(String, String, (OpaquePointer) -> Void)>>.fromOpaque(
+                        data!
+                    ).takeUnretainedValue()
+                let (namespaceInner, keyInner, handlerInner) = dataTyped.value
+
+                guard let walker = GVariant(params),
+                    walker.child(0)?.string() == namespaceInner,
+                    walker.child(1)?.string() == keyInner
+                else { return }
+
+                if let ptr = walker.child(2)?.pointer { handlerInner(ptr) }
+            }
+
+        let data = CallbackBox((namespace, key, handler))
+
+        let handlerId = g_signal_connect_data(
+            self.proxy, signal,
+            unsafeBitCast(callback, to: GCallback.self),
+            Unmanaged.passUnretained(data).toOpaque(),
+            nil, .init(0)
+        )
+
+        subscriptions.append((handlerId, data))
+
+        return handlerId
+    }
+
+    public func unsubscribe(from: UInt) {
+        g_signal_handler_disconnect(self.proxy, from)
+        subscriptions.removeAll { $0.0 == from }
+    }
+}
+
+private func decodeLocationData(from raw: OpaquePointer)
+    -> Freedesktop.Portal.Location.LocationData?
+{
+    guard let walker = GVariant(raw) else { return nil }
+    let timestamp = walker.lookup("Timestamp")
+    return Freedesktop.Portal.Location.LocationData(
+        latitude: walker.lookup("Latitude")?.double(),
+        longitude: walker.lookup("Longitude")?.double(),
+        altitude: walker.lookup("Altitude")?.double(),
+        accuracy: walker.lookup("Accuracy")?.double(),
+        speed: walker.lookup("Speed")?.double(),
+        heading: walker.lookup("Heading")?.double(),
+        timestampSeconds: timestamp?.child(0)?.int64(),
+        timestampMicroseconds: timestamp?.child(1)?.int64()
+    )
+}
+
+private func parseLocationUpdate(
+    params: OpaquePointer, sessionHandle: String
+) -> Freedesktop.Portal.Location.LocationData? {
+    guard GVariant(params)?.child(0)?.string() == sessionHandle else { return nil }
+    return GVariant(params)?.child(1).flatMap { decodeLocationData(from: $0.pointer) }
+}
+
+private func requestResponse(
+    bus: Bus, method: String, input: (String) -> GVariant?
+) throws(FreedesktopError.BusError) -> GVariant {
+    guard let connection = bus.connection() else { throw .failedCall }
+    guard let uniqueName = g_dbus_connection_get_unique_name(connection) else { throw .failedCall }
+
+    let token = "swift_cross_ui_\(UUID().uuidString.replacingOccurrences(of: "-", with: "_"))"
+    let sender = String(cString: uniqueName)
+        .replacingOccurrences(of: ":", with: "")
+        .replacingOccurrences(of: ".", with: "_")
+    let expectedHandle = "/org/freedesktop/portal/desktop/request/\(sender)/\(token)"
+
+    typealias State = (done: Bool, response: UInt32, results: GVariant?)
+    let state = UnsafeMutablePointer<State>.allocate(capacity: 1)
+    state.initialize(to: (false, 2, nil))
+    defer {
+        state.deinitialize(count: 1)
+        state.deallocate()
+    }
+
+    let callback: DBusSignalHandler = { _, _, _, _, _, params, data in
+        guard let params, let data else { return }
+        let s = data.assumingMemoryBound(to: State.self)
+        guard let walker = GVariant(params),
+            let response = walker.child(0)?.uint32()
+        else { return }
+        s.pointee.response = response
+        if let child1 = walker.child(1) {
+            s.pointee.results = GVariant(owning: g_variant_ref(child1.pointer))
+        }
+        s.pointee.done = true
+    }
+
+    var sub = g_dbus_connection_signal_subscribe(
+        connection, "org.freedesktop.portal.Desktop",
+        "org.freedesktop.portal.Request", "Response",
+        expectedHandle, nil, G_DBUS_SIGNAL_FLAGS_NONE,
+        callback, UnsafeMutableRawPointer(state), nil
+    )
+
+    guard let parameters = input(token) else {
+        g_dbus_connection_signal_unsubscribe(connection, sub)
+        throw .malformedInput
+    }
+
+    let result = try bus.call(method: method, parameters: parameters)
+
+    guard let actualHandle = result.child(0)?.string() else {
+        g_dbus_connection_signal_unsubscribe(connection, sub)
+        throw .failedCall
+    }
+
+    if actualHandle != expectedHandle {
+        g_dbus_connection_signal_unsubscribe(connection, sub)
+        sub = g_dbus_connection_signal_subscribe(
+            connection, "org.freedesktop.portal.Desktop",
+            "org.freedesktop.portal.Request", "Response",
+            actualHandle, nil, G_DBUS_SIGNAL_FLAGS_NONE,
+            callback, UnsafeMutableRawPointer(state), nil
+        )
+    }
+    defer { g_dbus_connection_signal_unsubscribe(connection, sub) }
+
+    while !state.pointee.done { g_main_context_iteration(nil, 1) }
+
+    guard state.pointee.response == 0, let results = state.pointee.results else {
+        throw .failedCall
+    }
+
+    state.pointee.results = nil
+    return results
+}
+
+public enum Freedesktop {
+    public enum Portal {
+        public enum Settings {
+            fileprivate static let bus = try! Bus(interfaceName: "Settings")
+
+            public enum Version {
+                public static func get() throws(FreedesktopError.BusError) -> UInt32 {
+                    guard let prop = bus.readProperty("version")
+                    else { throw .failedCall }
+                    return prop.uint32()
+                }
+            }
+
+            fileprivate struct SettingField<T> {
+                let namespace: String
+                let key: String
+                let decodeResult: (OpaquePointer) -> T?
+                let decodeSignal: (OpaquePointer) -> T?
+
+                func read() throws(FreedesktopError.BusError) -> T {
+                    let result = try bus.call(namespace: namespace, key: key, method: "Read")
+                    guard let value = decodeResult(result.pointer) else { throw .failedCall }
+                    return value
+                }
+
+                func observe() -> AsyncStream<T> {
+                    AsyncStream { continuation in
+                        let id = bus.subscribe(
+                            signal: "g-signal::SettingChanged",
+                            namespace: namespace,
+                            key: key
+                        ) { raw in
+                            guard let value = decodeSignal(raw) else { return }
+                            continuation.yield(value)
+                        }
+                        continuation.onTermination = { _ in bus.unsubscribe(from: id) }
+                    }
+                }
+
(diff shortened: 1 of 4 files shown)


Sources/GtkBackend/GtkBackend.swift
```diff
@@ -471,12 +471,109 @@ public final class GtkBackend: AppBackend {
...
-        defaultEnvironment
+        defaultEnvironment.with(\.colorScheme, getPreferredColorScheme())
     }
 
     public func setRootEnvironmentChangeHandler(to action: @escaping () -> Void) {
-        // TODO: React to theme changes
+        guard
+            let proxy = g_dbus_proxy_new_for_bus_sync(
+                G_BUS_TYPE_SESSION,
+                .init(0),
```
stackotter: I think you may be able to use `G_DBUS_PROXY_FLAGS_NONE` here? Otherwise use an explicit initialiser such as `gint(0)` or `guint(0)` depending on the required underlying type

Sources/GtkBackend/GtkBackend.swift
```diff
@@ -471,12 +471,109 @@ public final class GtkBackend: AppBackend {
...
+                let proxy = g_dbus_proxy_new_for_bus_sync(
+                    G_BUS_TYPE_SESSION,
+                    .init(0),
+                    nil,
+                    "org.freedesktop.portal.Desktop",
+                    "/org/freedesktop/portal/desktop",
+                    "org.freedesktop.portal.Settings",
+                    nil,
+                    nil
+                )
```
stackotter: Can we instead initialise this proxy once when initialising the backend and then reuse it as necessary?

Sources/GtkBackend/GtkBackend.swift
```diff
@@ -471,12 +471,109 @@ public final class GtkBackend: AppBackend {
...
+            else { return .light }
```
stackotter: As a formatting thing, we generally put the else body on its own line if the guard condition got its own line;

```
guard 
    ...
else {
    ...
}

File: Sources/GtkBackend/GtkBackend.swift [hunk 4]
```diff
@@ -471,12 +471,109 @@ public final class GtkBackend: AppBackend {
...
+            defer { g_object_unref(proxy) }
+
+            guard
+                let v =
+                    ("('org.freedesktop.appearance', 'color-scheme')".withCString {
+                        g_variant_parse(nil, $0, nil, nil, nil)
```
stackotter: I believe that you can pass the string as an argument directly and Swift should do the necessary conversion to a c string (because the parameter is a `const char *`)

Sources/GtkBackend/GtkBackend.swift
```diff
@@ -471,12 +471,109 @@ public final class GtkBackend: AppBackend {
...
+                    })
+            else { return .light }
+            defer { g_variant_unref(v) }
+
+            guard
+                let outerOuter = g_dbus_proxy_call_sync(
+                    proxy,
+                    "Read",
+                    v,
+                    .init(0),
```
stackotter: Is it possible to use `G_DBUS_CALL_FLAGS_NONE` here?

Sources/GtkBackend/GtkBackend.swift
```diff
@@ -471,12 +471,109 @@ public final class GtkBackend: AppBackend {
...
+                    .init(-1),
```
stackotter: Explicitly cast this to `gint` instead of using `.init` (just a style choice I've made for the code base)

Sources/GtkBackend/GtkBackend.swift
```diff
@@ -471,12 +471,109 @@ public final class GtkBackend: AppBackend {
...
+
+            guard let inner = g_variant_get_variant(outer) else { return .light }
+            defer { g_variant_unref(inner) }
+
+            guard let final = g_variant_get_variant(inner) else { return .light }
+            defer { g_variant_unref(final) }
+
+            let colorSchemeId = g_variant_get_uint32(final)
+
+            if colorSchemeId == 1 {
```
stackotter, line 520: Link to `https://github.com/flatpak/xdg-desktop-portal/blob/fae2c09a530eeef30af8d6401870fd5bae5d3989/data/org.freedesktop.portal.Settings.xml#L38` in a comment so that it's easy for people to look up these values

Sources/GtkBackend/GtkBackend.swift
```diff
@@ -471,12 +471,109 @@ public final class GtkBackend: AppBackend {
...
+                nil
+            )
+        else { return }
+
+        let handler:
+            @convention(c) (
+                OpaquePointer?, UnsafePointer<CChar>?, UnsafePointer<CChar>?,
+                OpaquePointer?, UnsafeMutableRawPointer?
+            ) -> Void = { _, _, sig, params, data in
+                let namespaceVariant = g_variant_get_child_value(params, 0)
```
stackotter: Do we have to also unref params to avoid a memory leak here?

Sources/GtkBackend/GtkBackend.swift
```diff
@@ -471,12 +471,109 @@ public final class GtkBackend: AppBackend {
         )
     }
 
+    private func getPreferredColorScheme() -> ColorScheme {
+        #if os(Linux)
+            guard
+                let proxy = g_dbus_proxy_new_for_bus_sync(
```
stackotter: Would you feel comfortable refactoring these two proxy implementations (proxy.call_sync and setting the signal handler) into a wrapper class in the `Gtk` target? I think it'd be great to have this bundled up in a reusable way so that we can easily fetch other settings too in future. For reference I'd recommend looking at the existing TextBuffer wrapper because it has an init, and a signal handler example.

stackotter (changes requested):
Thanks for tackling this! This will be a nice QoL improvement for Linux users. I've requested some changes, and I'll test locally next

Sources/GtkBackend/GtkBackend.swift
tosage05: I believe we only need to `unref` references we create, such as ones from `g_variant_get_*`... In this case `params` is a reference created by the caller of our handler, which `unref`s it after we're done with it (tested with a `defer unref(params)` in handler which caused an over-unref error to be logged).

stackotter: Ah yeah good point, I think I was looking at it as something we retrieved and forgot it was from the signal, disregard that

Sources/Gtk/Utility/GVariant.swift
```diff
@@ -0,0 +1,68 @@
...
+    private init(owning pointer: OpaquePointer) {
+        self.ptr = pointer
+        self.shouldUnrefOnDeinit = true
+    }
+
+    deinit {
+        if shouldUnrefOnDeinit { g_variant_unref(ptr) }
+    }
+
+    public var pointer: OpaquePointer { ptr }
```
stackotter: I believe that this is unnecessary? You can rename `ptr` to `pointer` instead and make it public

Sources/Gtk/Utility/Freedesktop.swift
```diff
@@ -0,0 +1,787 @@
...
+                public static func observe(options: Options = .init()) -> AsyncStream<Int?> {
+                    _field.observe(options: options)
+                }
+                public static func subscribe(
+                    options: Options = .init(),
+                    _ handler: @escaping (Int?) -> Void
+                ) -> Task<Void, Never> { _field.subscribe(options: options, handler) }
+            }
+        }
+
```
stackotter, line 853: Remove this blank line

Sources/Gtk/Utility/Freedesktop.swift
```diff
@@ -0,0 +1,787 @@
...
+                        for await value in observe() {
+                            handler(value)
+                        }
+                    }
+                }
+            }
+
+            public enum Appearance {
+                public enum ColorScheme {
+                    private static let _field = _SettingField(
```
stackotter: When I said that I wanted to reduce the number of lines in this file I should've probably been a bit more clear. I really meant to just reduce repetition, but the code itself should still be formatted for readability even if it uses up more lines.

E.g. each member of this these enums should have a blank line between them, and function bodies should always get their own line (to conform to our general code style).

Sources/Gtk/Utility/Freedesktop.swift
```diff
@@ -0,0 +1,787 @@
...
+                        throw .failedCall
+                    }
+                    return value
+                }
+
+                public enum Id {
+                    public static func read() throws(FreedesktopError.BusError) -> String {
+                        try UserInformation.read("id")
+                    }
+                }
```
stackotter, line 446: Put a blank line between each declaration

Sources/Gtk/Utility/Freedesktop.swift
```diff
@@ -0,0 +1,787 @@
...
+            public enum UserInformation {
+                private static func read(_ key: String) throws(FreedesktopError.BusError)
+                    -> String
+                {
+                    let raw = try requestResponse(
+                        bus: bus,
+                        method: "GetUserInformation"
+                    ) { token in
+                        g_variant_parse(
+                            nil, "('', @a{sv} {'handle_token': <'\(token)'>})", nil, nil, nil)
```
stackotter: Might be worth adding a comment to explain this a little more. Is the empty string the window identifier, and the rest is options dictionary? We should probably link to the docs too for people to refer to; https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Account.html#org-freedesktop-portal-account-getuserinformation

Sources/Gtk/Utility/Freedesktop.swift
```diff
@@ -0,0 +1,787 @@
...
+                }
+            }
+
+            private static var _bus: Bus?
+            private static var _sessionHandle: String?
+            private static var _started = false
+            private static let _lock = NSRecursiveLock()
+
+            private static func _getBus() throws(FreedesktopError.BusError) -> Bus {
+                if let bus = _bus { return bus }
```
stackotter: Private members don't need underscore prefixes, so we should remove them

Sources/Gtk/Utility/Freedesktop.swift
```diff
@@ -0,0 +1,787 @@
...
+                    self.distanceThreshold = distanceThreshold
+                    self.timeThreshold = timeThreshold
+                    self.parentWindow = parentWindow
+                }
+            }
+
+            private static var _bus: Bus?
+            private static var _sessionHandle: String?
+            private static var _started = false
+            private static let _lock = NSRecursiveLock()
```
stackotter: We should use a `Mutex` to protect shared state, rather than using a lock, because it's harder to mess up, and neater.

Sources/Gtk/Utility/Freedesktop.swift
```diff
@@ -0,0 +1,787 @@
...
+                    parts.append("'distance-threshold': <uint32 \(options.distanceThreshold)>")
+                }
+                if options.timeThreshold != 0 {
+                    parts.append("'time-threshold': <uint32 \(options.timeThreshold)>")
+                }
+                guard
+                    let input = g_variant_parse(
+                        nil, "(@a{sv} {\(parts.joined(separator: ", "))},)", nil, nil, nil)
+                else { throw .malformedInput }
+                defer { g_variant_unref(input) }
```
stackotter: These defers are easy to miss when rearranging code. We should probably add a `static func parse(_ string: String) -> GVariant?` method to GVariant (which can then handle unrefing automatically). You can then also update `call` to return a GVariant (owned).

Sources/Gtk/Utility/Freedesktop.swift
```diff
@@ -0,0 +1,787 @@
...
+                        nil, nil, nil)
+                }
+                g_variant_unref(response)
+                _started = true
+            }
+
+            public enum Version {
+                public static func get() throws(FreedesktopError.BusError) -> UInt32 {
+                    let bus = try _getBus()
+                    guard let prop = bus.readProperty("version"),
```
stackotter: readProperty should return a `GVariant`

Sources/Gtk/Utility/Freedesktop.swift
```diff
@@ -0,0 +1,787 @@
...
+                    .BusError) -> Double?
+                { try _field.read(options: options) }
+                public static func observe(options: Options = .init()) -> AsyncStream<Double?> {
+                    _field.observe(options: options)
+                }
+                public static func subscribe(
+                    options: Options = .init(),
+                    _ handler: @escaping (Double?) -> Void
+                ) -> Task<Void, Never> { _field.subscribe(options: options, handler) }
+            }
```
stackotter, line 809: This is all really dense, so spacing things out properly should aid readability. E.g. blank lines between functions, and function bodies on their own lines. The `throws(...)` clause of `read` has also been wrapped rather oddly

stackotter (changes requested):
I've reviewed GVariant.swift and GtkBackend.swift, but haven't finished reviewing Freedesktop.swift. Hopefully the comments that are there so far help you continue working on it until I get time to finish the review though.

Thanks for your refactoring of this stuff by the way! The GVariant wrapper has made your variant handling code much nicer than it was before

Sources/Gtk/Utility/GVariant.swift
```diff
@@ -0,0 +1,80 @@
...
+        guard let v = g_variant_parse(nil, format, nil, nil, nil) else { return nil }
+        self.init(owning: v)
+    }
+
+    public static func parse(_ format: String) -> GVariant? {
+        guard let v = g_variant_parse(nil, format, nil, nil, nil) else { return nil }
+        return GVariant(owning: v)
+    }
+
+    public func copy() -> GVariant {
```
stackotter, line 30: The name `copy` seems misleading. I think this should be called `cloneReference` instead, if the method is even needed at all

Sources/Gtk/Utility/GVariant.swift
```diff
@@ -0,0 +1,80 @@
...
+
+    public convenience init?(parsing format: String) {
+        guard let v = g_variant_parse(nil, format, nil, nil, nil) else { return nil }
+        self.init(owning: v)
+    }
+
+    public static func parse(_ format: String) -> GVariant? {
+        guard let v = g_variant_parse(nil, format, nil, nil, nil) else { return nil }
+        return GVariant(owning: v)
+    }
```
stackotter, line 28: These are identical. We should only keep one of them

Sources/Gtk/Utility/Freedesktop.swift
```diff
@@ -0,0 +1,855 @@
...
+                            continuation.finish()
+                            return
+                        }
+                        connection = conn
+                    } catch {
+                        continuation.finish()
+                        return
+                    }
+
+                    nonisolated(unsafe) let conn = connection
```
stackotter, line 627: I'd prefer if we could avoid `nonisolated(unsafe)`. Perhaps we can wrap the connection's OpaquePointer in an `@unchecked Sendable` wrapper class. Then there's only one piece of code to audit (rather than having to individually audit each `nonisolated(unsafe)` use site.

Sources/Gtk/Utility/Freedesktop.swift
```diff
@@ -0,0 +1,855 @@
...
+                    let box = CallbackBox((sessionHandle, continuation))
+
+                    let callback: DBusSignalHandler = { _, _, _, _, _, params, data in
+                        guard let params, let data else { return }
+                        let b = Unmanaged<
```
stackotter, line 632: Avoid opaque variable names. `box` would be suitable here

Sources/Gtk/Utility/Freedesktop.swift
```diff
@@ -0,0 +1,855 @@
...
+
+            fileprivate static func observeFull(
+                options: Options = .init()
+            ) -> AsyncStream<LocationData> {
+                AsyncStream { continuation in
+                    let sessionHandle: String
+                    let connection: OpaquePointer
+                    do {
+                        try startSession(options: options)
+                        guard let sh = session.withLock({ $0.handle }) else {
```
stackotter, line 611: Maybe rename this to `sessionHandleValue` or something (`sh` is a bit confusing)

stackotter (changes requested):
That's all looking much better now, thanks for applying all of that feedback!

I've requested some changes, and I'm also going to test this out locally. I'm excited to see how it goes

stackotter:
Oh, and would you be able to do a merge instead of a rebase this time when merging in the changes from upstream? This PR has a lot of code and rebases unfortunately destroy GitHub's PR review experience (even though there's no reason that they should given that GitHub can show me a diff of the rebase/force-push if I ask it nice enough)

stackotter:
<details>
<summary>Here's a backtrace in case it's useful</summary>

```
0 0x00007f5b60c6d084 g_wakeup_signal + 164 in libglib-2.0.so.0.8400.4
 1 0x00007f5b60c6827c g_variant_get_maybe + 155 in libglib-2.0.so.0.8400.4
 2 GVariant.variant() + 54 in CounterExample at /home/stackotter/scratch/mirror/Users/stackotter/Desktop/Projects/SwiftCrossUI/PRs/swift-cross-ui/Sources/Gtk/Utility/GVariant.swift:44:27

    42│ 
    43│     public func variant() -> GVariant? {
    44│         guard let inner = g_variant_get_variant(pointer) else { return nil }                                                   
      │                           ▲
    45│         return GVariant(owning: inner)
    46│     }

 3 closure #2 in variable initialization expression of static Freedesktop.Portal.Settings.Appearance.ColorScheme.field + 94 in CounterExample at /home/stackotter/scratch/mirror/Users/stackotter/Desktop/Projects/SwiftCrossUI/PRs/swift-cross-ui/Sources/Gtk/Utility/Freedesktop.swift:301:55

   299│                             GVariant($0)?.child(0)?.variant()?.variant()?.uint32()
   300│                         },
   301│                         decodeSignal: { GVariant($0)?.variant()?.uint32() }                                                    
      │                                                       ▲
   302│                     )
   303│

 4 closure #1 in closure #1 in Freedesktop.Portal.Settings.SettingField.observe() + 310 in CounterExample at /home/stackotter/scratch/mirror/Users/stackotter/Desktop/Projects/SwiftCrossUI/PRs/swift-cross-ui/Sources/Gtk/Utility/Freedesktop.swift:277:47

   275│                             key: key
   276│                         ) { raw in
   277│                             guard let value = decodeSignal(raw) else { return }                                                
      │                                               ▲
   278│                             continuation.yield(value)
   279│                         }

 5 closure #1 in Bus.subscribe(signal:namespace:key:handler:) + 2380 in CounterExample at /home/stackotter/scratch/mirror/Users/stackotter/Desktop/Projects/SwiftCrossUI/PRs/swift-cross-ui/Sources/Gtk/Utility/Freedesktop.swift:125:57

   123│                 else { return }
   124│ 
   125│                 if let ptr = walker.child(2)?.pointer { handlerInner(ptr) }                                                    
      │                                                         ▲
   126│             }
   127│

 6 0x00007f5b61c62aea g_closure_invoke + 361 in libgobject-2.0.so.0.8400.4
 7 0x00007f5b61c80aba signal_emit_unlocked_R.isra.0 + 2121 in libgobject-2.0.so.0.8400.4
 8 0x00007f5b61c82af6 signal_emit_valist_unlocked + 4533 in libgobject-2.0.so.0.8400.4
 9 0x00007f5b61c82d68 signal_emit_valist_unlocked + 5159 in libgobject-2.0.so.0.8400.4
10 0x00007f5b61c82e23 g_signal_emit + 146 in libgobject-2.0.so.0.8400.4
11 0x00007f5b607e47e1 on_signal_received + 256 in libgio-2.0.so.0.8400.4
12 0x00007f5b607cbcd0 emit_signal_instance_in_idle_cb + 127 in libgio-2.0.so.0.8400.4
13 0x00007f5b60c1fe0d g_idle_dispatch + 44 in libglib-2.0.so.0.8400.4
14 0x00007f5b60c19863 g_main_context_dispatch_unlocked.lto_priv.0 + 466 in libglib-2.0.so.0.8400.4
15 0x00007f5b60c227a8 g_main_context_iterate_unlocked.isra.0 + 727 in libglib-2.0.so.0.8400.4
16 0x00007f5b60c22953 g_main_context_iterate_unlocked.isra.0 + 1154 in libglib-2.0.so.0.8400.4
17 0x00007f5b607a98bd g_application_run + 492 in libgio-2.0.so.0.8400.4
18 Application.run(_:) + 282 in CounterExample at /home/stackotter/scratch/mirror/Users/stackotter/Desktop/Projects/SwiftCrossUI/PRs/swift-cross-ui/Sources/Gtk/Application.swift:76:22

    74│         self.windowCallback = windowCallback
    75│ 
    76│         let status = g_application_run(applicationPointer.cast(), 0, nil)                                                      
      │                      ▲
    77│         g_object_unref(applicationPointer)
    78│         return Int(status)

19 GtkBackend.runMainLoop(_:) + 209 in CounterExample at /home/stackotter/scratch/mirror/Users/stackotter/Desktop/Projects/SwiftCrossUI/PRs/swift-cross-ui/Sources/GtkBackend/GtkBackend.swift:116:16

   114│ 
   115│     public func runMainLoop(_ callback: @escaping @MainActor () -> Void) {
   116│         gtkApp.run { window in                                                                                                 
      │                ▲
   117│             self.measurementCustomLabel = (self.createTextView() as! CustomLabel)
   118│             self.precreatedWindow = window

20 _App.run() + 211 in CounterExample at /home/stackotter/scratch/mirror/Users/stackotter/Desktop/Projects/SwiftCrossUI/PRs/swift-cross-ui/Sources/SwiftCrossUI/_App.swift:54:17

    52│     /// Runs the app using the app's selected backend.
    53│     func run() {
    54│         backend.runMainLoop { [self] in                                                                                        
      │                 ▲
    55│             let baseEnvironment = EnvironmentValues(backend: backend)
    56│             environment = backend.computeRootEnvironment(

21 static App.main() + 491 in CounterExample at /home/stackotter/scratch/mirror/Users/stackotter/Desktop/Projects/SwiftCrossUI/PRs/swift-cross-ui/Sources/SwiftCrossUI/App.swift:151:14

   149│             }
   150│         }
   151│         _app.run()                                                                                                             
      │              ▲
   152│     }
   153│

22 0x00007f5b60508575 __libc_start_call_main + 116 in libc.so.6
23 0x00007f5b60508628 __libc_start_main@GLIBC_2.2.5 + 135 in libc.so.6

Backtrace took 1.82s
```
</details>

PR #520, opened 2026-04-10
fix: change frame dimensions from Int to Double for SwiftUI compatibility

### LINKED ISSUE #356: Frame width and height
In SwiftUI, the width and height of a frame are `CGFloat` whereas in SwiftCrossUI, they are `Int`.
This caused me some issues because a `GeometryProxy` offered an infinite height.

I've seen patterns where Apple recommends to use `.frame(width: proxy.size.width, height: proxy.size.height)`, but the type mismatch makes that impossible here.

mvanhorn:
## Summary

SwiftUI uses `CGFloat` (aliased to `Double`) for frame dimensions, but SwiftCrossUI used `Int`. This made it impossible to pass `GeometryProxy` sizes directly to `.frame(width:height:)`.

Changes `width`, `height`, `minWidth`, `idealWidth`, `minHeight`, and `idealHeight` parameters from `Int?` to `Double?` across both `frame()` overloads, their backing view structs, and the `KeyboardToolbar` frame helper. Removes the now-unnecessary `.map(Double.init)` conversions that existed internally.

## Changes

- `Sources/SwiftCrossUI/Views/Modifiers/Layout/FrameModifier.swift`: Both `frame()` functions, `StrictFrameView`, and `FlexibleFrameView` updated from `Int?` to `Double?`
- `Sources/UIKitBackend/KeyboardToolbar.swift`: `frame(width:)` updated from `Int?` to `Double?`

## Breaking change

Callers passing integer literals (`.frame(width: 100)`) continue to work since Swift auto-promotes `Int` literals to `Double`. Callers storing `Int` variables and passing them will need to convert to `Double`.

Fixes #356

This contribution was developed with AI assistance (Codex).

### DIFF
--- a/Sources/SwiftCrossUI/Views/Modifiers/Layout/FrameModifier.swift
+++ b/Sources/SwiftCrossUI/Views/Modifiers/Layout/FrameModifier.swift
@@ -8,6 +8,19 @@ extension View {
     ///   - height: The frame's exact height. `nil` lets the view choose its own
     ///     height instead.
     ///   - alignment: How to align the view within its container.
+    @available(*, deprecated, renamed: "frame(width:height:alignment:)")
+    public func frame(
+        width: Int? = nil,
+        height: Int? = nil,
+        alignment: Alignment = .center
+    ) -> some View {
+        return frame(
+            width: width.map(Double.init),
+            height: height.map(Double.init),
+            alignment: alignment
+        )
+    }
+
     public func frame(
         width: Double? = nil,
         height: Double? = nil,
@@ -38,6 +51,27 @@ extension View {
     ///   - maxHeight: The frame's maximum height. `nil` means the frame inherits
     ///     the maximum height of its content
     ///   - alignment: How to align the view within its container.
+    @available(*, deprecated, renamed: "frame(minWidth:idealWidth:maxWidth:minHeight:idealHeight:maxHeight:alignment:)")
+    public func frame(
+        minWidth: Int? = nil,
+        idealWidth: Int? = nil,
+        maxWidth: Int? = nil,
+        minHeight: Int? = nil,
+        idealHeight: Int? = nil,
+        maxHeight: Int? = nil,
+        alignment: Alignment = .center
+    ) -> some View {
+        return frame(
+            minWidth: minWidth.map(Double.init),
+            idealWidth: idealWidth.map(Double.init),
+            maxWidth: maxWidth.map(Double.init),
+            minHeight: minHeight.map(Double.init),
+            idealHeight: idealHeight.map(Double.init),
+            maxHeight: maxHeight.map(Double.init),
+            alignment: alignment
+        )
+    }
+
     public func frame(
         minWidth: Double? = nil,
         idealWidth: Double? = nil,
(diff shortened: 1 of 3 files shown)


stackotter (changes requested):
Thanks for making these changes! I believe that this resolves #356.

To avoid the breaking changes, would you be able to add back the original two frame modifiers (with Ints) as thin wrappers around the new Double-based modifiers, and just mark them as deprecated+renamed? If you need an example of how to mark the methods as renamed, there's an example near the top of the Slider implementation.

Btw, thanks for declaring your AI usage, I really appreciate when people do that. For a small PR like this it doesn't make too much difference of course, but it can definitely affect how I approach bigger PRs

stackotter:
Looks like the CI failures are just due to the tests still using the old Int-based frame modifiers

mvanhorn:
Fixed in 7342700 - updated the test to use `blueRectangleHeight` directly instead of wrapping in `Int()`.

stackotter:
> To avoid the breaking changes, would you be able to add back the original two frame modifiers (with Ints) as thin wrappers around the new Double-based modifiers, and just mark them as deprecated+renamed? If you need an example of how to mark the methods as renamed, there's an example near the top of the Slider implementation.

Will merge once that’s addressed too

mvanhorn:
Added deprecated Int-based frame modifiers in b11ff90. Both the strict (`frame(width:height:alignment:)`) and flexible (`frame(minWidth:idealWidth:maxWidth:minHeight:idealHeight:maxHeight:alignment:)`) variants have Int overloads marked with `@available(*, deprecated, renamed:)` that delegate to the Double versions.

Sources/SwiftCrossUI/Views/Modifiers/Layout/FrameModifier.swift
```diff
@@ -38,12 +51,33 @@ extension View {
...
     ///   - alignment: How to align the view within its container.
+    @available(*, deprecated, renamed: "frame(minWidth:idealWidth:maxWidth:minHeight:idealHeight:maxHeight:alignment:)")
     public func frame(
         minWidth: Int? = nil,
         idealWidth: Int? = nil,
-        maxWidth: Double? = nil,
+        maxWidth: Int? = nil,
         minHeight: Int? = nil,
         idealHeight: Int? = nil,
+        maxHeight: Int? = nil,
```
stackotter: maxWidth and maxHeight should be `Double` even in this original deprecated method, because we already had them as `Double`s (to support SwiftUI-style `maxWidth: .infinity` stuff).

stackotter (changes requested):

mvanhorn:
Good catch - fixed in a5ccaaa. Changed `maxWidth` and `maxHeight` to `Double?` in the deprecated wrapper to match the original pre-PR signature, so `maxWidth: .infinity` callers still work. Internal `.map(Double.init)` removed for those params since they're already Double now.

stackotter (dismissed):

stackotter:
Oh no, some of the examples fail to build now. You could maybe try adding `@_disfavoredOverload` to the deprecated methods

mvanhorn:
Good call, pushed e0498f7 adding `@_disfavoredOverload` to both deprecated Int-based overloads. Verified `CounterExample` and `AdvancedCustomizationExample` (which uses numeric frame literals) both build against the updated library on macOS.

stackotter (approved):
Looks good now, thanks for fixing that

PR #524, opened 2026-04-17
feat(TextField): add numeric initializer with value binding

### LINKED ISSUE #379: Textfields for inputting numbers
SwiftUI supports text fields bound to values other than strings, and handles the formatting and conversion, which is quite handy:

https://developer.apple.com/documentation/swiftui/textfield/init(_:value:formatter:)

mvanhorn:
## Summary

Closes #379. Adds two new `TextField` initializers that accept a numeric binding instead of a `Binding<String>`, mirroring SwiftUI's `TextField(_:value:formatter:)` convenience:

- `init<V: BinaryInteger & LosslessStringConvertible>(_:value:)`
- `init<V: BinaryFloatingPoint & LosslessStringConvertible>(_:value:)`

## Why this matters

SwiftUI's `TextField(_ title: value: formatter:)` lets you edit a non-String value (e.g. `Int`, `Double`) and hides the string/number conversion. SwiftCrossUI only had the `Binding<String>` form, which meant callers had to wire up their own string↔value conversion and handle bad-parse edge cases manually. The new initializers close that ergonomic gap for the common numeric case.

Why LosslessStringConvertible, not Formatter: SwiftUI's API relies on `Foundation.Formatter`, which isn't cross-platform ergonomic for every SwiftCrossUI backend. The issue author's linked SwiftUI docs use the formatter approach, but the numeric-focused use case they mentioned is well served by Swift's own `LosslessStringConvertible` constraint, which `Int`, `Int32`, `Double`, `Float`, etc. already satisfy. If a Formatter-based initializer is wanted, it's a natural follow-up PR.

## Behaviour

- Getter: `String(value.wrappedValue)` — standard numeric-to-string.
- Setter: `if let parsed = V(newString) { value.wrappedValue = parsed }` — on parse failure (user types a letter, `1.2.3`, etc.), the binding is not updated. Previous value is preserved. This matches SwiftUI's formatter-rejected behaviour.

## Usage

```swift
@State var minNum: Int = 0
@State var ratio: Double = 0.5

var body: some View {
    VStack {
        TextField("Minimum", value: $minNum)
        TextField("Ratio", value: $ratio)
    }
}
```

## Testing

- `swift build --target SwiftCrossUI` — passes cleanly on macOS with the change. Full `swift build` was partially blocked by an unrelated WinUI header (`ShObjIdl.h` not found on macOS); that's a pre-existing Windows-target build constraint, unrelated to this PR.
- No existing initializer was modified, so string-bound TextField usage is unaffected.
- No new imports. Both initializers use only `Binding` (already in scope) and standard-library protocol composition.

## Scope note

Didn't add a Formatter-based overload here — mentioned above as a natural follow-up. Also didn't touch the existing `TextField` examples in `Examples/` since they're all string-bound; a follow-up example PR could showcase the numeric binding.

Closes #379

This contribution was developed with AI assistance (Codex).

### DIFF
--- a/Sources/SwiftCrossUI/Views/TextField.swift
+++ b/Sources/SwiftCrossUI/Views/TextField.swift
@@ -45,7 +45,7 @@ public struct TextField: ElementaryView, View {
         self._text = Binding(
             get: { String(value.wrappedValue) },
             set: { newString in
-                if let parsed = V(newString) {
+                if let parsed = V(newString), parsed != value.wrappedValue {
                     value.wrappedValue = parsed
                 }
             }
@@ -71,7 +71,7 @@ public struct TextField: ElementaryView, View {
         self._text = Binding(
             get: { String(value.wrappedValue) },
             set: { newString in
-                if let parsed = V(newString) {
+                if let parsed = V(newString), parsed != value.wrappedValue {
                     value.wrappedValue = parsed
                 }
             }
--- a/Sources/SwiftCrossUI/Views/TextField.swift
+++ b/Sources/SwiftCrossUI/Views/TextField.swift
@@ -26,6 +26,58 @@ public struct TextField: ElementaryView, View {
         self._text = value ?? Binding(get: { dummy }, set: { dummy = $0 })
     }
 
+    /// Creates an editable text field bound to a binary integer value.
+    ///
+    /// The field's content is kept in sync with `value` via simple string
+    /// conversion. When the user enters text that cannot be parsed as the
+    /// target integer type, the binding is not updated (the previous value
+    /// is preserved), mirroring the behaviour of SwiftUI's
+    /// `TextField(_:value:formatter:)` when the formatter fails.
+    ///
+    /// - Parameters:
+    ///   - placeholder: The label to show when the field is empty.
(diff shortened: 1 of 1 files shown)


stackotter (changes requested):
The code you've got looks good; nice and simple. I will request one behavioural change though. When testing locally I found that the following sequence `1` -> `1a` -> `1` (type an invalid char then backspace it) triggers a warning about unnecessary writes to the underlying binding. You can also reproduce the warning by pasting the current value of the text field into the text field.

```
2026-04-18T13:12:32+1000 warning SwiftCrossUI: [SwiftCrossUI] Unnecessary write to text Binding of TextField detected, please open an issue at https://github.com/moreSwift/swift-cross-ui/issues/new so we can fix it for AppKitBackend.
```

For now that warning can be resolved by only writing to the underlying value binding when the new value doesn't match its current value.

mvanhorn:
Fixed in ac0708e - guarded both setters so the binding is only written when the parsed value differs from `value.wrappedValue`. That should suppress the warning on the `1` -> `1a` -> `1` path and on paste-current-value.

stackotter (approved):
Thanks for addressing that, looks good now

PR #526, opened 2026-04-23
Add compatibilty with Swift's @Observable macro

### LINKED ISSUE #407: Add compatibilty with Swift's `@Observable` macro
The goal is that a view is automatically updated if it uses properties of `@Observable` classes in their `body` and those properties change. Some discussion has already happened in pull request <https://github.com/moreSwift/swift-cross-ui/pull/395>.

@stackotter I think a 'dirty' flag would be a good solution for reducing the number of calls to `body`. That is also what I was thinking. However, I also see that overhauling the state dependency tracking would be a big first contribution. And if this wasn't a performance problem until now, maybe I should just implement the `@Observable` compatibility by itself first and accept multiple `body` executions for now. What do you think?

/cc @MiaKoring

s-k:
Resolves #407.

This pull request does the following:

- A view now automatically updates when `@Observable` or [`@Perceptible`](https://github.com/pointfreeco/swift-perception) objects used inside the view's `body` change.
- Adds an example illustrating the new behavior.
- Removes the `Observable` typealias because it would clash with the new functionality.
- Obsoletes the `ObservationIgnored` macro in OSs where the Observation framework is available. This fixes a problem where this macro would clash with the `ObservationIgnored` macro from the Observation framework (see #407). In my testing, this solution fully fixes the clash. In OSs where Observation is available, users may need to `import Observation` to continue using `@ObservationIgnored` together with `@ObservableObject`. The obsoletion error message also instructs users to do so.

I have split the PR into two commits. The first one contains a straightforward implementation of the new functionality. The second one simplifies the implementation using a protocol. If you prefer a straightforward solution, the second commit could be dropped. It adds no new functionality.

### DIFF
--- a/Examples/Bundler.toml
+++ b/Examples/Bundler.toml
@@ -87,3 +87,8 @@ version = '0.1.0'
 identifier = 'dev.swiftcrossui.ColorsExample'
 product = 'ColorsExample'
 version = '0.1.0'
+
+[apps.ObservableExample]
+identifier = 'dev.swiftcrossui.ObservableExample'
+product = 'ObservableExample'
+version = '0.1.0'
--- a/Examples/Bundler.toml
+++ b/Examples/Bundler.toml
@@ -110,7 +110,12 @@ identifier = 'dev.swiftcrossui.TapGesturesExample'
 product = 'TapGesturesExample'
 version = '0.1.0'
 
- [apps.ControlFocusabilityTest]
- identifier = 'dev.swiftcrossui.ControlFocusabilityTest'
- product = 'ControlFocusabilityTest'
- version = '0.1.0'
+[apps.ControlFocusabilityTest]
+identifier = 'dev.swiftcrossui.ControlFocusabilityTest'
+product = 'ControlFocusabilityTest'
+version = '0.1.0'
+
+[apps.ObservableExample]
+identifier = 'dev.swiftcrossui.ObservableExample'
+product = 'ObservableExample'
+version = '0.1.0'
--- a/Examples/Sources/ObservableExample/ObservableApp.swift
+++ b/Examples/Sources/ObservableExample/ObservableApp.swift
@@ -1,0 +1,77 @@
+import DefaultBackend
+import SwiftCrossUI
+
+#if canImport(SwiftBundlerRuntime)
+    import SwiftBundlerRuntime
+#endif
+
+@main
+@HotReloadable
+struct ObservableApp: App {
+    @State var model = ObservableModel()
+    @State var count = 0
+
+    var body: some Scene {
+        WindowGroup(model.windowTitle) {
+            VStack(spacing: 32) {
+                Text(model.windowText)
+                HStack {
+                    View1(model: model)
+                        .padding()
+                    View2(model: model)
+                        .padding()
+                }
+                ModifyingView(model: model)
+                    .padding()
+                if !model.automaticModeIsOn {
+                    Button("Start automatic Mode") {
+                        model.startAutomaticMode()
+                    }
+                }
+            }
+            .padding()
+        }
+        .defaultSize(width: 400, height: 200)
+    }
+}
+
+struct View1: View {
+    let model: ObservableModel
+    
+    var body: some View {
+        VStack {
+            Text(model.view1Text)
+            ModifyingView(model: model)
+        }
+        .padding()
+        .background(Color.green)
+    }
+}
+
+struct View2: View {
+    let model: ObservableModel
+    
+    var body: some View {
+        VStack {
+            Text(model.view2Text)
+            ModifyingView(model: model)
+        }
+        .padding()
+        .background(Color.red)
+    }
+}
+
+struct ModifyingView: View {
+    @Bindable var model: ObservableModel
+    
+    var body: some View {
+        if !model.automaticModeIsOn {
+            VStack {
+                TextField("Window Title", text: $model.windowTitle)
+                TextField("Window Text", text: $model.windowText)
+                TextField("View 1 Text", text: $model.view1Text)
+                TextField("View 2 Text", text: $model.view2Text)
+            }
+        }
+    }
+}
--- a/Examples/Sources/ObservableExample/ObservableApp.swift
+++ b/Examples/Sources/ObservableExample/ObservableApp.swift
@@ -23,7 +23,11 @@ struct ObservableApp: App {
                 }
                 ModifyingView(model: model)
                     .padding()
-                if !model.automaticModeIsOn {
+                if model.automaticModeIsOn {
+                    Button("Stop automatic Mode") {
+                        model.stopAutomaticMode()
+                    }
+                } else {
                     Button("Start automatic Mode") {
                         model.startAutomaticMode()
                     }
--- a/Examples/Sources/ObservableExample/ObservableApp.swift
+++ b/Examples/Sources/ObservableExample/ObservableApp.swift
@@ -41,7 +41,7 @@ struct ObservableApp: App {
 
 struct View1: View {
     let model: ObservableModel
-    
+
     var body: some View {
         VStack {
             Text(model.view1Text)
@@ -54,7 +54,7 @@ struct View1: View {
 
 struct View2: View {
     let model: ObservableModel
-    
+
     var body: some View {
         VStack {
             Text(model.view2Text)
@@ -67,7 +67,7 @@ struct View2: View {
 
 struct ModifyingView: View {
     @Bindable var model: ObservableModel
-    
+
     var body: some View {
         if !model.automaticModeIsOn {
             VStack {
--- a/Examples/Sources/ObservableExample/ObservableModel.swift
+++ b/Examples/Sources/ObservableExample/ObservableModel.swift
@@ -1,0 +1,58 @@
+import Foundation
+import Perception // Also works with `Observation`
+
+private var animalList = [
+    "Dog",
+    "Cat",
+    "Horse",
+    "Elephant",
+    "Giraffe",
+    "Zebra",
+    "Mouse",
+    "Bird",
+    "Fish",
+    "Lizard",
+    "Turtle",
+    "Octopus",
+    "Snake",
+    "Crab",
+    "Ant",
+    "Bee",
+    "Butterfly",
+    "Bat",
+    "Bat-eared fox",
+    "Owl",
+]
+
+@Perceptible // Also works with `@Observable`
+class ObservableModel {
+    var automaticModeIsOn = false
+    var windowTitle: String = "Window Title"
+    var windowText: String = "Window Text"
+    var view1Text: String = "View 1 Text"
+    var view2Text: String = "View 2 Text"
+    
+    func startAutomaticMode() {
+        guard !automaticModeIsOn else { return }
+        automaticModeIsOn = true
+        Task {
+            while true {
+                // Wait one second before changing the next text
+                try await Task.sleep(nanoseconds: 1_000_000_000)
+                
+                let animal = animalList.randomElement()!
+                let textIndex = Int.random(in: 0..<4)
+                switch textIndex {
+                case 0:
+                    windowTitle = animal
+                case 1:
+                    windowText = animal
+                case 2:
+                    view1Text = animal
+                default:
+                    view2Text = animal
+                }
+            }
+        }
+    }
+}
--- a/Examples/Sources/ObservableExample/ObservableModel.swift
+++ b/Examples/Sources/ObservableExample/ObservableModel.swift
@@ -1,5 +1,5 @@
 import Foundation
-import Perception // Also works with `Observation`
+import SwiftCrossUI
 
 private var animalList = [
     "Dog",
@@ -24,14 +24,14 @@ private var animalList = [
     "Owl",
 ]
 
-@Perceptible // Also works with `@Observable`
+@Observable
 class ObservableModel {
     var automaticModeIsOn = false
     var windowTitle: String = "Window Title"
     var windowText: String = "Window Text"
     var view1Text: String = "View 1 Text"
     var view2Text: String = "View 2 Text"
-    @PerceptionIgnored private var automaticModeTask: Task<Void, any Error>?
+    @ObservationIgnored private var automaticModeTask: Task<Void, any Error>?
     
     func startAutomaticMode() {
         guard !automaticModeIsOn else { return }
--- a/Examples/Sources/ObservableExample/ObservableModel.swift
+++ b/Examples/Sources/ObservableExample/ObservableModel.swift
@@ -33,6 +33,12 @@ class ObservableModel {
     var view2Text: String = "View 2 Text"
     @ObservationIgnored private var automaticModeTask: Task<Void, any Error>?
     
+    // Namespaced spellings of the ObservationIgnored and ObservationTracked
+    // macros are also supported when your deployment target is below that of
+    // the Observation module
+    @ObservationPolyfill.ObservationIgnored private var ignored1 = false
+    @SwiftCrossUI.ObservationIgnored private var ignored2 = false
+    
     func startAutomaticMode() {
         guard !automaticModeIsOn else { return }
         automaticModeIsOn = true
@@ -44,14 +50,14 @@ class ObservableModel {
                 let animal = animalList.randomElement()!
                 let textIndex = Int.random(in: 0..<4)
                 switch textIndex {
-                case 0:
-                    windowTitle = animal
-                case 1:
-                    windowText = animal
-                case 2:
-                    view1Text = animal
-                default:
-                    view2Text = animal
+                    case 0:
+                        windowTitle = animal
+                    case 1:
+                        windowText = animal
+                    case 2:
+                        view1Text = animal
+                    default:
+                        view2Text = animal
                 }
             }
         }
--- a/Examples/Sources/ObservableExample/ObservableModel.swift
+++ b/Examples/Sources/ObservableExample/ObservableModel.swift
@@ -31,12 +31,13 @@ class ObservableModel {
     var windowText: String = "Window Text"
     var view1Text: String = "View 1 Text"
     var view2Text: String = "View 2 Text"
+    @PerceptionIgnored private var automaticModeTask: Task<Void, any Error>?
     
     func startAutomaticMode() {
         guard !automaticModeIsOn else { return }
         automaticModeIsOn = true
-        Task {
-            while true {
+        automaticModeTask = Task {
+            while !Task.isCancelled {
                 // Wait one second before changing the next text
                 try await Task.sleep(nanoseconds: 1_000_000_000)
                 
@@ -54,5 +55,12 @@ class ObservableModel {
                 }
             }
         }
+    }
+    
+    func stopAutomaticMode() {
+        guard automaticModeIsOn else { return }
+        automaticModeIsOn = false
+        automaticModeTask?.cancel()
+        automaticModeTask = nil
     }
 }
--- a/Examples/Sources/ObservableExample/ObservableModel.swift
+++ b/Examples/Sources/ObservableExample/ObservableModel.swift
@@ -1,6 +1,10 @@
 import Foundation
 import SwiftCrossUI
 
+#if canImport(Observation)
+    import Observation
+#endif
+
 private var animalList = [
     "Dog",
     "Cat",
@@ -36,8 +40,8 @@ class ObservableModel {
     // Namespaced spellings of the ObservationIgnored and ObservationTracked
     // macros are also supported when your deployment target is below that of
     // the Observation module
-    @ObservationPolyfill.ObservationIgnored private var ignored1 = false
-    @SwiftCrossUI.ObservationIgnored private var ignored2 = false
+    //@ObservationPolyfill.ObservationIgnored private var ignored1 = false
+    //@SwiftCrossUI.ObservationIgnored private var ignored2 = false
     
     func startAutomaticMode() {
         guard !automaticModeIsOn else { return }
--- a/Examples/Sources/ObservableExample/ObservableModel.swift
+++ b/Examples/Sources/ObservableExample/ObservableModel.swift
@@ -36,13 +36,13 @@ class ObservableModel {
     var view1Text: String = "View 1 Text"
     var view2Text: String = "View 2 Text"
     @ObservationIgnored private var automaticModeTask: Task<Void, any Error>?
-    
+
     // Namespaced spellings of the ObservationIgnored and ObservationTracked
     // macros are also supported when your deployment target is below that of
     // the Observation module
-    //@ObservationPolyfill.ObservationIgnored private var ignored1 = false
-    //@SwiftCrossUI.ObservationIgnored private var ignored2 = false
-    
+    // @ObservationPolyfill.ObservationIgnored private var ignored1 = false
+    // @SwiftCrossUI.ObservationIgnored private var ignored2 = false
+
     func startAutomaticMode() {
         guard !automaticModeIsOn else { return }
         automaticModeIsOn = true
@@ -50,7 +50,7 @@ class ObservableModel {
             while !Task.isCancelled {
                 // Wait one second before changing the next text
                 try await Task.sleep(nanoseconds: 1_000_000_000)
-                
+
                 let animal = animalList.randomElement()!
                 let textIndex = Int.random(in: 0..<4)
                 switch textIndex {
@@ -66,7 +66,7 @@ class ObservableModel {
             }
         }
     }
-    
+
     func stopAutomaticMode() {
         guard automaticModeIsOn else { return }
         automaticModeIsOn = false
--- a/Package.swift
+++ b/Package.swift
@@ -159,6 +159,10 @@ let package = Package(
             url: "https://github.com/swhitty/swift-mutex",
             .upToNextMinor(from: "0.0.6")
         ),
+        .package(
+            url: "https://github.com/pointfreeco/swift-perception.git",
+            from: "2.0.10"
+        ),
         // .package(
         //     url: "https://github.com/stackotter/TermKit",
         //     revision: "163afa64f1257a0c026cc83ed8bc47a5f8fc9704"
@@ -181,6 +185,7 @@ let package = Package(
                 .product(name: "ImageFormats", package: "swift-image-formats"),
                 .product(name: "Logging", package: "swift-log"),
                 .product(name: "Mutex", package: "swift-mutex"),
+                .product(name: "PerceptionCore", package: "swift-perception"),
             ],
             exclude: [
                 "Builders/ViewBuilder.swift.gyb",
--- a/Package.swift
+++ b/Package.swift
@@ -185,6 +185,10 @@ let package = Package(
             url: "https://github.com/apple/swift-collections",
             .upToNextMajor(from: "1.5.0")
         ),
+        .package(
+            url: "https://github.com/pointfreeco/swift-perception.git",
+            from: "2.0.10"
+        ),
         // .package(
         //     url: "https://github.com/stackotter/TermKit",
         //     revision: "163afa64f1257a0c026cc83ed8bc47a5f8fc9704"
@@ -207,6 +211,7 @@ let package = Package(
                 .product(name: "ImageFormats", package: "swift-image-formats"),
                 .product(name: "Logging", package: "swift-log"),
                 .product(name: "Mutex", package: "swift-mutex"),
+                .product(name: "PerceptionCore", package: "swift-perception"),
             ],
             exclude: [
                 "Builders/ViewBuilder.swift.gyb",
--- a/Package.swift
+++ b/Package.swift
@@ -186,8 +186,8 @@ let package = Package(
             .upToNextMajor(from: "1.5.0")
         ),
         .package(
-            url: "https://github.com/pointfreeco/swift-perception.git",
-            from: "2.0.10"
+            url: "https://github.com/moreSwift/swift-observation-polyfill",
+            branch: "main"
         ),
         // .package(
         //     url: "https://github.com/stackotter/TermKit",
@@ -211,7 +211,8 @@ let package = Package(
                 .product(name: "ImageFormats", package: "swift-image-formats"),
                 .product(name: "Logging", package: "swift-log"),
                 .product(name: "Mutex", package: "swift-mutex"),
-                .product(name: "PerceptionCore", package: "swift-perception"),
+                .product(name: "ObservationPolyfillCore", package: "swift-observation-polyfill"),
+                .product(name: "ObservationPolyfill", package: "swift-observation-polyfill"),
             ],
             exclude: [
                 "Builders/ViewBuilder.swift.gyb",
--- a/Package.swift
+++ b/Package.swift
@@ -187,7 +187,7 @@ let package = Package(
         ),
         .package(
             url: "https://github.com/moreSwift/swift-observation-polyfill",
-            branch: "main"
+            .upToNextMinor(from: "0.1.0")
         ),
         // .package(
         //     url: "https://github.com/stackotter/TermKit",
--- a/Package.swift
+++ b/Package.swift
@@ -187,7 +187,7 @@ let package = Package(
         ),
         .package(
             url: "https://github.com/moreSwift/swift-observation-polyfill",
-            .upToNextMinor(from: "0.1.0")
+            .upToNextMinor(from: "0.1.1")
         ),
         // .package(
         //     url: "https://github.com/stackotter/TermKit",
--- a/Sources/SwiftCrossUI/State/Bindable.swift
+++ b/Sources/SwiftCrossUI/State/Bindable.swift
@@ -1,0 +1,138 @@
+import Foundation
+
+/// A property wrapper that exposes bindings to writable properties on a
+/// mutable model object.
+///
+/// Use `Bindable` when you want controls to edit data stored in an object that
+/// participates in observation, such as a type annotated with `@Observable` or
+/// `@Perceptible` from the
+/// [Perception](https://github.com/pointfreeco/swift-perception) package.
+/// Applying `@Bindable` lets you use `$` to derive bindings for the object's
+/// mutable members.
+///
+///     @Perceptible
+///     final class Profile {
+///         var name = "Taylor"
+///         var receivesNewsletter = false
+///     }
+///
+///     struct ProfileEditor: View {
+///         @Bindable var profile: Profile
+///
+///         var body: some View {
+///             Form {
+///                 TextField("Name", text: $profile.name)
+///                 Toggle("Newsletter", isOn: $profile.receivesNewsletter)
+///             }
+///         }
+///     }
+///
+/// `Bindable` can also be applied to stored properties, globals, and local
+/// variables. This is useful when you already have an observable or perceptible
+/// object and need bindings for only part of a view hierarchy. For example,
+/// you can introduce a local `@Bindable` value inside `body`:
+///
+///     @Observable
+///     final class TaskItem: Identifiable {
+///         let id = UUID()
+///         var title = ""
+///     }
+///
+///     struct TaskListView: View {
+///         @State private var tasks = [TaskItem(), TaskItem()]
+///
+///         var body: some View {
+///             VStack {
+///                 ForEach(tasks) { task in
+///                     @Bindable var task = task
+///                     TextField("Task", text: $task.title)
+///                 }
+///             }
+///         }
+///     }
+///
+/// The local `@Bindable` value supplies the binding that keeps ``TextField``
+/// synchronized with the corresponding model property.
+///
+/// The same pattern works for objects obtained from the environment. Fetch the
+/// model first, then create a local `@Bindable` wrapper and pass its projected
+/// bindings where needed.
+///
+///     struct AccountNameView: View {
+///         @Environment(Account.self) private var account
+///
+///         var body: some View {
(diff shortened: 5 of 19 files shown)


JoshBashed:
Does this support `OpenCombine`?

s-k:
@JoshBashed This PR tries to add compatibility with `@Observable` similar to how SwiftUI supports it. AFAIK, OpenCombine works very differently compared to the Observation framework. I'm not sure what kind of support you expect.

Examples/Sources/ObservableExample/ObservableApp.swift
```diff
@@ -0,0 +1,77 @@
...
+                Text(model.windowText)
+                HStack {
+                    View1(model: model)
+                        .padding()
+                    View2(model: model)
+                        .padding()
+                }
+                ModifyingView(model: model)
+                    .padding()
+                if !model.automaticModeIsOn {
```
stackotter: Can you please make it possible to disable automatic mode again without restarting the app? Will need to store a reference to the automatic task, and then can probably cancel it from an 'onChange(of: automaticModeIsOn)' handler

Sources/SwiftCrossUI/State/Bindable.swift
```diff
@@ -0,0 +1,138 @@
+import Foundation
+
+/// A property wrapper that exposes bindings to writable properties on a
+/// mutable model object.
+///
+/// Use `Bindable` when you want controls to edit data stored in an object that
+/// participates in observation, such as a type annotated with `@Observable` or
+/// `@Perceptible` from the
+/// [Perception](https://github.com/pointfreeco/swift-perception) package.
```
stackotter: If perception gets split into a separate PR, remember to update these doc comments

Sources/SwiftCrossUI/State/Bindable.swift
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    /// Creates a bindable wrapper from another bindable wrapper's value.
+    public init(projectedValue: Bindable<Value>) {
+        self.init(wrappedValue: projectedValue.wrappedValue)
+    }
+}
+
+extension Bindable : Identifiable where Value : Identifiable {
+
```
stackotter: Remove this blank line

Package.swift
```diff
@@ -159,6 +159,10 @@ let package = Package(
             url: "https://github.com/swhitty/swift-mutex",
             .upToNextMinor(from: "0.0.6")
         ),
+        .package(
+            url: "https://github.com/pointfreeco/swift-perception.git",
+            from: "2.0.10"
+        ),
```
stackotter, line 182: I believe that swift-perception support should be behind a trait, as it won't be our recommended solution and is more of a convenience integration for people who want to use swift-perception anyway. Our Swift tools version is currently too low for traits, but it's getting to a point where we should probably consider bumping our tools version (there have been quite a few good reasons lately).

Would you feel comfortable splitting the swift-perception support into a separate PR so we can merge Observable support without having to worry about the Swift tools version bump?

Sources/SwiftCrossUI/State/ViewModelObserver.swift
```diff
@@ -0,0 +1,66 @@
...
+/// This protocol can be adoopted by classes responsible for handling part of the view hierarchy. It makes
+/// it easy to automatically update the view when observable view models change.
+///
+/// The most important rule: Every computation of a view's `body` MUST be performed inside a call to
+/// `observe()`. For example:
+///
+///     let body = self.observe(in: backend) { view.body }
+///     // Use `body`
+///
+/// Then, `viewModelDidChange()` will automatically be called the next time a view model conforming
```
stackotter:
```
/// Then, ``viewModelDidChange(backend:)`` ...
```

(that should link the docs to the method correctly)

Sources/SwiftCrossUI/State/ViewModelObserver.swift
```diff
@@ -0,0 +1,66 @@
+import Foundation
+import PerceptionCore
+
+/// This protocol can be adoopted by classes responsible for handling part of the view hierarchy. It makes
+/// it easy to automatically update the view when observable view models change.
+///
+/// The most important rule: Every computation of a view's `body` MUST be performed inside a call to
+/// `observe()`. For example:
```
stackotter:
```
/// ``observe(with:_:)`` ...
```

Sources/SwiftCrossUI/State/ViewModelObserver.swift
```diff
@@ -0,0 +1,66 @@
...
+    /// multiply.
+    ///
+    /// - Parameters:
+    ///   - backend: The backend used to schedule calls to `viewModelDidChange()` on the
+    ///   main thread. A strong reference will be held until that call has been made.
+    ///   - computation: The computation to be tracked. Usually, accesses to a view's `body`
+    ///   property will be encapsulated in this closure.
+    /// - Returns: The result of the computation.
+    func observe<Backend: AppBackend, Result>(
+        in backend: Backend,
```
stackotter: I think this label would make more sense as `with` (rather than `in`), because you're observing it *with* the backend as additional context.

Sources/SwiftCrossUI/State/ViewModelObserver.swift
```diff
@@ -0,0 +1,66 @@
...
+    /// `Observable` or `Perceptible` inside the computation. The next time one of those
+    /// properties changes, `viewModelDidChange()` will be called.
+    ///
+    /// If this method is called multiple times, only the last call will be tracked. The reason is that view
+    /// updates may be caused by other triggers. If all of those would be tracked, view updates would
+    /// multiply.
+    ///
+    /// - Parameters:
+    ///   - backend: The backend used to schedule calls to `viewModelDidChange()` on the
+    ///   main thread. A strong reference will be held until that call has been made.
```
stackotter: Indent hanging lines an extra two spaces (so that they align with the beginning of the bullet point content rather than the bullet itself)

Sources/SwiftCrossUI/State/ViewModelObserver.swift
```diff
@@ -0,0 +1,66 @@
...
+    /// When this method has been called, it will not be called again until the next call to `observe()`.
+    ///
+    /// - Parameter backend: The backend passed to the last call to `observe()`.
+    func viewModelDidChange<Backend: AppBackend>(backend: Backend)
+}
+
+extension ViewModelObserver {
+    /// Performs a computation and tracks accesses to properties of objects conforming to
+    /// `Observable` or `Perceptible` inside the computation. The next time one of those
+    /// properties changes, `viewModelDidChange()` will be called.
```
stackotter:
```
/// ... ``viewModelDidChange(backend:)`` ...
```

Sources/SwiftCrossUI/State/ModelObserver.swift
```diff
@@ -0,0 +1,66 @@
...
+    ///   - computation: The computation to be tracked. Usually, accesses to a view's `body`
+    ///   property will be encapsulated in this closure.
+    /// - Returns: The result of the computation.
+    func observe<Backend: AppBackend, Result>(
+        in backend: Backend,
+        _ computation: () -> Result
+    ) -> Result {
+        let perceptionTrackingID = UUID()
+        self.currentViewModelObservationID = perceptionTrackingID
+        return withPerceptionTracking {
```
stackotter: I assume `withPerceptionTracking` also supports observing `@Observable` changes given that you haven't referred to `withObservationTracking`?

Sources/SwiftCrossUI/State/ModelObserver.swift
```diff
@@ -0,0 +1,66 @@
...
+            computation()
+        } onChange: { [backend, weak self] in
+            backend.runInMainThread {
+                guard
+                    self?.currentViewModelObservationID == perceptionTrackingID
+                else { return }
```
stackotter, line 63: We're avoiding duplicate calls to the update handler, but how many calls are actually still happening? Is there a way for use to properly cancel previous observations so that we don't end up blocking `n` unnecessary updates after `n` body computations?

Sources/SwiftCrossUI/State/ModelObserver.swift
```diff
@@ -0,0 +1,66 @@
...
+///
+/// - Important: `self` MUST only be used to observe a single view because
+/// `viewModelDidChange()`will only be called for the most recent call to `observe()` in order to
+/// prevent duplicate view updates.
+@MainActor
+protocol ViewModelObserver: AnyObject, Sendable {
+    /// Used by the `ViewModelObserver` protocol to prevent duplicate view updates.
+    var currentViewModelObservationID: UUID? { get set }
+    
+    /// This method is called at most once after a call to `observe()` if an object conforming to
```
stackotter, line 24: Link to the observe method properly using a docc symbol link like I've demonstrated in earlier comments

Sources/SwiftCrossUI/State/ViewModelObserver.swift
```diff
@@ -0,0 +1,66 @@
...
+/// `observe()`. For example:
+///
+///     let body = self.observe(in: backend) { view.body }
+///     // Use `body`
+///
+/// Then, `viewModelDidChange()` will automatically be called the next time a view model conforming
+/// to `Observable` or `Perceptible` and used inside the `body` computation will have change.
+///
+/// - Important: `self` MUST only be used to observe a single view because
+/// `viewModelDidChange()`will only be called for the most recent call to `observe()` in order to
```
stackotter:
```
/// ``viewModelDidChange(backend:)`` will only be called for the the most recent
/// call to ``observe(with:_:)`` to prevent duplicate view updates.
```

Sources/SwiftCrossUI/State/ViewModelObserver.swift
```diff
@@ -0,0 +1,66 @@
+import Foundation
+import PerceptionCore
+
+/// This protocol can be adoopted by classes responsible for handling part of the view hierarchy. It makes
+/// it easy to automatically update the view when observable view models change.
```
stackotter: This protocol is also used to observe 'App' and will probably be used to observe the scene graph once it supports state properly, so we should probably name this protocol something more generic (such as `ModelObserver`?) and update the language in the doc comments to use terms such as 'stateful hierarchy' instead of 'view hierarchy'? The doc comments don't have to be perfect given that this is just an internal implementation detail, so don't worry too much about that side of things

Sources/SwiftCrossUI/ViewGraph/ViewGraphNode.swift
```diff
@@ -59,6 +60,9 @@ public class ViewGraphNode<NodeView: View, Backend: AppBackend>: Sendable {
 
     /// The dynamic property updater for this view.
     private var dynamicPropertyUpdater: DynamicPropertyUpdater<NodeView>
+    
+    /// Used by the `ViewModelObserver` protocol to prevent duplicate view updates.
```
stackotter: Use double backticks to create a DocC symbol link

Sources/SwiftCrossUI/State/State.swift
```diff
@@ -66,6 +67,23 @@ extension State {
...
     }
+    
+    // NB: Needed to prevent deprecation warnings for `Observable` types, which
+    // *are* supported
+    @available(macOS 14.0, iOS 17.0, watchOS 10.0, tvOS 17.0, *)
+    public init(wrappedValue initialValue: Value) where Value: Observable & AnyObject {
+        implementation = StateImpl(initialStorage: Storage(initialValue))
+    }
+    
+#if os(macOS) || os(iOS) || os(tvOS) || os(watchOS) || os(visionOS)
```
stackotter: Just as a formatting thing we generally indent compile time conditionals to match surrounding code (with the body indented an additional level)

stackotter: Given that this is an 'Apple platforms' issue, you could consider using `#if canImport(Darwin)` to be resilient against future Apple platforms being introduced (such as would've happened to this same code a couple of years ago when visionOS was introduced)

Sources/SwiftCrossUI/State/ViewModelObserver.swift
```diff
@@ -0,0 +1,66 @@
+import Foundation
+import PerceptionCore
+
+/// This protocol can be adoopted by classes responsible for handling part of the view hierarchy. It makes
```
stackotter: `adopted` (typo)

Sources/SwiftCrossUI/State/Bindable.swift
```diff
@@ -0,0 +1,138 @@
...
+    /// Creates a bindable wrapper around an observable or perceptible object.
+    ///
+    /// In most cases, apply the `@Bindable` attribute to a property or local
+    /// variable instead of calling this initializer directly.
+    public init(wrappedValue: Value) {
+        self.wrappedValue = wrappedValue
+    }
+}
+
+extension Bindable where Value : AnyObject {
```
stackotter: We only put a space after colons (and not before them). I think you've done this in a couple of other places so just double check the rest of the PR as well (can probably just command+f for ` : `)

stackotter (commented):
Thank you for this PR 🙏 I particularly like the ViewModelObserver protocol that you've used to make observation more succinct.

I've left a bunch of comments requesting a variety of changes, the biggest being potentially splitting perceptible support into a separate PR so that I can merge Observable support faster.

There was one other thing that I didn't comment on with a review comment, because I didn't know where to attach it. I would like for us to be vend an `@Observable` implementation (borrowed from Swift itself) to support `@Observable` as our official observation solution across all OS versions supported by SwiftCrossUI. My guess is that that would involve a bunch of copy+pasting (with care taken to give correct attribution and separate it from our own code a bit), and fixing up the implementations to not require newer language features. Then if you do that we should `@_reexport import Observation` on platforms that do have `Observation` so that people don't have to conditionally import `Observation` themselves based on target platform/platform version. Let me know if any of that doesn't make sense or isn't feasible!

Thanks for the detailed documentation by the way!

s-k:
@stackotter Thank you for the kind words! And thanks for the detailed response. I'm happy to implement the changes. There's one thing I'm not 100% clear on and I want to clear that up before starting: You suggest removing `Perceptible` support. Do you only refer to the signatures I added that mention `Perceptible` or do you want me to remove the dependency on swift-perception as well? It seems that you intend the latter. If that's the case, then we might have a problem. Here's the thing:

swift-perception does all the heavy lifting behind the scenes: It allows `Observable`-style observation on older OSs where the Observation framework is not available. It also automatically switches to using the Observation framework where it is available. This means that we can't simply drop the swift-perception dependency or hide it behind a trait. I believe we have four options:

1. Copy a bunch of code from the Observation framework (and/or) swift-perception into the project and use it only if Observation isn't available
2. Only support Observation and only where it is available
3. Leave the implementation largely as is and optionally remove any mention of `@Perceptible` in the documentation
4. The same as Option 3 but add typealiases so that `@Observable`, etc. are also available on older OSs with the same names.

Option 1 would be by far the most complicated and maintenance-heavy option. And the only benefit over Option 4 that I can think of is that the project would have one fewer dependency. It would take a lot of work to implement and test and the added code would need maintenance.

Option 2 would probably be relatively straightforward. We would lose the dependency but also the ability to observe on older OSs.

Option 3 allows all users to use the observation functionality. If they use `@Perceptible`, they will automatically use `@Observable` under the hood where available. And if they use `@Perceptible` in a case where the app only supports newer OSs, they will get a warning to switch to `@Observable`.

Option 4 has the same benefits as Option 3 but would allow users to use `@Observable` on all OSs. We have to be cautious to add the typealiases in a way that they don't cause naming conflicts in some configurations. However, I think it could be done.

What do you think?

Sources/SwiftCrossUI/State/ModelObserver.swift
s-k: Yes. You are right. See my long comment to the PR.

Sources/SwiftCrossUI/State/ModelObserver.swift
s-k: For every call to `withPerceptionTracking()` there will be one call to the second closure if a model changes. Since the current version of SwiftCrossUI (before and after my changes) performs multiple (maybe 5) view updates whenever a model (e.g. in `@State`) changes, this many calls to the closure also occur.

AFAIK, there is no way to cancel observations. However, in case of duplicates, almost no code is executed. If this ever becomes a problem, we could also prevent the call to `backend.runInMainThread()` by adding a custom locking mechanism to currentViewModelObservationID. On thing to note: SwiftUI also uses `withObservationTracking` under the hood. So the missing option to cancel should not be a problem by itself.

stackotter:
Ah I didn't quite realise that swift-perception basically does exactly what I proposed with backporting the official Swift Observation implementation. I think option `4` is best. By default we should only expose `@Observable` (and `withObservationTracking`, etc), but if the user imports `Perceptible`, then those macros should work all the same (which should happen by default given that we'll be using Perceptible to implement things anyway).

Are there any `Observation` APIs that swift-perception is missing? I guess we'll just have to stay on top of new `Observation` feature proposals and make sure that swift-perception gets corresponding implementations before the features officially land in Swift?

s-k:
@stackotter Great, we're on the same page! I'll try to get to it next week.

I believe swift-perception supports all APIs from `Observation`, but I'll check. It even has additional ones. My impression is that the library is well-maintained. Also, I'm not sure it's crucial that swift-perception is at parity with all `Observation` APIs. The important thing is that `withPerceptionTracking()` triggers.

s-k:
Sorry for the delay. I’ve been sick and unexpectedly busy, so I haven’t had the capacity to work on this yet.

I still intend to update the PR, but I can’t give a reliable timeline at the moment. Thanks for your patience.

stackotter:
No worries at all!

s-k:
@stackotter It seems Option 4 is not as simple as I hoped. The macro `@Perceptible` explicitly references `Perception.Perceptible` in its expansion. This means that if we simply re-export `Perceptible` as `Observable`, users would still need to `import Perception` in their source files. There may be a way around it but it would probably be a brittle solution.

My question to you: Every OS since 2023 supports `@Observable`. How many users target older OSs? How many of those want to use observation? Wouldn't it be OK if those users need to use `@Perceptible`? It would definitely save us a lot of hassle. 😊

stackotter:
> My question to you: Every OS since 2023 supports `@Observable`. How many users target older OSs? How many of those want to use observation? Wouldn't it be OK if those users need to use `@Perceptible`? It would definitely save us a lot of hassle. 😊

My concerns are more from a library design perspective than an application development perspective. SwiftCrossUI states a minimum iOS version of iOS 13. If we encourage `Observable` as our official solution for state management, and then someone who has build their app on Android wants to try it out on iOS they'll get a nasty surprise when they get a bunch of availability warnings about a core SwiftCrossUI feature that they've made use of throughout their app. There wouldn't be any indication from the compiler that the solution is to rename all `Observable` annotations to `Perceptible`.

Would it help if I forked swift-perceptible as swift-observation-polyfill and modified it to become an `Observable` polyfill? Then we could simply re-export that and call it a day; I think?

s-k:
> Would it help if I forked swift-perceptible as swift-observation-polyfill and modified it to become an `Observable` polyfill? Then we could simply re-export that and call it a day; I think?

I've thought about it for some time. First of all, I was able to work around the `import` problem partially. I have a solution for `@Observable` (not yet for `@ObservationIgnored`) that compiles and appears to work as expected:

```swift
@_exported import Perception

@available(macOS, obsoleted: 14, message: "Add `import Observation`.")
@available(iOS, obsoleted: 17, message: "Add `import Observation`.")
@available(watchOS, obsoleted: 10, message: "Add `import Observation`.")
@available(tvOS, obsoleted: 17, message: "Add `import Observation`.")
@attached(member, names: named(_$perceptionRegistrar), named(access), named(withMutation), named(shouldNotifyObservers))
@attached(memberAttribute)
@attached(extension, conformances: Perceptible, Observable)
public macro Observable() =
    #externalMacro(
        module: "PerceptionMacros",
        type: "PerceptibleMacro"
    )
```

However, while it compiles, it generates this warning:
> warning: external macro implementation type 'PerceptionMacros.PerceptibleMacro' could not be found for macro 'Observable()'; plugin for module 'PerceptionMacros' not found

I assume that you would not be OK with shipping such a solution. So there is probably no clean way around owning/forking the macro implementation. Your proposal seems feasible to me but there may be stumbling blocks on the way. Also, there are some things we need to keep in mind:

1. `swift-perception` references some types from the `Observation` framework such as `Observable` and `ObservationRegistrar`. We may need to qualify them using the `Observation.` prefix before introducing these names ourselves.
2. For some reason I don't know, the `Perception` macros and types are only deprecated in OS versions 26:
```swift
@available(iOS, deprecated: 26, renamed: "ObservationTracked")
@available(macOS, deprecated: 26, renamed: "ObservationTracked")
@available(watchOS, deprecated: 26, renamed: "ObservationTracked")
@available(tvOS, deprecated: 26, renamed: "ObservationTracked")
```
We would probably need to change that to:
```swift
@available(macOS, obsoleted: 14, message: "Add `import Observation`.")
@available(iOS, obsoleted: 17, message: "Add `import Observation`.")
@available(watchOS, obsoleted: 10, message: "Add `import Observation`.")
@available(tvOS, obsoleted: 17, message: "Add `import Observation`.")
```

Do I understand correctly that you would like to take care of the fork/polyfill part, and that I should continue with the SwiftCrossUI integration once that is in place?

stackotter:
Thank you for looking into that exported import workaround, it's a pity that it didn't work. And yep, I'm happy to take care of the observable polyfill next week

stackotter:
I've got a basic polyfill going now. I've tested it in a macOS 12 VM to make sure it functions correctly on older macOS versions, and it does seem to be working there.

The polyfill lives at https://github.com/moreSwift/swift-observation-polyfill

We can regenerate the polyfill for newer versions of swift-perception by updating the submodule and running `make_polyfill.py` again.

For completeness, I believe that we should also support `Observation.Observable` when it's available, because that will aid people who incrementally port their SwiftUI apps or rely on models from external packages that they don't have control over. That would probably involve adding a couple more `@available`-gated initialisers to `@State` iiuc.

Let me know if you run into any issues! Hopefully this polyfill is all we need to get this finished 🤞

stackotter:
If that polyfill works, then I'll tag a release for you to depend on before I merge this PR

s-k:
@stackotter Thanks! I'll look into it 👍

s-k:
@stackotter I tried out your polyfill. Here's what I found out:

1. I think we should rename `withObservationTracking` because it clashes with the original implementation if both `Observation` and `SwiftCrossUI` are imported. Users of SwiftCrossUI are unlikely to use that function.

2. I found a curious bug with `@ObservationIgnored`. A simple app with the following code:
```swift
import ObservationPolyfill

@Observable
class AClass {
    @ObservationIgnored var aVariable = 42
}
```
generates two errors:
a) Expansion of macro 'ObservationIgnored()' produced an unexpected 'init' accessor
b) Expansion of macro 'ObservationIgnored()' produced an unexpected getter

I believe I also got the error in my tests before you created the polyfill when I tried to `typealias` `PerceptionIgnored` as `ObservationIgnored`. Maybe the name throws the compiler off? I'm not sure how (or if) we can fix this.

3. When calling `withObservationTracking()` on macOS 26.5.1, I got an infinite loop. I have opened a PR with a fix: https://github.com/moreSwift/swift-observation-polyfill/pull/1

4. I confirmed that we need to change the `@available` attributes to prevent name clashes, as I assumed [in a previous comment](https://github.com/moreSwift/swift-cross-ui/pull/526#issuecomment-4601277775). Specifically, at least the ones in Macros.swift and Observable.swift in the polyfill need to be changed. Otherwise, compiler errors occur if the app using SwiftCrossUI has a minimum requirement of macOS 14 or newer (or the corresponding version of the other apple platforms).

I committed a version to this PR that currently does not compile. But if you intend to make changes to swift-observation-polyfill, you can possibly use it to check if these changes work.

s-k:
@stackotter Just wanted to follow up and see whether you’ve had a chance to take a look at the issues I mentioned above.

stackotter:
Thanks for reminding me about this. I've merged in your fix for issue 3 now, and merged in my own fix for issue 2.

Issue 1 seems like more of a quality life issue to me, so I'll focus on issue 4 for now.

I'm honestly not quite sure what to do about issue 4. One approach could be not re-export `Observation` and instead just always use our polyfilled implementations?

s-k:
> I'm honestly not quite sure what to do about issue 4. One approach could be not re-export `Observation` and instead just always use our polyfilled implementations?

I'm not a 100% sure what you mean. Do you propose that users of SwiftCrossUI would need to manually import the polyfill?

stackotter:
Nah, I mean that the polyfill could potentially stop re-exporting `Observation` so that if someone only imports SwiftCrossUI, they only have the polyfill implementations in scope (avoiding the clash). If they import `Observation` and `SwiftCrossUI` at the same time they'd have to disambiguate which implementation they want to use, but imo that's acceptable.

We can probably bump to Swift 6.1 (to get access to traits) and then add a trait that allows people to exclude the polyfill entirely (for devs whose target platform version has `Observation` support).

We'd also have to remove the deprecation warnings on the polyfill implementations, or replace the existing deprecation warnings with ones that instruct people to disable the default `ObservationPolyfill` trait (or whatever we call it).

How does that sound?

s-k:
I have two problems with that solution:

1. The official `Observation` macros generate code that refers to some types from the Observation framework without qualifying them with `Observation.`. This means that the name clashes can occur even if the user writes `Observation.Observable`.

2. I prefer simple solutions and just changing the `@available` attributes seems simpler than bumping to Swift 6.1, adding a new trait plus code for the trait and removing the deprecation warnings. Either way, we would have to change the `@available` attributes.

stackotter:
Oh, we had a misunderstanding. Do you mean that changing the availability annotations in Macros.swift and Observable.swift fixes the issue? For some reason I read your message as meaning that we had to rename something to make things work 😅 Of course I agree with you that changing the availability annotations is the best solution, I thought you were saying we needed something else for some reason.

That's very strange that the availability annotations are sitting at 26. Not sure why they would've done that. I'll try lowering those.

stackotter:
Modifying the `@available` annotations to deprecate the polyfills when the `Observable` macro is available doesn't appear to fix the issue. It still says that `ObservationIgnored` is an ambiguous reference.

s-k:
Yes, just the `@available` annotations. Did you change `deprecated` to `obsoleted`?

I think something like this should do the trick as mentioned above:

> ```swift
> @available(iOS, deprecated: 26, renamed: "ObservationTracked")
> @available(macOS, deprecated: 26, renamed: "ObservationTracked")
> @available(watchOS, deprecated: 26, renamed: "ObservationTracked")
> @available(tvOS, deprecated: 26, renamed: "ObservationTracked")
> ```
> 
> We would probably need to change that to:
> 
> ```swift
> @available(macOS, obsoleted: 14, message: "Add `import Observation`.")
> @available(iOS, obsoleted: 17, message: "Add `import Observation`.")
> @available(watchOS, obsoleted: 10, message: "Add `import Observation`.")
> @available(tvOS, obsoleted: 17, message: "Add `import Observation`.")
> ```

stackotter:
That worked! The part that I was missing was `obsoleted` 🤦‍♂️ That's a very handy `@available` parameter to know about; I've Googled around for a way to achieve that effect many times and never discovered that parameter somehow

stackotter:
I've just merged that fix into the polyfill repo. Everything appears to be working on my end now with both macOS 13 and macOS 14 deployment targets.

For assurance purposes, would you be able to add the following code below your `automaticModeTask` property?

```swift
// Namespaced spellings of the ObservationIgnored and ObservationTracked
// macros are also supported when your deployment target is below that of
// the Observation module
@ObservationPolyfill.ObservationIgnored private var ignored1 = false
@SwiftCrossUI.ObservationIgnored private var ignored2 = false
```

Alternatively (or in addition to that), you could add something similar to one of the models in your unit tests, so that the tests would stop compiling if we broke something (not super elegant, but it probably works).

Examples/Sources/ObservableExample/ObservableModel.swift
```diff
@@ -0,0 +1,66 @@
...
+                switch textIndex {
+                case 0:
+                    windowTitle = animal
+                case 1:
+                    windowText = animal
+                case 2:
+                    view1Text = animal
+                default:
+                    view2Text = animal
+                }
```
stackotter, line 61: Indent these cases one level

Examples/Bundler.toml
```diff
@@ -109,3 +109,8 @@ version = '0.1.0'
 identifier = 'dev.swiftcrossui.TapGesturesExample'
 product = 'TapGesturesExample'
 version = '0.1.0'
+
+[apps.ObservableExample]
+identifier = 'dev.swiftcrossui.ObservableExample'
```
stackotter, line 114: Add this example to the `build-test-and-docs.yml` GitHub Actions workflow as well (to the examples section of each platform's build job)

Sources/SwiftCrossUI/State/ModelObserver.swift
```diff
@@ -0,0 +1,69 @@
...
+///
+/// The most important rule: Every computation of a `body` property MUST be performed inside a call to
+/// ``observe(with:_:)``. For example:
+///
+///     let body = self.observe(in: backend) { view.body }
+///     // Use `body`
+///
+/// Then, ``viewModelDidChange(backend:)`` will automatically be called the next time a view
+/// model conforming to `Observable` or `Perceptible` and used inside the `body` computation
+/// will have changed.
```
stackotter: Change `will have changed` to `changes`

Sources/SwiftCrossUI/State/State.swift
```diff
@@ -55,7 +56,7 @@ extension State {
         deprecated,
         message: """
             'State' does not work correctly with non-observable classes; conform \
-            your class to 'ObservableObject' or use a struct instead
+            your class to 'ObservableObject', use `@Observable` or use a struct instead
```
stackotter: Add a comma after `@Observable`

Sources/SwiftCrossUI/ViewGraph/ViewGraphNode.swift
```diff
@@ -230,18 +238,19 @@ public class ViewGraphNode<NodeView: View, Backend: BaseAppBackend>: Sendable {
...
         // view update. Under that assumption, we can cache view layout results.
+        let result = self.observe(with: backend) {
+            view.computeLayout(
+                widget,
+                children: children,
+                proposedSize: proposedSize,
+                environment: viewEnvironment,
+                backend: backend
+            )
+        }
```
stackotter, line 247: Move this back above the comment (because the comment is for the `resultCache[proposedSize] = result` line)

Package.swift
```diff
@@ -176,6 +176,10 @@ let package = Package(
             url: "https://github.com/swhitty/swift-mutex",
             .upToNextMinor(from: "0.0.6")
         ),
+        .package(
+            url: "https://github.com/moreSwift/swift-observation-polyfill",
+            branch: "main"
```
stackotter: Reminder to self: release the polyfill and update this to a versioned dependency (once we've confirmed that everything is working)

Sources/SwiftCrossUI/ViewGraph/ViewGraphNode.swift
```diff
@@ -277,13 +286,15 @@ public class ViewGraphNode<NodeView: View, Backend: BaseAppBackend>: Sendable {
...
-        )
+        self.observe(with: backend) {
+            view.commit(
+                widget,
+                children: children,
+                layout: currentLayout,
+                environment: parentEnvironment,
+                backend: backend
+            )
+        }
```
stackotter: I don't think that we should observe `commit`, because it's semantically just committing the result of the last `computeLayout` call, and shouldn't be reading any new state. At the moment this probably works fine in most cases because we're not very careful about unnecessary body computations, but once we start optimising that stuff this may start breaking in strange ways. I think only observing `computeLayout` should provide all of the information that we need.

stackotter (changes requested):
I've gone through all of your current changes (even though they're not complete yet) to give something for you to work off, and to reduce the work for future me when this PR is ready for a proper review.

Sources/SwiftCrossUI/ViewGraph/ViewGraphNode.swift
s-k: I believe I added this observation recently because in some circumstances, some views would not update automatically. Specifically, after ending the automatic mode in the example and then typing manually into the `TextField`s would not update certain texts. I was able to fix it with this observation. I'm not sure if I can find another spot for the observation that also fixes the issue.

s-k:
Great! I have also compiled and run successfully on my end. I can add the `ObservationIgnored` code to the example. However, it breaks as soon as the deployment target of the Examples package is updated to macOS 14 or higher. Shall I do it anyway?

I have not added any unit tests to the PR because I wasn't sure how I would set up a test that actually verifies that views would actually update in the real world. I'm also not sure if I can pull it off because I will have very little time in the coming weeks.

stackotter:
> However, it breaks as soon as the deployment target of the Examples package is updated to macOS 14 or higher. Shall I do it anyway?

Yeah, I think that's fine given that the comment mentions the caveat. If we ever bump the version we can just remove those.

> I have not added any unit tests to the PR because I wasn't sure how I would set up a test that actually verifies that views would actually update in the real world. I'm also not sure if I can pull it off because I will have very little time in the coming weeks.

Yeah no worries! I don't really expect any unit tests given that we don't have many to begin with. Someone will probably add some eventually given that state updates are on the more testable side of things that SwiftCrossUI deals with, or I will if I find issues with the existing implementation and fix them.

Sources/SwiftCrossUI/ViewGraph/ViewGraphNode.swift
s-k: I have tested again without this `observe()`. Somehow, it works totally fine without it. Maybe some other change fixed the behavior. I have removed the `observe()`.

Examples/Bundler.toml
s-k: I have added the example. Please check if I've done it correctly.

s-k:
I've addressed your comments. On my end, everything seems to work.

Package.swift
stackotter: Update this to `.upToNextMinor(from: "0.1.0")` now that I've made a release

stackotter (changes requested):
Thank you for your patience on this PR! It looks great now. My final requested change is to update the polyfill dependency to 0.1.0 now that you've helped me verify that everything works and I've made a release.

s-k:
Sorry for not answering sooner. I'm currently on vacation. I've made the requested change. Thanks again for all your effort!

stackotter (dismissed):
Thanks again for your patience on this PR. I'm excited that we can finally get this merged into SCUI!

I'll merge it once CI finishes

s-k:
It seems that the CI found compilation issues that seem a bit cryptic to me. I currently don't have a reliable Internet connection, which means that I won't be able to look at those in the near future. At first glance it looks like the problems occur in the macro expansion, so I don't know whether the problem lies within SwiftCrossUI or the polyfill.

@stackotter Do you happen to have any idea what might be causing the problems?

MiaKoring:
At least one of the issues is that `extension Foo: nonisolated Bar` is not valid afaik

(Also on mobile rn so reading logs isn't too fun xD)

stackotter:
The cause of this issue was kind of amusing; https://github.com/moreSwift/swift-observation-polyfill/pull/4

You should be able to bump your branch's polyfill dependency to `0.1.1` to get my fix

s-k:
@stackotter Thanks for fixing the issue. I have updated to version 0.1.1. The Example did compile and run on my end. However, I was not able to do in-depth testing because of my unreliable internet connection.

stackotter (dismissed):

stackotter:
The example seems to compile fine on my end; I guess we'll whether anything is broken once the CI finishes

stackotter:
Dang, the CI still fails (although some do pass now).

visionOS should be an easy fix; just remove the namespaced `ObservationIgnored` invocations that I told you to add (perhaps just leave them commented out for illustrative purposes). The issue is that visionOS 1.0 supports Observation (unlike our other minimum deployment targets), so Observation is unconditionally available and the namespaced spellings of ObservationIgnored are therefore invalid when building for visionOS.

The Linux and Android situation isn't as simple. The issue probably affects Windows as well, but we won't know without testing locally because the Windows CI doesn't build the examples (due to some weird SwiftPM bug we faced at some point).

I have to leave home now, but I'll look into the issue for non-Apple platforms when I get the chance.

s-k:
I believe the problem with Linux (and Windows) is that the `Observation` framework is always available. Therefore, we have to prevent the polyfill types to be loaded on these platforms. Either, we do that inside the polyfill or we only `@_exported import ObservationPolyfill` on platforms except for Linux, Windows and visionOS.

s-k:
I finally was able to work on the PR again. I have tried to fix the problems by only including `@_exported import ObservationPolyfill` on Apple platforms except visionOS. I have successfully compiled the example on macOS 26 and Linux (with Docker).

stackotter:
Thanks for pushing forward on this PR and fixing that issue! It appears that the only failing CI now is the formatting checker, which should be an easy fix. Just install a recent enough version of `swiftformat` and run `./Scripts/format.sh` to automatically fix the formatting issues (hopefully)

s-k:
Thanks for the quick response! The formatting should be fixed now.

stackotter (approved):

stackotter:
I'm so excited to have this finally merged into main, thanks for all of your work on this @s-k!

PR #531, opened 2026-04-29
CI: Use custom Swift DocC build to support WebP images

stackotter:
This updates the CI to compile documentation with my custom fork of swift-docc that supports WebP.


stackotter:
Once merged this should unblock #527 (which I said should use WebP images, not realising that vanilla DocC doesn't support WebP).

PR #532, opened 2026-05-01
Android picker support

bbrk24:
Gets the ControlsExample running on Android, and adds support for menu, radio button, and wheel-style pickers. Segmented picker support is possible in the future if we can import Material.

### DIFF
--- a/Examples/Sources/ControlsExample/ControlsApp.swift
+++ b/Examples/Sources/ControlsExample/ControlsApp.swift
@@ -21,6 +21,11 @@ enum BuiltInPickerStyle: CaseIterable, Equatable {
     }
 }
 
+#if canImport(AndroidBackend)
+// TODO(bbrk24): Update this once AndroidBackend supports scrolling
+typealias ScrollView = VStack
+#endif
+
 @main
 @HotReloadable
 struct ControlsApp: App {
@@ -56,6 +61,7 @@ struct ControlsApp: App {
                             Text("Count: \(count)")
                         }
 
+                        #if !canImport(AndroidBackend)
                         VStack {
                             Text("Menu button")
                             Menu("Menu") {
@@ -125,6 +131,7 @@ struct ControlsApp: App {
                                     .frame(width: progressViewSize, height: progressViewSize)
                             }
                         #endif
+                        #endif
 
                         #if !canImport(Gtk3Backend)
                             VStack {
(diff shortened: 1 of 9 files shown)


Examples/Sources/ControlsExample/ControlsApp.swift
```diff
@@ -21,6 +21,11 @@ enum BuiltInPickerStyle: CaseIterable, Equatable {
     }
 }
 
+#if canImport(AndroidBackend)
```
stackotter, line 24: Indent the bodies of the `#if` blocks one level

Sources/AndroidBackend/AndroidBackend.swift
```diff
@@ -77,137 +77,136 @@ extension App {
...
-
+    //    public typealias Menu = Never
+    //    public typealias Alert = Never
+    //    public typealias Path = Never
+    //    public typealias Sheet = Never
+    
     static let stdoutPipe = Pipe()
     static let stderrPipe = Pipe()
-
+
```
stackotter: Looks like some trailing whitespace crept in

Sources/AndroidBackend/AndroidBackend.swift
```diff
@@ -395,3 +394,86 @@ public final class AndroidBackend: BackendFeatures.BaseStubs {
...
+    
+// MARK: Picker
+extension AndroidBackend: BackendFeatures.Pickers {
+    public var supportedPickerStyles: [BackendPickerStyle] {
+        [.menu, .radioGroup, .wheel]
+    }
+
+    public func createPicker(style: BackendPickerStyle) -> Widget {
+        switch style {
+        case .radioGroup:
```
stackotter: Indent body of switch statement one more level

stackotter (changes requested):
Looks great, thanks for getting that working. My requested changes are purely formatting related

Sources/AndroidBackend/AndroidBackend.swift
bbrk24: Xcode automatically inserts this whenever I edit any part of this file, and the formatter doesn't remove it.

stackotter: Weird, specifically just this file?

bbrk24: Looks like I can manually delete it and Xcode doesn't put it back. That's annoying to have to do though.

stackotter (approved):
Thanks for that, looks good now

PR #536, opened 2026-05-01
Fix AppKit menu pickers

### LINKED ISSUE #535: AppKit Picker doesn’t work/crashes
Currently the pickerstyle picker in ControlsExample does nothing and the flavour selector leads to a crash for `index out of range` on AppKitBackend.

GtkBackend is not affected.

bbrk24:
Fixes #535 

This PR applies the behavior that other backends were already doing -- instead of throwing out and recreating the menu on every update, update the menu items in place, and only delete or add menu items when necessary.


stackotter (approved):
Looks great, thanks for fixing that. If I have any feedback for the AppBackend refactor then I'll merge this first, otherwise I'll merge the AppBackend refactor in order to get that out of the way (whichever PR gets merged second will need a bit of fixing)

PR #537, opened 2026-05-02
Implement WebView for Android

bbrk24:
Leaving this as draft until https://github.com/moreSwift/swift-bundler/issues/188 is addressed, because WebView requires `android.permission.INTERNET`.


stackotter (approved):
Looks great, will merge once CI passes.

PR #538, opened 2026-05-03
Implement toggles on Android

bbrk24:
Without XML styling, the default toggle button component shows no visual difference between the checked and unchecked states. I also removed the `defaultToggleStyle` property, which never actually existed.


stackotter (approved):
Looks great, thanks for this! I'll test locally then merge if all goes well

PR #542, opened 2026-05-04
Add alerts on Android

bbrk24:
This gives a few deprecation warnings because we're using the deprecated version of DialogFragment, but it appears to function mostly the same. The only thing I've noticed is that `setCancelable(false)` doesn't actually work.


Sources/AndroidBackend/AndroidBackend.swift
```diff
@@ -477,3 +476,39 @@ extension AndroidBackend: BackendFeatures.Pickers {
         }
     }
 }
+
+// MARK: Alert
```
stackotter: Remove the MARK (should be clear enough from the protocol clause)

stackotter (changes requested):
Just have one formatting related requested change, but other than that looks great 🙏

stackotter (approved):
Seems to work now after fixing up the automatic merge

PR #549, opened 2026-05-05
Cache window size

bbrk24:
Performance improvement. Primarily motivated by the fact that window measurement can take ~70ms on Android due to blocking on ioctls, but it should have a small performance boost on all platforms.


stackotter (approved):
Looks good; will just check on iOS before merging

Issue #545, opened 2026-05-04
AndroidBackend: On click handlers lag or wait until after animation completes


stackotter:
On click/on change handlers of most controls seem to be either lagging or waiting until after the state change animation completes.

I've seen the following message in logcat a few times (around when I clicked a control);

```
Choreographer: Skipped 58 frames!  The application may be doing too much work on its main thread.
```

Perhaps we are just being too slow at updating layouts? Whatever it is, it's likely Android-specific because we perform perfectly fine in these example apps on other platforms.

PR #550, opened 2026-05-05
AndroidBackend: Switch to custom container (RelativeLayout too slow)

stackotter:
RelativeLayout's onMeasure implementation has a branching factor of 2, leading to Android's measurement pass being exponential with regard to the depth of the SwiftCrossUI view hierarchy. That appears to be the cause of #545, which we found out thanks to @bbrk24's profiling earlier today.

This PR resolves the issue by switching over to a custom ViewGroup subclass for our container implementation. The new implementation is written in Kotlin because a) it's much more ergonomic interacting with Android APIs from Kotlin, and b) it reduces the number of JNI calls we have to make.

This is a draft because it seems to have broken a couple of views such as checkboxes. We may be assigning a size that is too small? I haven't looked into it yet.

### DIFF
--- a/Sources/AndroidBackend/Kotlin/CustomContainer.kt
+++ b/Sources/AndroidBackend/Kotlin/CustomContainer.kt
@@ -37,9 +37,6 @@ class CustomContainer(
     override protected fun onMeasure(widthMeasureSpec: Int, heightMeasureSpec: Int) {
         setMeasuredDimension(layoutParams.width, layoutParams.height)
 
-        val width = View.MeasureSpec.getSize(widthMeasureSpec)
-        val height = View.MeasureSpec.getSize(heightMeasureSpec)
-
         for (i in 0..<childCount) {
             val child = getChildAt(i)
             // If you're debugging this implementation and see a child with
--- a/Sources/AndroidBackend/Kotlin/CustomContainer.kt
+++ b/Sources/AndroidBackend/Kotlin/CustomContainer.kt
@@ -1,0 +1,76 @@
+package dev.swiftcrossui.androidbackend
+
+import android.app.Activity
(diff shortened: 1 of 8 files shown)


stackotter:
The issue with toggle buttons, switches, and checkboxes being invisible was just because those SwiftCrossUI views weren't setting the sizes of their underlying widgets. The AndroidBackend now requires that all widgets get explicitly assigned sizes, so that was causing the toggles to be rendered with a size of 0x0.

Sources/AndroidBackend/Kotlin/CustomContainer.kt
```diff
@@ -0,0 +1,76 @@
...
+        layoutParams: ViewGroup.LayoutParams?
+    ): ViewGroup.LayoutParams {
+        return LayoutParams(layoutParams!!.width, layoutParams!!.height, 0, 0)
+    }
+
+    override protected fun onMeasure(widthMeasureSpec: Int, heightMeasureSpec: Int) {
+        setMeasuredDimension(layoutParams.width, layoutParams.height)
+
+        val width = View.MeasureSpec.getSize(widthMeasureSpec)
+        val height = View.MeasureSpec.getSize(heightMeasureSpec)
```
bbrk24: Are these used?

bbrk24 (commented):
The hiccup on my segmented picker prototype went from over four seconds to under 8ms. I'd say that's a pretty good improvement

Sources/AndroidBackend/Kotlin/CustomContainer.kt
stackotter: Nope… good catch 😅 those were there from a log message I was creating in the past

PR #555, opened 2026-05-06
Add color support for Android

bbrk24:
Uses the default shades of yellow and brown because Android doesn't have system resources for those.

### DIFF
--- a/Sources/AndroidBackend/AndroidBackend.swift
+++ b/Sources/AndroidBackend/AndroidBackend.swift
@@ -647,45 +647,45 @@ extension AndroidBackend: BackendFeatures.ToggleButtons, BackendFeatures.Checkbo
     }
 }
 
-extension AndroidBackend: BackendFeatures.Colors {
-    public func createColorableRectangle() -> Widget {
-        AndroidKit.View(Self.activity, environment: Self.env)
-    }
-
-    private func toColorInt(_ color: SwiftCrossUI.Color.Resolved) -> Int32 {
-        // https://cs.android.com/android/platform/superproject/+/android-latest-release:frameworks/base/graphics/java/android/graphics/Color.java;drc=ba44b0b3545f60e2b82e864ea0e684ce7407349a;l=1338
-        let alpha = UInt32(color.opacity * 255.0 + 0.5)
-        let red = UInt32(color.red * 255.0 + 0.5)
-        let green = UInt32(color.green * 255.0 + 0.5)
-        let blue = UInt32(color.blue * 255.0 + 0.5)
+extension SwiftCrossUI.Color.Resolved {
+    func asColorInt() -> Int32 {
+        let alpha = UInt32(opacity * 255.0 + 0.5)
+        let red = UInt32(red * 255.0 + 0.5)
+        let green = UInt32(green * 255.0 + 0.5)
+        let blue = UInt32(blue * 255.0 + 0.5)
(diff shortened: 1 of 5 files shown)


Sources/AndroidBackend/AndroidBackend.swift
```diff
@@ -646,3 +646,76 @@ extension AndroidBackend: BackendFeatures.ToggleButtons, BackendFeatures.Checkbo
...
+    }
+    
+    private func toResolvedColor(_ int: Int32) -> SwiftCrossUI.Color.Resolved {
+        // inverse of the above
+        let uint = UInt32(bitPattern: int)
+
+        let alpha = (Float(uint >> 24) - 0.5) / 255.0
+        let red = (Float((uint & 0x00FF0000) >> 16) - 0.5) / 255.0
+        let green = (Float((uint & 0x0000FF00) >> 8) - 0.5) / 255.0
+        let blue = (Float(uint & 0x000000FF) - 0.5) / 255.0
```
stackotter: I believe that this is incorrect. A colour that comes into toColorInt with a `red` value of 0.0 would come out of toResolvedColor with a `red` value of -0.5. The `+ 0.5` in toColorInt appears to be a crude way of forcing rounding to the nearest integer. We can't invert rounding, so we should just drop the `0.5` in toResolvedColor entirely.

Sources/AndroidBackend/AndroidBackend.swift
```diff
@@ -646,3 +646,76 @@ extension AndroidBackend: BackendFeatures.ToggleButtons, BackendFeatures.Checkbo
...
+        let red = UInt32(color.red * 255.0 + 0.5)
+        let green = UInt32(color.green * 255.0 + 0.5)
+        let blue = UInt32(color.blue * 255.0 + 0.5)
+
+        let combined = (alpha << 24) | (red << 16) | (green << 8) | blue
+
+        return Int32(bitPattern: combined)
+    }
+    
+    private func toResolvedColor(_ int: Int32) -> SwiftCrossUI.Color.Resolved {
```
stackotter: As an instance method of AndroidBackend this naming comes across a bit unclear. toColorInt and toResolvedColor might be better placed in an extension of `SwiftCrossUI.Color.Resolved`. E.g. toColorInt would become `Color.Resolved.asARGBInt32`, and toResolvedColor would become `Color.Resolved.init(fromARGBInt32:)`

Sources/AndroidBackend/AndroidBackend.swift
```diff
@@ -646,3 +646,76 @@ extension AndroidBackend: BackendFeatures.ToggleButtons, BackendFeatures.Checkbo
...
+        // inverse of the above
+        let uint = UInt32(bitPattern: int)
+
+        let alpha = (Float(uint >> 24) - 0.5) / 255.0
+        let red = (Float((uint & 0x00FF0000) >> 16) - 0.5) / 255.0
+        let green = (Float((uint & 0x0000FF00) >> 8) - 0.5) / 255.0
+        let blue = (Float(uint & 0x000000FF) - 0.5) / 255.0
+
+        return .init(
```
stackotter: Use an explicit initializer

stackotter (changes requested):

stackotter (approved):
Looks great now, thanks for fixing those things up!

PR #558, opened 2026-05-07
Android text styling

bbrk24:
This PR adds a fonts example, implements text styling in AndroidBackend, and makes Android the first platform to use custom text styles (they actually grow and shrink in response to the system font size setting).

### DIFF
--- a/Sources/AndroidBackend/AndroidBackend+Fonts.swift
+++ b/Sources/AndroidBackend/AndroidBackend+Fonts.swift
@@ -7,6 +7,7 @@ extension AndroidKit.TextView {
     func setTypeface(_ tf: AndroidKit.Typeface?)
 }
 
+// swiftlint:disable force_try
 extension AndroidBackend {
     struct TextStyle {
         var color: Int32
--- a/Sources/AndroidBackend/AndroidBackend+Fonts.swift
+++ b/Sources/AndroidBackend/AndroidBackend+Fonts.swift
@@ -61,7 +61,11 @@ extension AndroidBackend {
         )
     }
 
-    public func resolveTextStyle(_ textStyle: SwiftCrossUI.Font.TextStyle) -> SwiftCrossUI.Font.TextStyle.Resolved {
+    public func resolveTextStyle(
+        _ textStyle: SwiftCrossUI.Font.TextStyle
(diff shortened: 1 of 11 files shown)


Sources/AndroidBackend/AndroidBackend+Fonts.swift
```diff
@@ -0,0 +1,88 @@
...
+        return TextStyle(
+            color: colorInt,
+            fontSizePixels: Float(resolvedFont.pointSize),
+            lineHeightPixels: Int32(resolvedFont.lineHeight),
+            typeface: typeface
+        )
+    }
+
+    public func resolveTextStyle(_ textStyle: SwiftCrossUI.Font.TextStyle) -> SwiftCrossUI.Font.TextStyle.Resolved {
+        let textSize =
```
stackotter, line 69: Add a comment to explain that the ratios were chosen to roughly match the distribution of iOS's text styles.

Sources/AndroidBackend/AndroidBackend+Fonts.swift
```diff
@@ -0,0 +1,88 @@
...
+                case .caption2: helpers.getSmallTextSize(Self.activity) * 0.846
+                case .footnote: helpers.getSmallTextSize(Self.activity)
+            }
+
+        // TODO(bbrk24): Update this once everything else uses dp
+        return SwiftCrossUI.Font.TextStyle.Resolved(
+            pointSize: Double(textSize),
+            weight: textStyle == .headline ? .semibold : .regular,
+            emphasizedWeight: .semibold,
+            lineHeight: Double(textSize) * 1.15
```
stackotter: Add a comment to explain that Android seems to default to `lineHeight == fontSize`, and this default value matches iOS closer

stackotter (changes requested):
Code looks great 🙏 Just figured we should document a couple of the arbitrary-ish decisions we discussed so that we can rediscover the reasoning in future if we ever want to change things

stackotter:
In my local testing, I found that these changes break the default font colour of text when using dark mode on Android. The best way to solve that is probably for us to properly detect and respect the system theme.

stackotter (approved):
Looks great, thanks! 🙏 Gonna test locally and then merge

PR #560, opened 2026-05-08
Android device class support

bbrk24:
Doesn't do much on its own currently.

### DIFF
--- a/Sources/AndroidBackend/AndroidBackend.swift
+++ b/Sources/AndroidBackend/AndroidBackend.swift
@@ -84,8 +84,15 @@ public final class AndroidBackend: BackendFeatures.BaseStubs {
     static let stdoutPipe = Pipe()
     static let stderrPipe = Pipe()
 
-    // TODO(stackotter): Dynamically determine the device class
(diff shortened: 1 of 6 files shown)


Sources/AndroidBackend/AndroidBackend.swift
```diff
@@ -84,8 +84,15 @@ public final class AndroidBackend: BackendFeatures.BaseStubs {
...
-    // TODO(stackotter): Dynamically determine the device class
-    public let deviceClass = DeviceClass.phone
+    public lazy var deviceClass: DeviceClass =
+        switch helpers.getDeviceClass(Self.activity) {
+            case 0: .desktop
+            case 1: .phone
+            case 2: .tablet
+            case 3: .tv
+            case 4: .watch
+            default: fatalError()
```
stackotter: Add a fatal error message to aid debugging

stackotter (changes requested):
Just requested one tiny change, but other than that looks great

stackotter (approved):

PR #564, opened 2026-05-08
Detect current time zone, locale, and calendar on Android

bbrk24:
Draft primarily because I still need to figure out how to refer to `Serializable` in the `implements` list for `AndroidCalendar`. I'll also need to come up with a better name for `ifNotEmpty`.

### DIFF
--- a/Sources/AndroidBackend/Kotlin/AndroidBackendHelpers.kt
+++ b/Sources/AndroidBackend/Kotlin/AndroidBackendHelpers.kt
@@ -3,7 +3,9 @@ package dev.swiftcrossui.androidbackend
 import android.R
 import android.app.Activity
 import android.content.res.Configuration
+import android.icu.util.TimeZone
 import android.os.Build
+import android.text.TextUtils
 import android.util.TypedValue
 import android.view.WindowInsets
 import android.widget.TextView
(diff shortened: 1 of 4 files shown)


stackotter (approved):
Looks great; just going to test things locally before merging

Sources/AndroidBackend/Kotlin/AndroidBackendHelpers.kt
```diff
@@ -133,4 +135,19 @@ class AndroidBackendHelpers {
...
     }
+
+    fun getTimeZoneIdentifier(): String? {
+        val tz = TimeZone.getDefault()
+
+        // getCanonicalID uses outdated information. In Android 16, instead of updating it, they
+        // added a whole new method that uses up-to-date information. The new method returns
+        // TimeZone.UNKNOWN_ZONE_ID on error; the old method returns null on error.
+        if (Build.VERSION.SDK_INT >= 36) {
+            val id = TimeZone.getIanaID(tz.getID())
```
stackotter, line 196: I get 'unresolved reference getIanaID' when attempting to compile this code locally 🤔 My SDK version appears to be 35. Is there a way for us to only compile that SDK checking code when the compiling SDK is high enough in the first place? Otherwise we always lock compilation to the highest SDK version that we conditionally support features from

stackotter (changes requested):

Sources/AndroidBackend/Kotlin/AndroidBackendHelpers.kt
bbrk24: I don't think so -- Kotlin doesn't have `#if` (nor does Java). One thing we might be able to do, though this would require swift-bundler support, is:
- Mark a function as `expect`
- Have an `actual` implementation in another file that always calls `getCanonicalID`
- Have an `actual` implementation in another file that checks the version number, like I am here
- Conditionally compile only one of those two files based on your compile SDK version

(Aside: I'm honestly not 100% sure how `expect`/`actual` works in Kotlin. That's just my first idea off the top of my head.)

bbrk24: Coming back to this now, the Gradle file that sbun generates for me has `compileSdk = 36`. Shouldn't that mean you need SDK 36 to build it anyways?

stackotter: It generates with the latest SDK that you have available (unless the project specifies a compileSDK)

stackotter (approved):

PR #566, opened 2026-05-10
Split up AndroidBackend

bbrk24:
Splits each `extension AndroidBackend` into its own file.


stackotter (approved):
Looks great, thanks for doing that, it'll definitely make navigating around a little easier (especially for those who don't live in the fuzzy finder).

PR #567, opened 2026-05-10
[AndroidBackend] Use DIP/SP for sizes instead of PX

bbrk24:


Sources/AndroidBackend/AndroidBackend+Paths.swift
```diff
@@ -109,44 +116,49 @@ extension AndroidBackend: BackendFeatures.Paths {
         }
     }
 
-    private func apply(actions: [SwiftCrossUI.Path.Action], to path: AndroidGraphics.Path) {
+    private func apply(
+        actions: [SwiftCrossUI.Path.Action],
+        environment: EnvironmentValues,
+        to path: AndroidGraphics.Path
+    ) {
+        let d = environment.windowScaleFactor
```
stackotter: Rename this to `density` for clarity

Sources/AndroidBackend/AndroidBackend.swift
```diff
@@ -320,22 +320,25 @@ public final class AndroidBackend: BackendFeatures.BaseStubs {
...
     public func setSize(of widget: Widget, to size: SIMD2<Int>) {
+        let density = widget.getResources().getDisplayMetrics().density
         let layoutParams = widget.getLayoutParams()!
-        layoutParams.width = Int32(size.x)
-        layoutParams.height = Int32(size.y)
+        layoutParams.width = Int32(Float(size.x) * density)
+        layoutParams.height = Int32(Float(size.y) * density)
         widget.setLayoutParams(layoutParams)
         
         // TODO(stackotter): Use density-adaptive units everywhere
```
stackotter: I think you've completed this?

Sources/AndroidBackend/AndroidBackend.swift
```diff
@@ -436,15 +439,20 @@ public final class AndroidBackend: BackendFeatures.BaseStubs {
         // 0x3FFFFFFF = View.MeasureSpec.makeMeasureSpec(Int32.max, View.MeasureSpec.UNSPECIFIED)
         let widthSpec =
             if let proposedWidth {
-                Int32(bitPattern: 0x80000000 | UInt32(proposedWidth))
+                Int32(bitPattern: 0x80000000 | UInt32(Double(proposedWidth) * environment.windowScaleFactor) & ~0x40000000)
```
stackotter, line 440: Should we consistently round up? I think it'd be a good idea given that the density can be fractional

stackotter (changes requested):
Just requested a couple of small changes. Looks great for the most part!

Sources/AndroidBackend/AndroidBackend.swift
bbrk24: Whoops 😅

Sources/AndroidBackend/AndroidBackend+Paths.swift
bbrk24: I wanted to give it a shorter name because I refer to it so many times, but I can definitely do that.

Sources/AndroidBackend/AndroidBackend.swift
bbrk24: At the very least, we should round down in `setSize(of:to:)`. If we round up both there and in the measurement functions, we could unnecessarily make things bigger than they should be. Here's some graphs with density = 1.2, where the blue line is the hypothetical world in which there's no rounding at all, and the points are with rounding.

Rounding up in naturalSize, down in setSize (current implementation):
Screenshot 2026-05-10 at 9 33 13 PM

Rounding down in both (obviously wrong):
Screenshot 2026-05-10 at 9 34 09 PM

Rounding up in both (your suggestion):
Screenshot 2026-05-10 at 9 34 39 PM

Rounding down in naturalSize, up in setSize:
Screenshot 2026-05-10 at 9 35 20 PM

I looked at these graphs for a few different values of density and concluded that the first was most correct.

stackotter: Thanks for those graphs, makes sense now! Looks like you really thought about it

stackotter (approved):

PR #568, opened 2026-05-11
Implement progress views on Android

bbrk24:
I'm not sure whether the spinner should be `Widget_ProgressBar_Small` or `Widget_ProgressBar_Large`. The large one has a higher-resolution image and looks better at very large sizes, but the small one has a closer natural size to other platforms.

### DIFF
--- a/Sources/AndroidBackend/AndroidBackend+Progress.swift
+++ b/Sources/AndroidBackend/AndroidBackend+Progress.swift
@@ -1,0 +1,58 @@
+import AndroidKit
+import SwiftCrossUI
+import SwiftJava
+
+// swiftlint:disable force_try
+extension AndroidBackend: BackendFeatures.ProgressBars, BackendFeatures.ProgressSpinners {
+    public func createProgressBar() -> Widget {
+        let Rstyle = try! JavaClass<AndroidKit.R.style>()
+
+        let widget = AndroidKit.ProgressBar(
+            Self.activity,
+            nil,
+            0,
+            Rstyle.Widget_ProgressBar_Horizontal,
+            environment: Self.env
+        )
+
+        widget.setMin(0)
+        widget.setMax(10_000)
+
+        return widget.as(AndroidKit.View.self)!
+    }
+
+    public func updateProgressBar(
+        _ widget: Widget,
+        progressFraction: Double?,
+        environment: EnvironmentValues
+    ) {
+        let progressBar = widget.as(AndroidKit.ProgressBar.self)!
+
+        if let progressFraction {
+            progressBar.setProgress(Int32(progressFraction * 10_000))
+            progressBar.setIndeterminate(false)
+        } else {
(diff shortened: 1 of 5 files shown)


Sources/AndroidBackend/AndroidBackend+Progress.swift
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    public func createProgressSpinner() -> Widget {
+        let Rstyle = try! JavaClass<AndroidKit.R.style>()
+
+        let widget = AndroidKit.ProgressBar(
+            Self.activity,
+            nil,
+            0,
+            Rstyle.Widget_ProgressBar_Small,
```
stackotter, line 46: From the [ProgressBar docs](https://developer.android.com/reference/kotlin/android/widget/ProgressBar#determinate-progress)

> Other progress bar styles provided by the system include:
>
> - [Widget.ProgressBar.Horizontal](https://developer.android.com/reference/kotlin/android/R.style.html#Widget_ProgressBar_Horizontal:kotlin.Int)
> - [Widget.ProgressBar.Small](https://developer.android.com/reference/kotlin/android/R.style.html#Widget_ProgressBar_Small:kotlin.Int)
> - [Widget.ProgressBar.Large](https://developer.android.com/reference/kotlin/android/R.style.html#Widget_ProgressBar_Large:kotlin.Int)
> - [Widget.ProgressBar.Inverse](https://developer.android.com/reference/kotlin/android/R.style.html#Widget_ProgressBar_Inverse:kotlin.Int)
> - [Widget.ProgressBar.Small.Inverse](https://developer.android.com/reference/kotlin/android/R.style.html#Widget_ProgressBar_Small_Inverse:kotlin.Int)
> - [Widget.ProgressBar.Large.Inverse](https://developer.android.com/reference/kotlin/android/R.style.html#Widget_ProgressBar_Large_Inverse:kotlin.Int)
> 
> The "inverse" styles provide an inverse color scheme for the spinner, which may be necessary if your application uses a light colored theme (a white background).

Should we be using the 'small.inverse' style when the color scheme is light?

stackotter (changes requested):
Looks great! I don't have any changes to request, but I did come across something in the docs that may affect the implementation. You'll probably know better than me whether it's something that we have to worry about though

Sources/AndroidBackend/AndroidBackend+Progress.swift
bbrk24: I tested it and, at least on my emulator, `Small.Inverse` and `Large.Inverse` look exactly the same as `Small` and `Large` respectively. It's possible it depends on Android version or vendor, so I can do more testing later.

stackotter: Ok, I'm happy with that; I guess we can check again later if it becomes a noticeable issue

stackotter (approved):

PR #571, opened 2026-05-13
Android tap gesture support

bbrk24:
I also created a very basic TapGesturesExample, and added it and and FontsExample to CI.

### DIFF
--- a/Sources/AndroidBackend/AndroidBackend+TapGestures.swift
+++ b/Sources/AndroidBackend/AndroidBackend+TapGestures.swift
@@ -1,0 +1,45 @@
+import AndroidKit
+import SwiftCrossUI
+
+extension AndroidBackend: BackendFeatures.TapGestures {
+    public func createTapGestureTarget(wrapping child: Widget, gesture: TapGesture) -> Widget {
+        child
+    }
+
+    public func updateTapGestureTarget(
+        _ tapGestureTarget: Widget,
+        gesture: TapGesture,
+        environment: EnvironmentValues,
+        action: @escaping () -> Void
+    ) {
+        switch gesture.kind {
+            case .primary:
+                if environment.isEnabled {
+                    tapGestureTarget.setOnClickListener(
+                        ViewOnClickListener(action: action, environment: Self.env)
+                            .as(AndroidKit.View.OnClickListener.self)!
+                    )
+                } else {
+                    tapGestureTarget.setOnClickListener(nil)
+                }
+            case .secondary:
+                if environment.isEnabled {
+                    tapGestureTarget.setOnTouchListener(
+                        SecondaryClickListener(action: action, environment: Self.env)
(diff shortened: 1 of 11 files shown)


Sources/AndroidBackend/AndroidBackend+TapGestures.swift
```diff
@@ -0,0 +1,45 @@
...
+        environment: EnvironmentValues,
+        action: @escaping () -> Void
+    ) {
+        switch gesture.kind {
+            case .primary:
+                if environment.isEnabled {
+                    tapGestureTarget.setOnClickListener(
+                        ViewOnClickListener(action: action, environment: Self.env)
+                            .as(AndroidKit.View.OnClickListener.self)!
+                    )
```
stackotter, line 21: How annoying would it be for us to reuse the underlying handler when non-nil by instead just setting its action? If it's too annoying don't worry about it (maybe just add a TODO or create an issue about optimising that)

Sources/AndroidBackend/ViewOnLongClickListener.swift
```diff
@@ -0,0 +1,32 @@
...
+
+@JavaImplementation("dev.swiftcrossui.androidbackend.ViewOnLongClickListener")
+extension ViewOnLongClickListener {
+    @JavaMethod
+    func onLongClick() -> Bool {
+        let action = getAction()!.value() as! Action
+        action()
+        return true
+    }
+}
```
stackotter: Can we implement this on the Kotlin side instead to minimise the amount of JNI going on?

stackotter (changes requested):

Sources/AndroidBackend/ViewOnLongClickListener.swift
bbrk24: ViewOnLongClickListener is pretty much a copy-paste from the existing ViewOnClickListener. Should I change both?

Sources/AndroidBackend/AndroidBackend+TapGestures.swift
bbrk24: There is no `getOnClickListener` etc methods, so I'd have to create an entire container widget just to add the ability to get the underlying handler back out of the view.

stackotter: Yeah ok that's the main thing I was worried about. Don't bother then

Sources/AndroidBackend/ViewOnLongClickListener.swift
stackotter: Ah ok, makes sense. I didn't see the OnClick one because it's not in the diffs 😅 yeah it'd be great if you could update both if it's not too annoying, otherwise we can leave it for later

stackotter (approved):
Thanks for cleaning up those handlers, they look much cleaner now

PR #577, opened 2026-05-14
Feature Parity: Gradients on AndroidBackend

MiaKoring:
Adds support for all currently supported types of gradients on AndroidBackend.

- [x] LinearGradient
- [x] RadialGradient
- [x] AngularGradient
- [x] Fix class not found for RadialGradient
- [x] Test RadialGradient

Current issue with RadialGradient:
```
SwiftRuntime: AndroidGraphics/RadialGradient.swift:7: Fatal error: 'try!' expression unexpectedly raised an error: java.lang.ClassNotFoundException: Didn't find class "android.graphics.RadialGradient" on path: DexPathList[[zip file "/data/app/~~59ImuJ13QCBQxXjCQGr2Yw==/dev.swiftcrossui.gradientsexample-Hku2_NZTVGz5A6mElLgVUQ==/base.apk"],nativeLibraryDirectories=[/data/app/~~59ImuJ13QCBQxXjCQGr2Yw==/dev.swiftcrossui.gradientsexample-Hku2_NZTVGz5A6mElLgVUQ==/lib/arm64, /data/app/~~59ImuJ13QCBQxXjCQGr2Yw==/dev.swiftcrossui.gradientsexample-Hku2_NZTVGz5A6mElLgVUQ==/base.apk!/lib/arm64-v8a, /system/lib64, /system_ext/lib64]]
05-14 22:01:44.713  6923  7048 E Swift   : AndroidGraphics/RadialGradient.swift:7: Fatal error: 'try!' expression unexpectedly raised an error: java.lang.ClassNotFoundException: Didn't find class "android.graphics.RadialGradient" on path: DexPathList[[zip file "/data/app/~~59ImuJ13QCBQxXjCQGr2Yw==/dev.swiftcrossui.gradientsexample-Hku2_NZTVGz5A6mElLgVUQ==/base.apk"],nativeLibraryDirectories=[/data/app/~~59ImuJ13QCBQxXjCQGr2Yw==/dev.swiftcrossui.gradientsexample-Hku2_NZTVGz5A6mElLgVUQ==/lib/arm64, /data/app/~~59ImuJ13QCBQxXjCQGr2Yw==/dev.swiftcrossui.gradientsexample-Hku2_NZTVGz5A6mElLgVUQ==/base.apk!/lib/arm64-v8a, /system/lib64, /system_ext/lib64]]
```

Edit: The above issue was resolved.

This PR implements Gradient support on Android.
To make the upcoming support for gradient filled shapes easier, it is already built on Path.
Android Paths can be filled by setting a shader on the respective paint. 
android.graphics.LinearGradient/RadialGradient/SweepGradient(conic/angular) implement Shader.

Shaders expect dimensions to be passed in pixels, size is dp. I convert it to pixels by multiplying by density.

Radial and AngularGradient use the existing adjusted stops computed property to polyfill support for startRadius on RadialGradient and endAngle on AngularGradient.

To match our existing AngularGradient behavior on all other supported platforms, the SweepGradient is transformed via a android.graphics.matrix.
The matrix both stretches/squeezes the SweepGradientShader so its elliptical, like CAGradientLayer/SwiftUI and rotates it by startAngle.degrees, to polyfill support for startAngle.

@bbrk24 knows Path and the AndroidBackend better than I do, so I welcome it’s suggestions regarding performance optimizations/simplifications. 

Though I would like to mention that all create/updateFooGradient functions will be removed in the future and handled inside the path methods.

### DIFF
--- a/Examples/Sources/GradientsExample/GradientsApp.swift
+++ b/Examples/Sources/GradientsExample/GradientsApp.swift
@@ -65,16 +65,16 @@ struct GradientsApp: App {
         case .linear:
             LinearGradientView()
         case .radial:
-            #if !canImport(AndroidBackend)
             RadialGradientView()
-            #endif
         case .angular:
             #if !canImport(WinUIBackend) && !canImport(GtkBackend) && !canImport(AndroidBackend)
-            ScrollView(.horizontal) {
+                ScrollView(.horizontal) {
+                    AngularGradientView()
+                }
+            #elseif canImport(AndroidBackend)
                 AngularGradientView()
-            }
             #else
-            Text("Angular Gradients are not supported on \(App.Backend)")
+                Text("Angular Gradients are not supported on \(App.Backend)")
             #endif
         case .none:
             Text("Please select a gradient type.")
--- a/Examples/Sources/GradientsExample/GradientsApp.swift
+++ b/Examples/Sources/GradientsExample/GradientsApp.swift
@@ -20,7 +20,7 @@ struct GradientsApp: App {
 
     var body: some Scene {
         WindowGroup("Gradients Example") {
-            #if !canImport(UIKitBackend)
+            #if !canImport(UIKitBackend) && !canImport(AndroidBackend)
                 NavigationSplitView {
                     // TODO: replace with List once bug described in #556 is fixed
                     ForEach(GradientType.allCases, id: \.rawValue) { type in
@@ -50,23 +50,34 @@ struct GradientsApp: App {
 
     @ViewBuilder
     func scrollViewWithGradient() -> some View {
+        #if !canImport(AndroidBackend)
         ScrollView {
-            switch gradientType {
-                case .linear:
-                    LinearGradientView()
-                case .radial:
-                    RadialGradientView()
-                case .angular:
-                    #if !canImport(WinUIBackend) && !canImport(GtkBackend)
-                        ScrollView(.horizontal) {
-                            AngularGradientView()
-                        }
-                    #else
-                        Text("Angular Gradients are not supported on \(App.Backend)")
-                    #endif
-                case .none:
-                    Text("Please select a gradient type.")
+            gradients()
+        }
+        #else
+            gradients()
+        #endif
+    }
+
+    @ViewBuilder
+    func gradients() -> some View {
+        switch gradientType {
+        case .linear:
+            LinearGradientView()
+        case .radial:
+            #if !canImport(AndroidBackend)
+            RadialGradientView()
+            #endif
+        case .angular:
+            #if !canImport(WinUIBackend) && !canImport(GtkBackend) && !canImport(AndroidBackend)
+            ScrollView(.horizontal) {
+                AngularGradientView()
             }
+            #else
+            Text("Angular Gradients are not supported on \(App.Backend)")
+            #endif
+        case .none:
+            Text("Please select a gradient type.")
         }
     }
 }
--- a/Examples/Sources/GradientsExample/GradientsApp.swift
+++ b/Examples/Sources/GradientsExample/GradientsApp.swift
@@ -51,9 +51,9 @@ struct GradientsApp: App {
     @ViewBuilder
     func scrollViewWithGradient() -> some View {
         #if !canImport(AndroidBackend)
-        ScrollView {
-            gradients()
-        }
+            ScrollView {
+                gradients()
+            }
         #else
             gradients()
         #endif
@@ -62,22 +62,22 @@ struct GradientsApp: App {
     @ViewBuilder
     func gradients() -> some View {
         switch gradientType {
-        case .linear:
-            LinearGradientView()
-        case .radial:
-            RadialGradientView()
-        case .angular:
-            #if !canImport(WinUIBackend) && !canImport(GtkBackend) && !canImport(AndroidBackend)
-                ScrollView(.horizontal) {
+            case .linear:
+                LinearGradientView()
+            case .radial:
+                RadialGradientView()
+            case .angular:
+                #if !canImport(WinUIBackend) && !canImport(GtkBackend) && !canImport(AndroidBackend)
+                    ScrollView(.horizontal) {
+                        AngularGradientView()
+                    }
+                #elseif canImport(AndroidBackend)
                     AngularGradientView()
-                }
-            #elseif canImport(AndroidBackend)
-                AngularGradientView()
-            #else
-                Text("Angular Gradients are not supported on \(App.Backend)")
-            #endif
-        case .none:
-            Text("Please select a gradient type.")
+                #else
+                    Text("Angular Gradients are not supported on \(App.Backend)")
+                #endif
+            case .none:
+                Text("Please select a gradient type.")
         }
     }
 }
--- a/Sources/AndroidBackend/AndroidBackend+Gradients.swift
+++ b/Sources/AndroidBackend/AndroidBackend+Gradients.swift
@@ -5,7 +5,8 @@ import AndroidGraphics
 // swiftlint:disable force_try
 extension AndroidBackend: BackendFeatures.Gradients {
     public func createLinearGradientWidget() -> Widget {
-        return createPathWidget()
+        GradientWidget(Self.activity, environment: Self.env)
+            .as(AndroidKit.View.self)!
     }
 
     public func updateLinearGradientWidget(
@@ -14,13 +15,8 @@ extension AndroidBackend: BackendFeatures.Gradients {
         withSize size: SIMD2<Int>,
         in environment: EnvironmentValues
     ) {
-        let scuiPath = SwiftCrossUI.Rectangle().path(in: .init(origin: .zero, size: .init(Double(size.x), Double(size.y))))
-        let path = createPath()
-
         let tileClass = try! JavaClass<AndroidGraphics.Shader.TileMode>()
         let colorClass = try! JavaClass<AndroidGraphics.Color>()
-
-        let density = Float(environment.windowScaleFactor)
 
         let count = gradient.gradient.stops.count
         var stops = [Float]()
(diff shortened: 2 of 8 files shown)


MiaKoring:
Huge thanks to @bbrk24 for helping me with working around RadialGradient class not being found by subclassing it and accessing the subclass instead

MiaKoring:
Improves compatibility as mentioned in https://github.com/moreSwift/swift-cross-ui/issues/427

Sources/AndroidBackend/AndroidBackend+Gradients.swift
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    public func updateLinearGradientWidget(
+        _ widget: Widget,
+        gradient: SwiftCrossUI.LinearGradient,
+        withSize size: SIMD2<Int>,
+        in environment: EnvironmentValues
+    ) {
+        let scuiPath = SwiftCrossUI.Rectangle().path(in: .init(origin: .zero, size: .init(Double(size.x), Double(size.y))))
+        let path = createPath()
```
bbrk24: I don't love that we're creating a new Path every time update is called. I understand you reused the existing PathView for simplicity, but this should probably be a separate widget that creates the Path and Paint on initialization. If you want, I can help you with that later (~5h from now) because I know you aren't familiar with Kotlin.

Sources/AndroidBackend/AndroidBackend+Gradients.swift
```diff
@@ -0,0 +1,218 @@
...
+
+        let tileClass = try! JavaClass<AndroidGraphics.Shader.TileMode>()
+        let colorClass = try! JavaClass<AndroidGraphics.Color>()
+
+        let density = widget.getResources().getDisplayMetrics().density
```
bbrk24: This is also present on the environment so we don't need JNI calls to fetch it every time.

Sources/AndroidBackend/AndroidBackend+Gradients.swift
```diff
@@ -0,0 +1,218 @@
...
+
+        updatePath(
+            path,
+            scuiPath,
+            bounds: .init(origin: .zero, size: .init(x: Double(size.x), y: Double(size.y))),
+            pointsChanged: true,
+            environment: environment
+        )
+
+        setSize(of: widget, to: size)
```
bbrk24: Is this needed here? Ideally SCUI should be calling this for every view anyways.

Sources/AndroidBackend/AndroidBackend+Gradients.swift
```diff
@@ -0,0 +1,218 @@
...
+            stops,
+            environment: Self.env
+        )
+
+        let gradientMatrix = AndroidGraphics.Matrix()
+
+        let scaleX: Float = 1.0
+        let scaleY: Float = Float(size.y) / Float(size.x)
+
+        gradientMatrix.reset()
```
bbrk24: Is this call needed on a freshly created matrix? Though ideally, like with Path and Paint above, the Matrix would be reused too.

Sources/AndroidBackend/Kotlin/CustomRadialGradient.kt
```diff
@@ -0,0 +1,13 @@
+package dev.swiftcrossui.androidbackend
+
+import android.graphics.RadialGradient
+import android.graphics.Shader
+
+class CustomRadialGradient(
```
bbrk24, line 6: We should probably have a comment here explaining why this exists.

Sources/AndroidBackend/AndroidBackend+Gradients.swift
MiaKoring: I will give it a try. With the diff you gave me I believe I have some basic idea of how it could work. If I don’t succeed I will take the help, but I also want to learn how this works so I can help more efficiently with android things in the future :)

Sources/AndroidBackend/AndroidBackend+Gradients.swift
MiaKoring: ahh windowScaleFactor 👍

Sources/AndroidBackend/AndroidBackend+Gradients.swift
MiaKoring: It was, in fact, not needed lol
I’m not sure why I added it

Sources/AndroidBackend/AndroidBackend+Gradients.swift
MiaKoring: I’m now re-using path and matrix.
All gradients call set(Shader, width, height).
this function sets the shader on the paint and updates the path to a rectangle with the passed dimensions on the kotlin side.

AngularGradient calls setMatrix after set() to add the matrix to the shader, re-using the Matrix instance on the Kotlin side. 

matrix is initialized with null and set by setMatrix when needed, because linear and angular don’t need a matrix.

Color.RED on the paint is just an arbitrary, opaque color, Color.TRANSPARENT causes the gradient to not be drawn. I assume this is because a shader needs opaque pixels to do its shady shader things?

Sources/AndroidBackend/Kotlin/GradientWidget.kt
```diff
@@ -0,0 +1,54 @@
...
+import android.graphics.SweepGradient
+import android.graphics.Matrix
+import android.view.View
+
+class GradientWidget(activity: Activity): View(activity) {
+    private var path = Path()
+    private var matrix: Matrix? = null
+    private var fillPaint = Paint().apply {
+        style = Paint.Style.FILL
+        color = Color.RED
```
bbrk24: I know you said setting `Color.TRANSPARENT` doesn't work, but do you need to set color at all?

Sources/AndroidBackend/Kotlin/GradientWidget.kt
```diff
@@ -0,0 +1,54 @@
...
+        isAntiAlias = true
+    }
+    
+    fun set(shader: Shader, width: Float, height: Float) {
+        this.fillPaint.shader = shader
+        
+        // Reset path and draw rectangle for current bounds
+        this.path.apply {
+            reset()
```
bbrk24: Since the general structure is the same every time, it's probably better to call `rewind()` than `reset()`.

bbrk24 (commented):
Mostly looks good to me, just a couple very minor comments

Sources/AndroidBackend/Kotlin/GradientWidget.kt
MiaKoring: Ah read the docs for it, I agree it sounds well suited (and quite nice), will change it later

Sources/AndroidBackend/Kotlin/GradientWidget.kt
MiaKoring: I do not. I did’t think about this being a possibility xD
I was like Paint needs Color.

bbrk24 (approved):

Examples/Sources/GradientsExample/GradientsApp.swift
```diff
@@ -50,23 +50,34 @@ struct GradientsApp: App {
 
     @ViewBuilder
     func scrollViewWithGradient() -> some View {
+        #if !canImport(AndroidBackend)
         ScrollView {
```
stackotter: Indent this ScrollView

Examples/Sources/GradientsExample/GradientsApp.swift
```diff
@@ -50,23 +50,34 @@ struct GradientsApp: App {
...
+        }
+        #else
+            gradients()
+        #endif
+    }
+
+    @ViewBuilder
+    func gradients() -> some View {
+        switch gradientType {
+        case .linear:
```
stackotter: Indent this switch

Sources/AndroidBackend/AndroidBackend+Gradients.swift
```diff
@@ -0,0 +1,173 @@
...
+        let count = gradient.gradient.stops.count
+        var stops = [Float]()
+        var colors = [Int32]()
+        stops.reserveCapacity(count)
+        colors.reserveCapacity(count)
+
+        for stop in gradient.gradient.stops {
+            stops.append(Float(stop.location))
+            colors.append(stop.color.resolve(in: environment).asColorInt())
+        }
```
stackotter, line 25: I think we can simplify this like so;

```swift
let stops = gradient.gradient.stops.map(\.location)
let colors = gradient.gradient.stops.map { stop in
    stop.color.resolve(in: environment).asColorInt()
}
```

For our purposes the performance should be essentially the same (or maybe slightly better), and importantly it's more readable (less noise)

Sources/AndroidBackend/AndroidBackend+Gradients.swift
```diff
@@ -0,0 +1,173 @@
...
+        let count = gradient.gradient.stops.count
+        var stops = [Float]()
+        var colors = [Int32]()
+        stops.reserveCapacity(count)
+        colors.reserveCapacity(count)
+
+        for stop in gradient.adjustedStops {
+            stops.append(Float(stop.location))
+            colors.append(stop.color.resolve(in: environment).asColorInt())
+        }
```
stackotter, line 70: Same as previous comment

Sources/AndroidBackend/AndroidBackend+Gradients.swift
```diff
@@ -0,0 +1,173 @@
...
```
stackotter, line 117: Same as previous comment

Sources/AndroidBackend/AndroidBackend+Gradients.swift
```diff
@@ -0,0 +1,173 @@
...
+        let gradient = AndroidGraphics.SweepGradient(
+            centerX,
+            centerY,
+            colors,
+            stops,
+            environment: Self.env
+        )
+
+        let scaleX: Float = 1.0
+        let scaleY: Float = Float(size.y) / Float(size.x)
```
stackotter: Unnecessary type annotation

stackotter (changes requested):
Looks great for the most part! And worked well in my local tests.

Just left a few small comments (mostly for formatting stuff)

Sources/AndroidBackend/AndroidBackend+Gradients.swift
MiaKoring: had to make the first one { Float($0.location) } (because it would be double)

No named closure parameter is okay for same-line body right?

stackotter: Yeah that'd be an acceptable use for a numbered parameter. Although, you could write `.map(\.location).map(Float.init(_:))` instead I think

stackotter: But either is fine

stackotter (approved):
Thanks for addressing those comments!

PR #578, opened 2026-05-14
Android sheets

bbrk24:


Sources/AndroidBackend/AndroidBackend+Sheets.swift
```diff
@@ -0,0 +1,68 @@
...
+            let width = helpers.getSafeWindowWidth(Self.activity)
+            let widthMeasureSpec = (width & 0x3FFFFFFF) | 0x40000000
+            content.measure(widthMeasureSpec, 0x3FFFFFFF)
+            let density = content.getResources().getDisplayMetrics().density
+            let height = Float(content.getMeasuredHeight()) / density
+            return SIMD2(Int(width), Int(height.rounded(.up)))
+        } else {
+            return .zero
+        }
+    }
```
stackotter, line 22: Is this effectively just measuring back the size that SwiftCrossUI has assigned to the sheet's content? I guess it's working, but I feel like that'd return odd results on the first update. Now that I look at the UIKitBackend sheet implementation, it may have similarly weird issues... I think the sheets API will need to be updated to account for mobile-style fixed-width variable-height sheets a bit better, but that should be left for a future PR

Sources/AndroidBackend/AndroidBackend+Sheets.swift
```diff
@@ -0,0 +1,68 @@
...
+        sheet.setOnDismissListener(SwiftAction(environment: Self.env, action: onDismiss))
+
+        let backgroundColor =
+            if let backgroundColor {
+                backgroundColor
+            } else {
+                switch environment.colorScheme {
+                    case .dark: SwiftCrossUI.Color.Resolved(red: 0, green: 0, blue: 0)
+                    case .light: SwiftCrossUI.Color.Resolved(red: 1, green: 1, blue: 1)
+                }
```
stackotter, line 69: Are these the default sheet background colours on Android? The black background is almost invisible on my phone in dark mode, but if that's how Android does things we can do that for now.

---

I just tried it out for myself by commenting out the background colour logic in CustomSheet, and it made the sheet background basically transparent?? 🤔

stackotter (changes requested):

stackotter:
Other than my last comment, everything introduced in this PR seems to be working well

Sources/AndroidBackend/AndroidBackend+Sheets.swift
bbrk24: No; in dark mode, the default sheet background color is a shade of gray rather than black, but I didn't find a resource ID corresponding to that color. I suppose I could color-pick it from the emulator and hardcode that in?

stackotter: Yeah I think that'd be an acceptable approach for now (with a comment about where the magic value came from)

stackotter (approved):
Fix looks good! Just going to test locally rq and then merge

Issue #579, opened 2026-05-15
AndroidBackend crashes immediately pre-API 34


bbrk24:
- The window measurement functions call `activity.getWindowManager().getCurrentWindowMetrics()`. `getCurrentWindowMetrics()` was introduced in API 30, so this crashes at runtime on API 28 or 29.
- There's a few places where we're calling `TypedValue.deriveDimension`. This was introduced in API 34, so it crashes at runtime on API 33 or earlier.
  - The solution to this one is to use the one on [`TypedValueCompat`](https://developer.android.com/reference/androidx/core/util/TypedValueCompat#deriveDimension(int,float,android.util.DisplayMetrics)) instead. However, I'm not 100% sure whether we have access to `androidx.core` currently.
- In AlertFragment.kt, we're calling `TypedValue.convertDimensionToPixels`. This is an alternate spelling of `TypedValue.applyDimension` with worse availability, so we should switch it to that instead.

These are all the problematic calls I found in Kotlin, but there may be more in Swift.

bbrk24:
Ah, we don't seem to have access to androidx.core, and it doesn't seem to be possible to determine the size of a window before the view is attached to in on API 29.

finagolfin:
Many app stores [don't support lower than API 30](https://seller.samsungapps.com/notice/getNoticeDetail.as?csNoticeID=0000008212) anyway, so better to [just check for that](https://developer.android.com/reference/kotlin/android/os/Build.VERSION) and fall back or error.

bbrk24:
> Many app stores [don't support lower than API 30](https://seller.samsungapps.com/notice/getNoticeDetail.as?csNoticeID=0000008212) anyway

That seems to be talking about targetSdk, not minSdk? In any case, I would be fine with just changing our minSdk from 28 to 30, now that the bad `TypedValue` calls have been removed.

bbrk24:
In case this information helps: API 30 is equivalent to iOS 14 by release year and iOS 18 by usage. By usage, iOS 13 would be roughly equivalent to API 23, which is also the oldest minSdk androidx supports (and older than Swift currently supports).

finagolfin:
[You would be wrong about API 23 support](https://forums.swift.org/t/android-api-minimum-for-the-swift-sdk-for-android/82874/38), 😉 at least for the upcoming release.

PR #581, opened 2026-05-16
Use AppCompatDialogFragment

bbrk24:
When I first wrote `AlertFragment`, I didn't realize we already had access to `androidx.appcompat`. This PR switches over, and also changes `convertDimensionToPixels` to `applyDimension` as mentioned in #579.


stackotter (approved):
Looks good! Just going to test locally before merging

PR #582, opened 2026-05-16
AndroidBackend: API 30 compatibility

bbrk24:
This PR updates the helpers so that `AndroidBackendHelpers` doesn't use any functionality introduced in API 34, partially resolving #579. It still requires API 30 and crashes instantly on API 28 or 29. I also didn't touch the AlertFragment since I already did that in #581.

### DIFF
--- a/Sources/AndroidBackend/Kotlin/AndroidBackendHelpers.kt
+++ b/Sources/AndroidBackend/Kotlin/AndroidBackendHelpers.kt
@@ -3,6 +3,7 @@ package dev.swiftcrossui.androidbackend
 import android.R
 import android.app.Activity
 import android.content.res.Configuration
+import android.os.Build
 import android.util.TypedValue
 import android.view.WindowInsets
 import android.widget.TextView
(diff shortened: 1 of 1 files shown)


MiaKoring:
image

The most important part: Works!

Thank you, thats really helpful. I will give onHover a shot again once this is merged

MiaKoring:
The code changes look logical too. Basically same logic, but different source for density and basically polyfilling the one function for older versions. 

Only question I have: Is there a benefit to using the function on the newer API version instead of the division / do we know what the function does internally? if the function just does a division too, it might be nicer to use it for both. Otherwise, if the function is cheaper it undoubtedly makes sense to do it like that

bbrk24:
Yes, starting with API 34 the relationship between SP and PX is no longer guaranteed to be linear, so a simple division is not necessarily correct.

MiaKoring (approved):
lgtm

stackotter (approved):
Thanks for resolving these compatibility issues 🙏

PR #587, opened 2026-05-24
[FEAT] ContentUnavailableView

### LINKED ISSUE #365: ContentUnavailableView
`ContentUnavailableView` is a nice view to use as a placeholder:

https://developer.apple.com/documentation/swiftui/contentunavailableview

Adwaita has a similar component:

https://developer.gnome.org/hig/patterns/feedback/placeholders.html

MiaKoring:
This pullrequest adds a new View `ContentUnavailableView`.
Due to us not having icons yet, I only added the ViewBuilder initializer.

SwiftUI’s `ContentUnavailableView` adds some constraints and EnvironmentValues, which I tried to mimic

The Font for the respective components differs by platform, so does the color for the label. Those differences are resolved via computed properties, checking the deviceClass.

The actions are HStack on desktop and VStack on everything else I tested, as reflected in the body.

I did not test GtkBackend and WinUIBackend, but due to it being an SCUI Level change without backend modification, I don’t expect any issues as long as the device class == .desktop.

ContentUnavailableExample can be removed after testing or merged with a different Example, depending on what is preferred

Partially resolves https://github.com/moreSwift/swift-cross-ui/issues/365

I will implement .search and the other initializers once we have an icon system.


Examples/Sources/ContentUnavailableExample/ContentUnavailableViewApp.swift
```diff
@@ -0,0 +1,27 @@
...
+    import SwiftBundlerRuntime
+#endif
+
+@main
+@HotReloadable
+struct ContentUnavailableViewApp: App {
+    var body: some Scene {
+        WindowGroup("Content Unavailable") {
+            #hotReloadable {
+                ContentUnavailableView(){
```
stackotter: Should be able to remove the `()` here

Examples/Sources/ContentUnavailableExample/ContentUnavailableViewApp.swift
```diff
@@ -0,0 +1,27 @@
...
+                    Color.blue
+                        .frame(height: 200)
+                    Text("Test")
+                } description: {
+                    Text("test")
+                } actions: {
+                    Button("Test") {}
+                    Button("Test2") {}
+                }
```
stackotter: If this example is staying in the repo, change the text and button labels to make sense for some hypothetical use case of `ContentUnavailableView` (rather than just `Test` with varying capitalisation and numbers)

Sources/SwiftCrossUI/Views/ContentUnavailableView.swift
```diff
@@ -0,0 +1,76 @@
+/// An interface, consisting of a label and additional content,
+/// that you display when the content of your app is unavailable to users.
+public struct ContentUnavailableView<Label: View, Description: View, Actions: View>: View {
+
```
stackotter: Remove this blank line. Does the formatter do that automatically? If not we may have some config to change

Sources/SwiftCrossUI/Views/ContentUnavailableView.swift
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    private var label: Label
+    private var description: Description
+    private var actions: Actions
+
+    @Environment(\.backend) var backend
+
+    var labelFont: Font {
+        switch backend.deviceClass {
```
stackotter: Switch on `deviceClass.kind` so that the switches can be exhaustive (and replace the `default` with any cases that are missing from the existing switches, if there are any)

Sources/SwiftCrossUI/Views/ContentUnavailableView.swift
```diff
@@ -0,0 +1,76 @@
...
+            default: .subheadline
+        }
+    }
+
+    var labelColor: Color {
+        if backend.deviceClass == .desktop { return .gray }
+        return .adaptive(light: .black, dark: .white)
+    }
+
+    public var body: some View {
```
stackotter, line 51: Does SwiftUI's ContentUnavailableView react to foregroundColor from the environment at all? If so we might have to be a bit more careful about overriding the foreground colour of e.g. the action buttons

Sources/SwiftCrossUI/Views/ContentUnavailableView.swift
```diff
@@ -0,0 +1,76 @@
...
+                    actions
+                        .font(.body)
+                }
+            }
+        }
+        .if(backend.deviceClass != .desktop) { view in
+            view.padding(30)
+        }
+        .if(backend.deviceClass == .desktop) { view in
+            view.frame(width: 360)
```
stackotter: This seems rather wide for a strict frame. Is the ContentUnavailableView really always exactly 360 px/pt wide on desktop in SwiftUI?

stackotter (changes requested):

Examples/Sources/ContentUnavailableExample/ContentUnavailableViewApp.swift
```diff
@@ -0,0 +1,27 @@
...
+
+@main
+@HotReloadable
+struct ContentUnavailableViewApp: App {
+    var body: some Scene {
+        WindowGroup("Content Unavailable") {
+            #hotReloadable {
+                ContentUnavailableView(){
+                    Color.blue
+                        .frame(height: 200)
```
stackotter: We should remove this `Color.blue` rectangle, it's a bit random

Sources/SwiftCrossUI/Views/ContentUnavailableView.swift
stackotter: It does not seem to be so

Sources/SwiftCrossUI/Views/ContentUnavailableView.swift
```diff
@@ -0,0 +1,76 @@
...
+    var descriptionFont: Font {
+        switch backend.deviceClass {
+            case .phone, .tablet, .tv: .subheadline
+            case .desktop: .body
+            default: .subheadline
+        }
+    }
+
+    var labelColor: Color {
+        if backend.deviceClass == .desktop { return .gray }
```
stackotter, line 47: On macOS the foreground colour seems like a different gray in light mode, but maybe my eyes deceive me

stackotter (changes requested):

Sources/SwiftCrossUI/Views/ContentUnavailableView.swift
MiaKoring: I forgot to test getting smaller 🤦‍♀️ But I did not see it getting bigger than 360. did you? If not I would just change it to maxWidth

stackotter: Oh interesting, it does appear to limit the width to 360. I think maxWidth 360 makes sense

Examples/Sources/ContentUnavailableExample/ContentUnavailableViewApp.swift
MiaKoring: Would you like to keep the Example? I mean its just one simple View 😅

I just used the blue rectangle as an image replacement in testing.

stackotter: I think we should delete the example or merge its view into some other example. I don't see any other examples where it would immediately make sense to use, other than maybe when there aren't any notes available in NotesExample?

MiaKoring: Once we have more icons I would add it to NotesExample, thats a food idea. until then I would delete it. Is that okay?

stackotter: Yep sounds good

Sources/SwiftCrossUI/Views/ContentUnavailableView.swift
MiaKoring: No that was me

Sources/SwiftCrossUI/Views/ContentUnavailableView.swift
MiaKoring: it does, good catch. I updated it to use the environment foreground color if set.

MiaKoring: Screenshot 2026-05-25 at 08 52 42
(Just so you can see it working without the ability to test when I remove the example)

Sources/SwiftCrossUI/Views/ContentUnavailableView.swift
MiaKoring: it is, but thats an issue with our gray being different in general. SwiftUIs gray is the same gray as the text.

stackotter: The gray seems to match much closer in dark mode. If that's not an optical illusion then I guess SwiftUI is using Color.secondary (or similar) which must react differently to the color scheme than regular grey?

MiaKoring: If I do a Color.gray rectangle in light mode above it it looks the same to me (in SwiftUI). Our gray is just different

stackotter (approved):

Issue #433, opened 2026-02-08
Runtime crash inside swift_weakLoadStrong


bbrk24:
Reproducer: https://github.com/bbrk24/weakLoadStrong_crash_reproducer

In GtkBackend specifically, if a Color's tap gesture removes it from the view, the `guard let self` inside the signal handler causes a crash. To reproduce, run the project, then click anywhere in the window.

stackotter:
Resolved by #588

Issue #434, opened 2026-02-09
[GtkBackend] Clicking pixel-perfect on the edge of something with onTapGesture{} causes a crash inside libgobject


bbrk24:
No idea why this is. Backtrace:
```
💣 Program crashed: Bad pointer dereference at 0x00005652aa7c9d50

Platform: x86_64 Linux (Ubuntu 24.04.3 LTS)

Thread 0 "WeatherApp" crashed:

  0 0x00005652aa7c9d50
  1 0x00007f97cea856bd <unknown> in libgobject-2.0.so.0.8000.0
  2 0x00007f97cea857c1 <unknown> in libgobject-2.0.so.0.8000.0
  3 0x00007f97cea85883 <unknown> in libgobject-2.0.so.0.8000.0
  4 0x00007f97cf86d9a0 <unknown> in libgtk-4.so.1.1400.5
  5 0x00007f97cd93c532 <unknown> in libglib-2.0.so.0.8000.0
  6 0x00007f97cd93b49e <unknown> in libglib-2.0.so.0.8000.0
  7 0x00007f97cd99a737 <unknown> in libglib-2.0.so.0.8000.0
  8 0x00007f97cd93aa63 <unknown> in libglib-2.0.so.0.8000.0
  9 0x00007f97cd50887d <unknown> in libgio-2.0.so.0.8000.0
 10 Application.run(_:) + 282 in WeatherApp at /home/bbrk24/distrobox/ubuntu/workspace/weather-app/.build/checkouts/swift-cross-ui/Sources/Gtk/Application.swift:76:22

    74│         self.windowCallback = windowCallback
    75│ 
    76│         let status = g_application_run(applicationPointer.cast(), 0, nil)                                                                                                                
      │                      ▲
    77│         g_object_unref(applicationPointer)
    78│         return Int(status)

 11 GtkBackend.runMainLoop(_:) + 209 in WeatherApp at /home/bbrk24/distrobox/ubuntu/workspace/weather-app/.build/checkouts/swift-cross-ui/Sources/GtkBackend/GtkBackend.swift:94:16

    92│ 
    93│     public func runMainLoop(_ callback: @escaping @MainActor () -> Void) {
    94│         gtkApp.run { window in                                                                                                                                                           
      │                ▲
    95│             self.precreatedWindow = window
    96│             callback()

[...]

PR #588, opened 2026-05-24
Fix: disconnect GTK signal handlers in `deinit` to prevent crashes

furbytm:
GTK signals are bound to the widget's lifetime, not the Swift wrapper's. When the Swift wrapper deallocated, its signal boxes were freed - but GTK held live signal connections that kept firing into the now-freed memory, causing `EXC_BAD_ACCESS` and `EXC_BAD_INSTRUCTION` crashes.

> [!IMPORTANT]
> The crash was confirmed by inspecting the closure context in the debugger, which showed an `NSDateComponent`'s object occupying memory that previously held a `SignalBox1` - a clear sign of "use-after-free".

Fixes:
- Disconnect all GTK signal handlers in `deinit` before releasing the `GObject`, ensuring no callbacks fire after the Swift wrapper is gone.
- ~~`guard` against duplicate signal connections if a view is re-parented, which previously caused multiple handlers to fire per event.~~
- move all signal connections into a `registerSignals()` method and out of `didMoveToParent()` calls, which previously caused multiple handlers to fire per event every time a view was re-parented.

Potentially related issue(s) this fix would solve:
- https://github.com/moreSwift/swift-cross-ui/issues/434
- https://github.com/moreSwift/swift-cross-ui/issues/433

### DIFF
--- a/Sources/Gtk/Generated/Entry.swift
+++ b/Sources/Gtk/Generated/Entry.swift
@@ -101,10 +101,17 @@ open class Entry: Widget, CellEditable, Editable {
             gtk_entry_new_with_buffer(buffer)
         )
     }
+    
+    /// Prevents duplicate GTK signal connections if the view is re-parented.
+    private var signalsConnected = false
 
     open override func didMoveToParent() {
         super.didMoveToParent()
 
+        // guard against reconnecting signals if already set up.
+        guard !signalsConnected else { return }
+        signalsConnected = true
+      
         addSignal(name: "activate") { [weak self] () in
             guard let self else { return }
             self.activate?(self)
--- a/Sources/Gtk/Generated/Entry.swift
+++ b/Sources/Gtk/Generated/Entry.swift
@@ -101,17 +101,10 @@ open class Entry: Widget, CellEditable, Editable {
             gtk_entry_new_with_buffer(buffer)
         )
     }
-    
-    /// Prevents duplicate GTK signal connections if the view is re-parented.
-    private var signalsConnected = false
 
-    open override func didMoveToParent() {
-        super.didMoveToParent()
+    open override func registerSignals() {
+        super.registerSignals()
 
-        // guard against reconnecting signals if already set up.
-        guard !signalsConnected else { return }
-        signalsConnected = true
-      
         addSignal(name: "activate") { [weak self] () in
             guard let self else { return }
             self.activate?(self)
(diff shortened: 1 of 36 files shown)


bbrk24:
Is this related to #433?

furbytm:
> Is this related to #433?

Ah nice! Looks like it could be related as well, I added it under potential ticket references above, for ticket triaging.

stackotter:
Just checked and this does appear to resolve #433

Sources/Gtk/Generated/Entry.swift
```diff
@@ -101,10 +101,17 @@ open class Entry: Widget, CellEditable, Editable {
...
+    /// Prevents duplicate GTK signal connections if the view is re-parented.
+    private var signalsConnected = false
 
     open override func didMoveToParent() {
         super.didMoveToParent()
 
+        // guard against reconnecting signals if already set up.
+        guard !signalsConnected else { return }
+        signalsConnected = true
+
```
stackotter: This fix seems oddly generic to live in Entry.

I think the issue can be resolved instead by outputting generated signal handlers to `registerSignals` instead of `didMoveToParent`.

Either way, this is a generated file so it shouldn't be manually edited. The generator at Source/GtkCodeGen should be edited instead.

stackotter (changes requested):

Sources/Gtk/Generated/Entry.swift
furbytm: Oh duh 😅 I had literally just noticed this file is in a `Generated` subdirectory, no problem I can do that.

furbytm: done.

Sources/GtkCodeGen/GtkCodeGen.swift
```diff
@@ -644,7 +642,7 @@ struct GtkCodeGen {
             exprs.append(expr.description)
         }
 
-        let methodName = isWidget ? "didMoveToParent" : "registerSignals"
+      let methodName = "registerSignals"
```
stackotter, line 645: This line seems outdented

stackotter (changes requested):
Thanks for fixing that up! Implementation looks good now. Just commented on one formatting issue

Sources/GtkCodeGen/GtkCodeGen.swift
furbytm: Yep, I saw that, I fixed it here already:
https://github.com/moreSwift/swift-cross-ui/pull/588/changes/5fda606858a42c478f4af1a35ccb80b1f120de0a

stackotter (approved):

stackotter:
Will merge once CI passes

PR #590, opened 2026-05-25
Better Buttons

MiaKoring:
This PR adds support for arbitrary Views as Button labels as well as 3 basic ButtonStyles: plain, borderless and bordered.

<sup>This PR only adds viewbuilder support for freestanding buttons. Buttons for Toggle, Menu or Menu children are excluded.</sup>

The entire implementation of buttons is on the backend side, so to SCUI they are just another widget.
The backend implementations are a bunch of code, but pretty uneventful.

## Expectations from Backends:
- render arbitrary views as button label
- animate press
- mouse activation
- keyboard activation (if supported by the platform)
- mouse/finger drag highlight/no highlight depending on being in bounds or not
- call action on mouse release, only when in button bounds
- be focusable (UIKit only tested on tvOS for that)
- show focus indicator
- support button style .bordered
- support button style .plain
- support button style .borderless
- support disabled
- buttons should be visible to screen readers as such (which I can’t test, but I did my best with adding accessibilty functions)
- if a focus ring is used, buttons without border should have a corner radius of 0

### Layout
- Buttons are always LabelSize + backend defined padding. 
- Backends are expected to place buttons centered and honor the padding they tell SCUI. 
- Backends are expected to not enforce a minSize.
- Backends are expected to handle all states of a Button in a platform-native feeling way.
- Backends are expected to only call the action  when the pointer/finger is in button bounds on release.

### Exceptions:
AppKitBackend and UIKitBackend do not handle hovered state even for .bordered, as SwiftUI doesn’t do it either. The other backends do, because there its what the native ones do. 

borderless and plain do not highlight on hover.

AndroidBackend and Gtk do not do the drag in out highlighting thing, because the native buttons don’t appear to do it.

tvOS always gets full padding, as its focus highlighting makes the background white, even on plain. As mentioned this is not a concern for SCUI though, as it just respects the backend’s decisions.

### Non-standards
For repeated activation via keyboard there is no standard, every backend does it in their native way. AppKitBackend mimics SwiftUI.

### SwiftCrossUI
- Made `Button` a generic
- updated String init to be Button<Text>
- added `@ViewBuilder` init
- changed Views relying on the old button to require Button<Text> and get their label from there
- renamed old button create/update to fooSimpleButton
- `BackendFeatures.Buttons`
  - added `createButton(_:)` taking the label widget
  - added `updateButton(_:)` same as old, but without label parameter
  - added `buttonPadding(_:)` takes environment, returns total horizontally and vertically added spacing, gets used in `Button/computeLayout(_:)`
  - added `defaultButtonStyle()` self explanatory, gets used in `Button/computeLayout(_:)`
- `Button`
  - now has children
  - `computeLayout(_:)` 
    1. modifies the environment for children for certain button styles requiring default foreground colors[^1]
    2. computes children layout with modified environment
    3. updates button with original environment
    4. requests button padding for original environment
    5. adds padding to child result size and returns that
  - `commit(_:)`
    1. sets size of button widget
    2. commits children
- `ButtonStyle`
  - added cases plain, borderless and bordered
- `View/buttonStyle(_:)
  - sets buttonstyle in the environment
- `EnvironmentValues`
  - added `var buttonStyle: ButtonStyle?`
  - added `var defaultButtonStyle: ButtonStyle`
  - added `var resolvedButtonStyle: ButtonStyle`
- `BaseStubs`
  - updated naming of old button functions
  - added new button functions

### AndroidBackend
- implemented new required functions
- added CustomButton swift accessor for the kotlin class
- added `CustomButton.kt` based on FrameLayout

updates are applied via set.
ButtonStyle.bordered uses android.R.attr.selectableItemBackground as foreground for tap animations
others use android.R.attr.selectableItemBackgroundBorderless as foreground for tap animations

ButtonStyle is converted to an integer so handling everything on the kotlin side is possible.

unbordered buttons are dimmed to 38% opacity when disabled.

### AppKitBackend
- implemented new required functions
- added `NSCustomButton` based on NSView

`NSCustomButton` overrides mouseUp, mouseDown and mouseDragged for mouse event handling
`NSCustomButton` overrides keyDown for key presses (SwiftUI doesn’t trigger on release so neither do we)

An NSButtonCell is used as background for bordered style and also draws the focus ring.
an extension of ButtonStyle handles rendered style modifications, to keep the main class smaller. The functions are called by didSet of their corresponding properties.

click animation on unbordered buttons dimms to 75% opacity, disabled to 50%

It implements accessibility functions identifying it as a button and making it possible to be called by voiceover. When its a text button the texts string is provided as accessibility description, which should be replaced by proper accessibility features later.

### GtkBackend and Gtk3Backend
- implemented new required functions
- added CustomButton based on Button
  - removed min size
  - handles states with just some css, I don’t think that needs explanation ¯\_(ツ)_/¯
  - adds/removes classes based on style to change behavior

### UIKitBackend
- implemented new required functions
- added `UICustomButton` based on UIControl
- added CustomButtonWidget based on WrapperWidget<UICustomButton>

Bordered style is implemented via a child button with userinteraction disabled. I tried doing a button based implementation, but it was weird and like that it should be easier to support arbitrary buttonstyles.

Highlighted state is animated via UIView.animate, without that it didn’t animate short presses at all
When disabled opacity is 50%

it registers as accessibility element and button + handles accessibility activate

### WinUIBackend
- implemented new required functions
- added `CustomButton` based on WinUI.Button

Bordered style is implemented by not doing anything. the other styles set everything to a transparent brush xD
The brush stuff should make support for future custom styles easy.

Borderless and plain buttons are 70% opacity when highlighted and 36% when disabled

### SwiftCrossUI Tests
Updated `SwiftCrossUITests/testBasicLayout()`
It now uses a mathematical approach that should theoretically work for all backends if needed.

1. For each string in the view a Text() is measured
2. The buttons add the backend’s button spacing to their width
3. the largest width is determined via max() on all widths
4. the total view height is calculated by adding all text heights, 2 times button padding, 2 times Vstack spacing and 2 times default padding
5. a viewsize is created from largest width + 2x padding and the total height
6. the viewsize is compared to the resulting viewsize of the full view by the layout engine.

### Info
The ._* gitignore addition has the purpose of ignoring sshfs generated intermediate/cache files. I don’t think it would currently conflict with something.

Thank you to @kaascevich for making me test on macOS 15, otherwise I would’ve missed my implementation previously being messed up there due to a breaking change in on the switch to macOS 26.
Thank you to @bbrk24 for helping me with making my Kotlin code more Kotlin-y.
And of course thank you to @stackotter for reviewing this big PR and discussing solutions.

[^1]:The solution and colors are not perfect. This can likely be improved once ButtonStyle is a protocol and maybe backends could do the resolution in the future so its not bound by deviceclass. On the apple backends it should be relatively close to native. Since their stuff isn’t open source its hard to get exactly right. And foregroundStyle is also not a publicly accessible property it looks like.


MiaKoring:
Supporting all that on AppKit alone took a surprisingly large amount of time 😅

Sources/AndroidBackend/Kotlin/CustomButton.kt
```diff
@@ -0,0 +1,137 @@
...
+    private val borderedBackground by lazy {
+        GradientDrawable().apply {
+            shape = GradientDrawable.RECTANGLE
+            setColor(getAdaptiveGray(context))
+            cornerRadius = 3f * resources.displayMetrics.density
+        }
+    }
+
+    class ButtonStyle {
+        companion object {
```
bbrk24: Could be `object ButtonStyle` if we don't need to create multiple instances of `ButtonStyle`.

Sources/AndroidBackend/Kotlin/CustomButton.kt
```diff
@@ -0,0 +1,137 @@
...
+    var buttonStyle: Int = ButtonStyle.BORDERED
+    var action: SwiftAction? = null
+
+    private val density = resources.displayMetrics.density
+
+    private val borderedDrawable by lazy {
+        val outValue = TypedValue()
+        activity.theme.resolveAttribute(android.R.attr.selectableItemBackground, outValue, true)
+        activity.getDrawable(outValue.resourceId)
+    }
```
bbrk24: Hm... I understand what you're doing but I don't love it organizationally. How about something like this?

```kt
companion object {
  private val BORDERED_ATTR = R.attr.selectableItemBackground
  private val BORDERLESS_ATTR = R.attr.selectableItemBackgroundBorderless

  private fun getDrawable(activity: Activity, attrId: Int): Drawable { ... }
}

// Then, further down
val foregroundAttr = when (buttonStyle) {
  ButtonStyle.PLAIN,
  ButtonStyle.BORDERLESS -> BORDERLESS_ATTR
  else -> BORDERED_ATTR
}
foreground = getDrawable(activity, foregroundAttr)
```

I believe you will have to re-fetch the drawable each time the root environment changes (even though we don't actually handle that yet).

Sources/AndroidBackend/Kotlin/CustomButton.kt
```diff
@@ -0,0 +1,137 @@
...
+    fun set(action: SwiftAction, buttonType: Int, isEnabled: Boolean) {
+        this.buttonStyle = buttonType
+        this.action = action
+        this.isEnabled = isEnabled
+
+        updateButtonStyle()
+    }
+
+    fun updateButtonStyle() {
+        alpha = if (isEnabled || buttonStyle == ButtonStyle.BORDERED) 1.0f else 0.38f
```
bbrk24: Where does this come from?

Sources/AndroidBackend/Kotlin/CustomButton.kt
```diff
@@ -0,0 +1,137 @@
...
+        frameParams.width = ViewGroup.LayoutParams.WRAP_CONTENT
+        frameParams.height = ViewGroup.LayoutParams.WRAP_CONTENT
+
+        super.addView(child, index, frameParams)
+    }
+
+    fun getAdaptiveGray(context: Context): Int {
+        val isDarkMode =
+            (context.resources.configuration.uiMode and Configuration.UI_MODE_NIGHT_MASK) ==
+                Configuration.UI_MODE_NIGHT_YES
```
bbrk24: We already have a function for this on AndroidBackendHelpers. Or, y'know, Swift could pass it in?

Sources/AndroidBackend/Kotlin/CustomButton.kt
```diff
@@ -0,0 +1,137 @@
...
+
+        return if (isDarkMode) {
+            Color.DKGRAY
+        } else {
+            Color.LTGRAY
+        }
```
bbrk24: Nitpick: Could be
```kt
return if (isDarkMode) Color.DKGRAY else Color.LTGRAY
```

Sources/AndroidBackend/AndroidBackend+CustomButton.swift
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    public func updateButton(
+        _ button: Widget,
+        environment: EnvironmentValues,
+        action: @escaping () -> Void
+    ) {
+        let button = button.as(CustomButton.self)!
+        button.set(
+            action: SwiftAction(action: action),
```
bbrk24: Forgot to pass the JNI env here?

Sources/AndroidBackend/CustomButton.swift
```diff
@@ -0,0 +1,19 @@
+import AndroidKit
+import SwiftJava
+
+@JavaClass(
+    "dev.swiftcrossui.androidbackend.CustomButton",
+    extends: AndroidKit.FrameLayout.self
+)
+class CustomButton: AndroidKit.FrameLayout {
+    static let horizontalPadding = 11
+    static let verticalPadding = 5
```
bbrk24: If you want to only define these once, instead of in both Swift and Kotlin,

In Kotlin:
```kt
class CustomButton /* ... */ {
  companion object {
    // either:
    @JvmField
    val horizontalPadding = 11
    // or:
    const val verticalPadding = 5
  }

  // ...
}
```
(`@JvmField` is a Kotlin built-in that requires no imports.)

In Swift:
```swift
extension JavaClass where JavaClass_T == CustomButton {
  @JavaStaticField(isFinal: true)
  var horizontalPadding: Int32

  @JavaStaticField(isFinal: true)
  var verticalPadding: Int32
}
```

The difference between a `val` and a `const val` is that the latter is inlined into other Kotlin code, while the former is always a JVM property access at runtime. `@JvmField` exposes the val on the companion object as a Java static field, but for some reason it's implied on const vals.

Examples/Sources/ControlsExample/ControlsApp.swift
```diff
@@ -49,19 +49,48 @@ struct ControlsApp: App {
...
     @State var pickerStyle: BuiltInPickerStyle? = .automatic
+    @State var buttonStyle: ButtonStyle? = nil
 
     @Environment(\.supportedDatePickerStyles) var supportedDatePickerStyles
     @Environment(\.isPickerStyleSupported) var isPickerStyleSupported
 
+    var defaultButtonStyle: ButtonStyle {
+        backend.defaultButtonStyle()
+    }
+
```
bbrk24: I feel like this should be exposed on the environment or something, if app code might need it.

Sources/SwiftCrossUI/Backend/BackendFeatures/Controls/Buttons.swift
bbrk24, line 1: I feel like these should be two separate protocols.

Sources/SwiftCrossUI/Values/ButtonStyle.swift
bbrk24, line 1: Do we want this to be a `public enum`? That would both make the rawValue public API and preclude us from adding more in the future. I think it should be a struct-wrapping-an-enum like DeviceClass.

bbrk24 (commented):
Only reviewed Android, plus a little bit of the backend-independent code.

Sources/AndroidBackend/AndroidBackend+CustomButton.swift
MiaKoring: apparently 😅

out of curiosity: why does it still work?

Sources/AndroidBackend/CustomButton.swift
MiaKoring: Would it be more expensive computationally due to jni?

Sources/AndroidBackend/AndroidBackend+CustomButton.swift
```diff
@@ -15,16 +15,17 @@ extension AndroidBackend {
     ) {
         let button = button.as(CustomButton.self)!
         button.set(
-            action: SwiftAction(action: action),
+            action: SwiftAction(environment: Self.env, action: action),
             buttonStyle: Int32(
-                (environment.buttonStyle ?? defaultButtonStyle()).rawValue
+                (environment.buttonStyle ?? defaultButtonStyle()).kind.rawValue
             ),
```
bbrk24: If AndroidBackend is the only one that uses the rawValue, we could make it `Int32` (or even `Int16`, and then have `Short` on the Kotlin side)

Examples/Sources/ControlsExample/ControlsApp.swift
```diff
@@ -85,6 +82,7 @@ struct ControlsApp: App {
                                             Color.purple.opacity(0.5)
                                                 .cornerRadius(5)
                                         }
+                                        .foregroundColor(.adaptive(light: .black, dark: .white))
```
bbrk24: Why is this necessary? Isn't that the default value of `environment.suggestedForegroundColor`?

Sources/AndroidBackend/AndroidBackend+CustomButton.swift
```diff
@@ -15,16 +15,17 @@ extension AndroidBackend {
...
-            isEnabled: environment.isEnabled
+            isEnabled: environment.isEnabled,
+            isDarkMode: environment.colorScheme == .dark
         )
     }
 
     public func buttonPadding(in environment: EnvironmentValues) -> SIMD2<Int> {
-        switch environment.buttonStyle ?? defaultButtonStyle() {
+        switch (environment.buttonStyle ?? defaultButtonStyle()).kind {
```
bbrk24: Maybe environment could have something for this, the same way it has foregroundColor and suggestedForegroundColor?

Sources/SwiftCrossUI/Backend/BackendFeatures/Controls/Buttons.swift
```diff
@@ -25,7 +25,13 @@ extension BackendFeatures {
             environment: EnvironmentValues,
             action: @escaping () -> Void
         )
+    }
 
+    /// Backend methods for more complex buttons supporting an arbitrary ``View`` as label.
+    ///
+    /// These are used by ``Button``.
+    @MainActor
+    public protocol Buttons: Core {
```
bbrk24: This name is slightly confusing. I was half expecting a `typealias Buttons = FirstProtocol & SecondProtocol`, the same way we have for e.g. `FileDialogs`.

Sources/WinUIBackend/WinUIBackend+Button.swift
```diff
@@ -196,7 +196,9 @@ extension ButtonStyle {
             case .plain, .borderless:
                 button.opacity = button.enabled
                     ? button.isHighlighted ? 0.7: 1.0
-                    : 0.36
+                    : 0.38
+                // Why 38% opacity was chosen (I couldn't find a Fluent guideline):
```
bbrk24: Did you check the XAML files? I've had to do that when writing WinUI code before.

Sources/AndroidBackend/CustomButton.swift
bbrk24: Probably. Not sure how noticeable it would be.

Examples/Sources/ControlsExample/ControlsApp.swift
MiaKoring: gets overriden on mobile due to blue being the default foreground in borderless and bordered

bbrk24: Right, I forgot iOS does that.

Sources/WinUIBackend/WinUIBackend+Button.swift
MiaKoring: I couldn't find something referencing an opacity change, because windows usually handles those highlighting cases in a different way

Sources/AndroidBackend/CustomButton.swift
MiaKoring: I guess we’ll apply don’t optimize until profiled here. From that viewpoint I can agree with making a jni call for that xDD

MiaKoring: ok, nope I’ll pass it to kotlin instead. 
I can’t create a second companion object and I don’t want to remove the name on `object ButtonStyle`

bbrk24 (commented):
Android code looks good to me, but I haven't reviewed the rest of it too closely.

Sources/SwiftCrossUI/Backend/BackendFeatures/Controls/Buttons.swift
```diff
@@ -19,11 +21,56 @@ extension BackendFeatures {
...
+
+        /// Buttons are set to label size + padding by SwiftCrossUI.
+        /// Backends may choose different amounts of padding for different button styles.
+        ///
+        /// - Parameters:
+        ///   - environment: The current environment.
+        ///
+        /// - Returns: A vector containing the **total** spacing horizontally and vertically.
+        ///
+        /// A padding of (0, 0) is recommended for all styles without a system defined background.
```
stackotter: Place this before the 'parameters' section;

```
/// Blah blah blah overview blah blah.
///
/// Discussion blah blah blah.
///
/// - Parameters:
///   ...
```

Sources/SwiftCrossUI/Backend/BackendFeatures/Controls/Buttons.swift
```diff
@@ -19,11 +21,56 @@ extension BackendFeatures {
...
+        func buttonPadding(in environment: EnvironmentValues) -> SIMD2<Int>
+
+        /// The default style of a button the backend expects, that's used when no ``ButtonStyle`` is set.
```
stackotter: `The default button style that the backend desires.`

Sources/SwiftCrossUI/Backend/BackendFeatures/Controls/Buttons.swift
```diff
@@ -19,11 +21,56 @@ extension BackendFeatures {
...
             action: @escaping () -> Void
         )
     }
+
+    /// Backend methods for more complex buttons supporting an arbitrary ``View`` as label.
+    ///
+    /// These are used by ``Button``.
+    @MainActor
+    public protocol ViewLabelButtons: Core {
+        /// Creates a button supporting ``View`` as label with an action triggered on click/tap.
```
stackotter: `Creates a button that uses `widget` as its label with an action triggered on click/tap.`

Sources/SwiftCrossUI/Values/ButtonStyle.swift
```diff
@@ -0,0 +1,30 @@
+/// A type that applies standard interaction behavior and a custom appearance to all buttons within a view hierarchy.
+public struct ButtonStyle: Hashable, Sendable {
+    // ALWAYS KEEP ORDER CONSISTENT with CustomButton.ButtonStyle in Kotlin.
+    package enum Kind: Int16 {
+        case bordered
+        case plain
+        case borderless
+    }
```
stackotter: I think we should avoid making this fragile. I saw @bbrk24 suggested something earlier which might be the same, but could we just add some static fields to CustomButton that defines the constant values corresponding to this Kind enum, so that AndroidBackend can just switch over `Kind` and retrieve the correct values from Kotlin? The JNI cost should be relatively minimal. Especially if we fetch the values when backend gets initialised rather than every time we update a button.

Sources/SwiftCrossUI/Views/Button.swift
```diff
@@ -1,74 +1,162 @@
...
+        _ label: String,
+        action: @escaping @MainActor @Sendable () -> Void = {}
+    ) where Content == Text {
+        self.body = Text(label)
         self.action = action
+        self.label = label
     }
 
-    /// A temporary button width solution until arbitrary labels are supported.
-    public func _buttonWidth(_ width: Int?) -> Button {
```
stackotter, line 21: Would we be able to keep this around as a deprecated method? Perhaps it can just modify the button to have a frame modifier around its label?

Maybe add buttonPadding as a computed environment property (which queries it using the backend method), and then make a small helper view here;

```swift
public func _buttonWidth(_ width: Int?) -> Button<some View> {
    struct ConstrainedButtonLabel<Label: View>: View {
        @Environment(\.buttonPadding.x) var horizontalPadding

        var content: Label
        var width: Int?
        
        var body: some View {
            content.ifLet(width) { view, width in
                view.frame(width: width - horizontalPadding)
            }
        }
    }
    
    return Button(action) { ConstrainedButtonLabel(content: body) }
}
```

Sources/SwiftCrossUI/Views/Button.swift
```diff
@@ -1,74 +1,162 @@
...
         layout: ViewLayoutResult,
         environment: EnvironmentValues,
         backend: Backend
     ) {
         backend.setSize(of: widget, to: layout.size.vector)
+        _ = children.child?.commit()
+    }
+}
+
+class ButtonLabelChild<Child: View>: ViewGraphNodeChildren {
```
stackotter: Can we use TupleViewChildren1 instead of this custom children type?

Sources/SwiftCrossUI/Views/Button.swift
```diff
@@ -1,74 +1,162 @@
...
-    public func commit<Backend: BaseAppBackend>(
+    func commit<Backend: BaseAppBackend>(
         _ widget: Backend.Widget,
+        children: Children,
         layout: ViewLayoutResult,
         environment: EnvironmentValues,
         backend: Backend
     ) {
         backend.setSize(of: widget, to: layout.size.vector)
+        _ = children.child?.commit()
```
stackotter: We generally commit the child first. I think it's just out of habit, but could be best to stay consistent to avoid surprise issues

Sources/SwiftCrossUI/Views/Button.swift
```diff
@@ -1,74 +1,162 @@
...
+                )
+            } else if backend.deviceClass == .tv {
+                childEnvironment = childEnvironment.with(
+                    \.foregroundColor,
+                    environment.foregroundColor ?? .white.opacity(0.5)
+                )
+            }
+        }
+        // The disabled opacities and defaults are based on discoveries in SwiftUI.
+        // iOS appears to dimm even foregroundColors in the environment.
```
stackotter: Quite a few of these defaults feel like idioms taken from Apple platforms, rather than things that SwiftCrossUI is wanting to enforce across all non-Apple platforms as well. E.g. AndroidBackend running on a TV may need different bordered button foreground colour behaviour.

We should make it possible for backends to override this behaviour. It's not possible with this current implementation because we always make the foreground colour non-nil, making it impossible for backends to separate user choice from SCUI defaults.

I believe that the solution is to move all of this foreground colour logic to a `suggestedButtonForegroundColor` computed property of `EnvironmentValues`, similar to what I've already done in the past with suggestedForegroundColor for text (which was done to allow iOS buttons to retain their blue fg colour when not explicitly overridden).

Sources/SwiftCrossUI/Views/Button.swift
```diff
@@ -1,74 +1,162 @@
 /// A control that initiates an action.
-public struct Button: Sendable {
+public struct Button<Content: View> {
     /// The label to show on the button.
-    package var label: String
+    public var body: Content
```
stackotter: We should store the body as a closure rather than precomputing it. That will help avoid annoying bugs when we begin reducing the number of body recomputations we perform.

Sources/SwiftCrossUI/Views/Button.swift
```diff
@@ -1,74 +1,162 @@
...
         environment: EnvironmentValues,
         backend: Backend
     ) {
         backend.setSize(of: widget, to: layout.size.vector)
+        _ = children.child?.commit()
+    }
+}
+
+class ButtonLabelChild<Child: View>: ViewGraphNodeChildren {
+    var child: AnyViewGraphNode<Child>?
```
stackotter: If we do have to keep this this custom Children class, then make this property non-optional. I don't think there's a reason for it to be optional because we always set it immediately (so we might as well take it as an initialiser parameter)

Sources/UIKitBackend/KeyboardToolbar.swift
stackotter: if we add `where Content == Text`, can we eliminate `Button`'s `label` property and instead use `.body.string` within this implementation?

Sources/SwiftCrossUI/Views/Button.swift
```diff
@@ -1,74 +1,162 @@
 /// A control that initiates an action.
-public struct Button: Sendable {
+public struct Button<Content: View> {
```
stackotter: We should name the type parameter `Label` instead of `Content` to match SwiftUI (and make things a bit more intuitive).

Sources/SwiftCrossUI/Environment/EnvironmentValues.swift
```diff
@@ -444,6 +444,22 @@ extension EnvironmentValues {
...
+    @Entry public var buttonStyle: ButtonStyle?
+
+    /// The default button style as declared by the backend.
+    @MainActor
+    public var defaultButtonStyle: ButtonStyle {
+        backend.defaultButtonStyle()
+    }
+
+    /// The resolved ``ButtonStyle`` for backends to use.
+    /// Either the set buttonStyle in environment or the ``defaultButtonStyle`` if nil.
```
stackotter:
```
/// The resolved ``ButtonStyle``. Either ``buttonStyle``, or ``defaultButtonStyle`` if nil.
```

Sources/GtkBackend/GtkBackend+Button.swift
```diff
@@ -0,0 +1,137 @@
...
+        // (switch to the variable when we have adwaita)
+
+        // Why 80% for active(pressed) was chosen:
+        // A pressed SwiftUI .plain button looks visually the same as
+        // a not pressed one at 0.8 opacity.
+    }
+}
+
+extension ButtonStyle.Kind {
+    fileprivate func setClass(_ button: GtkCustomButton) {
```
stackotter: label the parameter `from` to make callsites more clear

Sources/GtkBackend/GtkBackend+Button.swift
```diff
@@ -0,0 +1,137 @@
...
+        if let cssClass {
+            gtk_widget_add_css_class(button.widgetPointer, cssClass)
+        }
+    }
+
+    fileprivate func removeClass(_ button: GtkCustomButton) {
```
stackotter: label the parameter `from` to make callsites more clear

Sources/GtkBackend/GtkBackend+Button.swift
```diff
@@ -0,0 +1,137 @@
...
+        button.css.clear()
+        button.css.set(properties: Self.cssProperties(for: environment, isControl: true))
+    }
+
+    public func createButton(wrapping widget: Widget) -> Widget {
+        let button = GtkCustomButton()
+        gtk_button_set_child(button.widgetPointer.cast(), widget.widgetPointer)
+
+        gtk_widget_set_halign(widget.widgetPointer, GTK_ALIGN_CENTER)
+        gtk_widget_set_valign(widget.widgetPointer, GTK_ALIGN_CENTER)
```
stackotter: Use `horizontalAlignment` and `verticalAlignment` properties of `widget` instead of the C functions

Sources/GtkBackend/GtkBackend+Button.swift
```diff
@@ -0,0 +1,137 @@
...
+    ) {
+        let button = button as! GtkCustomButton
+        button.clicked = { _ in action() }
+        button.buttonStyle = environment.resolvedButtonStyle.kind
+        button.sensitive = environment.isEnabled
+    }
+
+    public func buttonPadding(in environment: EnvironmentValues) -> SIMD2<Int> {
+        switch environment.resolvedButtonStyle.kind {
+            case .bordered: SIMD2<Int>(
```
stackotter: Place the `SIMD...` onto its own line

Sources/GtkBackend/GtkBackend+Button.swift
```diff
@@ -0,0 +1,137 @@
...
+    fileprivate func removeClass(_ button: GtkCustomButton) {
+        if let cssClass {
+            gtk_widget_remove_css_class(button.widgetPointer, cssClass)
+        }
+    }
+
+    var cssClass: String? {
+        switch self {
+            case .bordered: nil
+            case .plain, .borderless: "flat"
```
stackotter, line 170: Why is this class called `flat`, is it a Gtk term? To me it sounds like it'd still have a background but without shading (e.g. 'flat design'). `borderless` is the best term for that style of button in my opinion.

Sources/Gtk3Backend/Gtk3Backend+Button.swift
```diff
@@ -0,0 +1,145 @@
+import Gtk3
```
stackotter, line 1: Apply all of the GtkBackend feedback to this file as well. I believe it should all still be applicable because this code looks basically identical.

Sources/DummyBackend/DummyBackend.swift
stackotter: Rename this to `SimpleButton`

Sources/DummyBackend/DummyBackend.swift
```diff
@@ -78,6 +78,12 @@ public final class DummyBackend:
         public var action: (() -> Void)?
     }
 
+    public class CustomButton: Widget {
```
stackotter: Rename this to `Button`

Tests/SwiftCrossUITests/SwiftCrossUITests.swift
```diff
@@ -186,20 +186,60 @@ struct SwiftCrossUITests {
...
             backend.setChild(ofWindow: window, to: viewGraph.rootNode.widget.into())
 
-            let result = viewGraph.computeLayout(
+            let contentViewResult = viewGraph.computeLayout(
                 proposedSize: ProposedViewSize(200, 200),
                 environment: environment
             )
 
-            #expect(
-                result.size == ViewSize(92, 96),
```
stackotter, line 195: Do you know why this view size changed? The main point of this test was to catch if the behaviour changed from the known correct size. Using the backend itself to compute this value makes this test less likely to fail when the backend has a bug (and moreso becomes a test of SCUI's layout logic).

If buttons changed size then we should investigate and make sure we restore whatever the original sizing was.

Sources/AndroidBackend/AndroidBackend+CustomButton.swift
```diff
@@ -0,0 +1,40 @@
...
+            buttonStyle: environment.resolvedButtonStyle.kind.rawValue,
+            isEnabled: environment.isEnabled,
+            isDarkMode: environment.colorScheme == .dark
+        )
+    }
+
+    public func buttonPadding(in environment: EnvironmentValues) -> SIMD2<Int> {
+        let buttonClass = try! JavaClass<CustomButton>()
+        return switch environment.resolvedButtonStyle.kind {
+            case .bordered: SIMD2(
```
stackotter: Put `SIMD...` on its own line

stackotter (changes requested):
Thank you for all your hard work on this one over the past week. A lot of this looks like it was quite tedious to figure out.

And for the most part, the SCUI side of things looks pretty good! I've reviewed everything other than the custom button implementations for the backends that you've implemented custom non-native buttons for. I have looked over those implementations, but my feedback isn't something that I can attach to any single piece of code. The rest of this comment is my general feedback that applies to all of those custom non-native button implementations;

---

As I see it, re-implementing native buttons ourselves has a number of downsides;

1. Advanced users using the '.inspect' modifiers basically can't work with buttons at all any more
2. We take on a much higher risk of bugs than if we can reuse the native buttons, and as such we also take on a much higher maintenance burden (to keep styles, layout, and accessibility attributes up-to-date with each platform's latest conventions)
3. It also introduces a lot more effort if we'd like to have our buttons naturally adapt to different versions of the same platform. E.g. it would now take significant effort for us to implement liquid glass buttons, and it would also be non-trivial to verify that such an implementation acts correctly (perhaps liquid glass buttons have different reactions when pressed etc)
4. On a high level it means that our buttons are simply not native any more, which people are bound to have complaints about eventually

If it's only for one or two platforms, then I'm happy to live with that, but at the moment I believe that the Gtk backends are the only backends that defers to the native implementation for interactions, base styling, and accessibility.

I see a couple of ways to deal with this;

1. We can reduce the impact of the additional risk of bugs by using a proper native button when the button has a simple text label, and only deferring to these custom non-native buttons when the label isn't `Text`
2. We should aim to replace these custom implementations with native-backed implementations wherever possible

I understand that many of the backends have proven quite difficult in terms of reusing native buttons. I think there are some tricks we can use, but I do realise that none of this feedback is simple to implement in practice, and some backends may always require a non-native button implementation in some cases.

With my first solution implemented, I would feel much more comfortable merging in the non-native button polyfills and leaving it as a future job to slowly replace as much of those as we can.

Sources/GtkBackend/GtkBackend+Button.swift
MiaKoring: flat is a gtk thing, yes

Sources/SwiftCrossUI/Values/ButtonStyle.swift
bbrk24: Not that it needs to happen in this PR, since it's already large enough, but whatever approach we decide on here should probably also be applied to DeviceClass (since currently that uses separately-hardcoded constants in Swift and Kotlin).

MiaKoring: I will do the same as with the padding here for now.

Sources/SwiftCrossUI/Views/Button.swift
MiaKoring:
```swift
public func _buttonWidth(_ width: Int?) -> Button<some View> {
    return Button<TupleView1<ConstrainedButtonLabel<TupleView1<Label>>>>(
        action: action,
        label: { ConstrainedButtonLabel(content: body, width: width) }
    )
}
```
if I declare the exact type of the button to get the new button out of the Label constraint, it works.

And the struct cannot be nested in the generic function as the compiler told me.

Sources/SwiftCrossUI/Views/Button.swift
MiaKoring: yes we can and yes we do (now) :P

Sources/SwiftCrossUI/Views/Button.swift
MiaKoring: I’m not sure suggestedButtonForegroundColor is the right way to do it, as it wouldn’t affect the label views unless its set as foregroundColor, which would kind of defeat the backend modifiability again.

I would like to propose doing `BackendFeatures.ViewLabelButtons/buttonLabelEnvironment(from environment: EnvironmentValues)` instead, with that handling as a default implementation.

That would also allow for more custom handling than just color for backends with more different conventions.

Sources/GtkBackend/GtkBackend+Button.swift
MiaKoring: I decided to use on here, because from is more of a removing thing imo

Sources/GtkBackend/GtkBackend+Button.swift
MiaKoring: https://docs.gtk.org/gtk3/const.STYLE_CLASS_FLAT.html or
https://github.com/GNOME/gtk/blob/abceb4e7432d1b80bfdb1fdde7170d67b300069b/gtk/theme/Default/_common.scss#L567

Sources/SwiftCrossUI/Backend/BackendFeatures/Controls/Buttons.swift
```diff
@@ -72,5 +72,71 @@ extension BackendFeatures {
...
+        /// The default implementation applies SwiftUI-like behavior.
+        ///
+        /// - Returns: The modified environment.
+        func buttonLabelEnvironment(
+            from environment: EnvironmentValues
+        ) -> EnvironmentValues
+    }
+}
+
+extension BackendFeatures.ViewLabelButtons {
```
kaascevich, line 89: Maybe add a `// MARK: - Default Implementations\n` comment above here to match the other files with default impls? Not really necessary though.

Sources/SwiftCrossUI/Backend/BackendFeatures/Controls/Buttons.swift
```diff
@@ -60,17 +60,84 @@ extension BackendFeatures {
...
         /// - Returns: The default ``ButtonStyle``.
         func defaultButtonStyle() -> ButtonStyle
+
+        /// Modifies the environment for the body of a button label.
+        ///
+        /// Backends may implement their own to align with the platform's conventions more closely.
+        /// The default implementation applies SwiftUI-like behavior.
+        ///
+        /// - Returns: The modified environment.
+        func buttonLabelEnvironment(
```
stackotter: Name this `computeButtonLabelEnvironment(from:)` to match the naming of other similar methods (e.g. `computeRootEnvironment`)

Sources/SwiftCrossUI/Backend/BackendFeatures/Controls/Buttons.swift
```diff
@@ -60,17 +60,84 @@ extension BackendFeatures {
...
+                labelEnvironment = labelEnvironment.with(
+                    \.foregroundColor,
+                    environment.foregroundColor?.opacity(0.8) ?? .blue.opacity(0.8)
+                )
+            } else if deviceClass == .tv {
+                labelEnvironment = labelEnvironment.with(
+                    \.foregroundColor,
+                    environment.foregroundColor ?? .white.opacity(0.5)
+                )
+            }
```
stackotter: TV is handled twice. The last branch of the if will never be taken afaict

Sources/SwiftCrossUI/Backend/BackendFeatures/Controls/Buttons.swift
```diff
@@ -60,17 +60,84 @@ extension BackendFeatures {
...
+                labelEnvironment = labelEnvironment.with(
+                    \.foregroundColor,
+                    environment.foregroundColor ?? .gray
+                        .opacity(0.5) // SwiftUI is closer to secondary.
+                )
+            } else if deviceClass == .phone || deviceClass == .tablet {
+                labelEnvironment = labelEnvironment.with(
+                    \.foregroundColor,
+                    environment.foregroundColor?.opacity(0.8) ?? .blue.opacity(0.8)
+                )
```
stackotter: This iOS logic (in particular, everything that handles `blue`) is very iOS-specific. IMO it's not mobile-specific, so it should go in UIKitBackend's implementation of this method. E.g. AndroidBackend shouldn't have to override it back again if it doesn't want blue iOS buttons. The default should probably just be the desktop styling because it's very sensible and mostly looks good on each platform.

Sources/SwiftCrossUI/Backend/BackendFeatures/Controls/Buttons.swift
```diff
@@ -60,17 +60,84 @@ extension BackendFeatures {
...
+                environment.foregroundColor ?? .gray
+            )
+        } else if
+            deviceClass == .phone || deviceClass == .tablet,
+            buttonStyle == .borderless || buttonStyle == .bordered
+        {
+            labelEnvironment = labelEnvironment.with(
+                \.foregroundColor,
+                environment.foregroundColor ?? .blue // TODO: Replace with .accent
+            )
```
stackotter: This iOS-specific logic should also be moved to UIKitBackend's implementation of this method

Sources/SwiftCrossUI/Backend/BackendFeatures/Controls/Buttons.swift
```diff
@@ -60,17 +60,84 @@ extension BackendFeatures {
...
+            } else if deviceClass == .tv {
+                labelEnvironment = labelEnvironment.with(
+                    \.foregroundColor,
+                    environment.foregroundColor ?? .white.opacity(0.5)
+                )
+            }
+        }
+
+        // The disabled opacities and defaults are based on discoveries in SwiftUI.
+        // iOS appears to dimm even foregroundColors in the environment.
```
stackotter: `dim`

Sources/SwiftCrossUI/Backend/BackendFeatures/Controls/Buttons.swift
```diff
@@ -60,17 +60,84 @@ extension BackendFeatures {
...
+
+        let buttonStyle = environment.resolvedButtonStyle.kind
+        let deviceClass = environment.backend.deviceClass
+
+        if !environment.isEnabled, buttonStyle == .bordered {
+            if deviceClass == .desktop || deviceClass == .tv {
+                labelEnvironment = labelEnvironment.with(
+                    \.foregroundColor,
+                    environment.foregroundColor ?? .gray
+                        .opacity(0.5) // SwiftUI is closer to secondary.
```
stackotter: Sorry if I've asked this already, but if I have then I couldn't find back where I did. Is this gray meant to be the same in both light and dark mode?

Sources/SwiftCrossUI/Views/Button.swift
```diff
@@ -15,10 +14,9 @@ public struct Button<Content: View> {
     public init(
         _ label: String,
         action: @escaping @MainActor @Sendable () -> Void = {}
-    ) where Content == Text {
-        self.body = Text(label)
+    ) where Label == TupleView1<Text> {
```
stackotter, line 17: This should be `Label == TupleView1`. I believe that you're ending up with two layers of tuple views at the moment.

Sources/SwiftCrossUI/Environment/EnvironmentValues.swift
```diff
@@ -454,12 +454,16 @@ extension EnvironmentValues {
...
-    /// The resolved ``ButtonStyle`` for backends to use.
-    /// Either the set buttonStyle in environment or the ``defaultButtonStyle`` if nil.
+    /// The resolved ``ButtonStyle``. Either ``buttonStyle``, or ``defaultButtonStyle`` if nil.
     @MainActor
     public var resolvedButtonStyle: ButtonStyle {
         buttonStyle ?? defaultButtonStyle
     }
+
+    @MainActor
+    public var buttonPadding: SIMD2<Int> {
```
stackotter, line 465: Given that this is public, I think we should make the name a bit more clear. Perhaps 'borderedButtonPadding'?

Also add some documentation;

```swift
/// The amount of padding that the current backend applies to the labels of bordered buttons.
```

Sources/SwiftCrossUI/Builders/AlertActionsBuilder.swift
```diff
@@ -21,13 +22,14 @@ public struct AlertActionsBuilder {
         first.actions
     }
 
+    @MainActor
     public static func buildPartialBlock(
         accumulated: [AlertAction],
-        next: Button<Text>
+        next: Button<TupleView1<Text>>
```
stackotter, line 28: With some of my other suggested changes, this should become `Button<Text>` again

Sources/UIKitBackend/KeyboardToolbar.swift
```diff
@@ -64,7 +64,7 @@ public enum ToolbarBuilder {
 
 @available(tvOS, unavailable)
 @available(visionOS, unavailable)
-extension Button: ToolbarItem {
+extension Button: ToolbarItem where Label == TupleView1<Text> {
```
stackotter, line 67: This should become `Label == Text` with my other suggested changes

Sources/GtkBackend/GtkBackend+Button.swift
stackotter: I don't know why I said `from` lol, I think I must've blindly copied from my other feedback, but I would've meant `of` or `on`.

Sources/WinUIBackend/WinUIBackend+Button.swift
stackotter: Here's where they set the background colour for a disabled button: https://github.com/microsoft/microsoft-ui-xaml/blob/fc2f821173298e8130fb5b143373ba70793bc251/src/controls/dev/CommonStyles/Common_themeresources_any.xaml#L38-L39

There should be other similar values for the other button states, and the foreground vs the background. We may be able to read those resource values manually at run time. We override similar resources at around line 160 of WinUIBackend. There may be a way to adapt that approach to read resource values.

Alternatively, you can just manually find the right values for each theme and manually put them in your implementation (linking back to the source for each value).

Sources/AppKitBackend/AppKitBackend+Button.swift
```diff
@@ -52,25 +51,44 @@ extension AppKitBackend {
...
     }
 
     public func defaultButtonStyle() -> ButtonStyle {
         .bordered
     }
+
+    func measureBorderedButtonPadding() -> SIMD2<Int> {
+        if let borderedButtonPadding { return borderedButtonPadding }
+
+        let testString = "Test"
```
stackotter: We should probably make this a single character because there's not a reason to have it longer. I know it shouldn't make a noticeable performance difference, but I think we should just do it out of principle.

Sources/SwiftCrossUI/Views/MenuItem.swift
```diff
@@ -1,7 +1,7 @@
 /// An item of a ``Menu`` or ``CommandMenu``.
 public enum MenuItem {
     /// A button.
-    case button(Button<Text>)
+    case button(Button<TupleView1<Text>>)
```
stackotter, line 4: This should become `Button<Text>` with my other suggested changes

Sources/SwiftCrossUI/Views/MenuItem.swift
```diff
@@ -28,7 +28,7 @@ protocol MenuItemRepresentable: View {
     nonisolated var asMenuItem: MenuItem { get }
 }
 
-extension Button: MenuItemRepresentable where Content == Text {
+extension Button: MenuItemRepresentable where Label == TupleView1<Text> {
```
stackotter: Will become `Label == Text`

stackotter (changes requested):
I've reviewed the SCUI code and some of the backend-specific code. All of your feedback resolutions were great. You'll be glad to know that most of the changes that I've requested this time are small changes/refactors rather than the huge feedback I had last time 😅

I will review the backend-specific implementations separately after lunch.

Sources/UIKitBackend/UIKitBackend+CustomButton.swift
```diff
@@ -0,0 +1,193 @@
...
+            case .plain, .borderless:
+                // The borderless special treatment is handled in environment.
+                button.configuration = .plain()
+        }
+    }
+
+    fileprivate func applyModifications(_ button: UICustomButton) {
+        guard #available(iOS 15.0, tvOS 15.0, macCatalyst 15.0, *) else {
+            button.label.alpha = button.isEnabled
+                ? button.isHighlighted ? 0.8: 1.0
```
stackotter: Is this how swiftformat formats things? With no space before the colon?

Sources/SwiftCrossUI/Views/Button.swift
```diff
@@ -1,74 +1,129 @@
...
-        //   `finalContentSize != contentSize` check in WindowGroupNode to catch any weird
-        //   behaviour. Without that extra safety net logic, buttons all end up label-less
-        //   whenever the window grows due to a view containing buttons appearing. Not sure
-        //   why all buttons lose their labels (until you click off the window, forcing it to
-        //   refresh), but the reason Gtk 3 doesn't like it is that the window gets set smaller
-        //   than its content I think.
-        //   See: https://github.com/moreSwift/swift-cross-ui/blob/27f50579c52e79323c3c368512d37e95af576c25/Sources/SwiftCrossUI/Scenes/WindowGroupNode.swift#L140
+        var childEnvironment = backend.buttonLabelEnvironment(from: environment)
+
+        let childrenResult = children.child0.computeLayout(
```
stackotter: Name this `childResult`

stackotter:
## macOS issues

### Double button VStack crash

The following snippet crashes

```swift
import DefaultBackend
import SwiftCrossUI

@main
struct Example: App {
    var body: some Scene {
        WindowGroup("Example") {
            Button { } label: {
                Color.blue
            }

            Button { } label: {
                Color.blue
            }
        }
    }
}
```

```
info: Running 'CounterExample.app'
Swift/arm64e-apple-macos.swiftinterface:39429: Fatal error: Double value cannot be converted to Int because it is either infinite or NaN
```

The `childResult` is `ViewSize(width: 900.0, height: inf)`. I'm not sure why the height is coming out infinite though.

## Disabled foreground colour too dark

```swift
import DefaultBackend
import SwiftCrossUI

@main
struct Example: App {
    var body: some Scene {
        WindowGroup("Example") {
            Button { } label: {
                ZStack {
                    Color.blue
                    Text("Hello")
                }
            }
            .background(Color.red)
            .frame(width: 200, height: 100)
            .disabled()

            Button { } label: {
                ZStack {
                    Color.blue
                    Text("Hello")
                }
            }
            .background(Color.red)
            .frame(width: 200, height: 100)
        }
    }
}
```

On the left if SwiftCrossUI, and on the right is SwiftUI.

Screenshot 2026-06-08 at 4 13 35 pm

### Button sizing issue

In the previous example, you can see that the SwiftCrossUI buttons are too big. The issue appears to be that you're not subtracting the button padding before passing the proposed size on to the label's `computeLayout` method. You can look at the `padding` modifier's implementation for how to do that.

## iOS issues

My testing is on iOS 18, I'm not sure how this differs on iOS 26.

### Label opacity in various states

This is using the same code as the `Disabled foreground colour too dark` macOS issue.

Here's the layout in SwiftUI. On the left is regular, and on the right is with the bottom button pressed.

<p float="left">
IMG_9932
IMG_9931
</p>

Here's the same layout in SwiftCrossUI. On the left is regular, and on the right is with the bottom button pressed. By comparing these screenshots to the SwiftUI ones above, we can see that the button labels aren't meant to become transparent when the button is disabled, but they are meant to become significantly more transparent when pressed.

<p float="left">
IMG_9930
IMG_9929
</p>

### Bordered buttons on iOS 18

Here's the same layout again but with SwiftUI. The image on the right has the bottom button pressed. If you zoom in you can see that the label becomes visible when pressed (even though the text is the same colour as the `Color.blue` rectangle). That indicates that the default foreground colour of the button changes when pressed.

<p float="left">
IMG_9935
IMG_9936
</p>

Here's the same layout as above but with the bordered style and no red background, in SwiftCrossUI. The image on the right has the bottom button pressed. If you compare to the SwiftUI version above, you can see that the disabled buttons are significantly darker in SwiftCrossUI.

<p float="left">
IMG_9933
IMG_9934
</p>

### Observations

- Pressing a bordered button shouldn't make its label transparent
- Pressing a bordered button changes the default foreground color
- Bordered buttons have a much higher corner radius in SwiftUI (this one isn't as high priority for now if it's not our fault; we can add special logic to match SwiftUI's style better later)
- Pressing a non-bordered button makes the label way more transparent than SCUI is currently making it

---

I apologise if any of that comes off harsh, I'm not intending it that way of course. I'm just dumping the results of my local testing for future reference.

Sources/SwiftCrossUI/Views/Button.swift
```diff
@@ -1,74 +1,129 @@
...
+            with: body.view0,
+            proposedSize: proposedSize,
```
stackotter: Subtract button padding before computing the label's layout

Sources/UIKitBackend/UIKitBackend+CustomButton.swift
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    public var onTap: (() -> Void)?
+
+    public var buttonStyle: ButtonStyle.Kind = .borderless
+
+    override public var isHighlighted: Bool {
+        didSet {
+            UIView.animate(
+                withDuration: 0.1,
```
stackotter: It seems (from my local testing) that the duration of the animation is meant to be longer when releasing than when pressing the button

Sources/AndroidBackend/Kotlin/CustomButton.kt
```diff
@@ -0,0 +1,135 @@
...
+        // Why 38% opacity was chosen:
+        // https://m2.material.io/design/interaction/states.html#disabled
+        alpha = if (isEnabled) 1.0f else 0.38f
+
+        val borderedBackground by lazy {
+            GradientDrawable().apply {
+                shape = GradientDrawable.RECTANGLE
+                setColor(getAdaptiveGray(isDarkMode))
+                cornerRadius = 3f * resources.displayMetrics.density
+            }
```
stackotter, line 71: Where do these values come from? Is there a way that we can rely on the default button background rendering?

Sources/AndroidBackend/Kotlin/CustomButton.kt
```diff
@@ -0,0 +1,135 @@
...
+                ButtonStyle.PLAIN,
+                ButtonStyle.BORDERLESS -> BORDERLESS_ATTR
+                else -> BORDERED_ATTR
+            }
+
+        if (isEnabled) {
+            foreground = getDrawable(context, foregroundAttr)
+        } else {
+            foreground = null
+        }
```
stackotter, line 115: What is this drawable? What is the foreground of the button that Android has to render? It feels more like this is rendering a background, so if this is what we're meant to be doing I think it should have a comment explaining why Android wants a background resource as a foreground, and why it's only needed when enabled

Sources/AppKitBackend/AppKitBackend+Button.swift
```diff
@@ -0,0 +1,326 @@
...
+        // The ViewBuilder button init is not covered by this current solution.
+        (subviews.first as? NSTextField)?.stringValue
+    }
+
+    override public func draw(_ dirtyRect: NSRect) {
+        super.draw(dirtyRect)
+    }
+
+    override public var acceptsFirstResponder: Bool {
+        // Even though its called FullKeyboardAccess, it actuall corresponds to
```
stackotter: `it's` and `actually`

Sources/AppKitBackend/AppKitBackend+Button.swift
```diff
@@ -0,0 +1,326 @@
...
+        isHighlighted = true
+        action?()
+
+        // Task with Task.sleep could be used in the future,
+        // it has a min version requirement of macOS 13.
+        let workItem = DispatchWorkItem { [weak self] in
+            self?.isHighlighted = false
+        }
+        highlightResetWorkItem = workItem
+        DispatchQueue.main.asyncAfter(deadline: .now() + 0.1, execute: workItem)
```
stackotter, line 187: Add a comment explaining how 0.1 was chosen (whether it was an estimate or measured etc)

Sources/AppKitBackend/AppKitBackend+Button.swift
```diff
@@ -0,0 +1,326 @@
...
+        button.button.isHidden = true
+        switch self {
+            case .bordered:
+                button.button.isHidden = false
+                button.button.isEnabled = button.isEnabled
+                button.button.isHighlighted = button.isHighlighted
+            case .plain, .borderless:
+                button.alphaValue = button.isEnabled
+                    ? button.isHighlighted ? 0.80: 1.0
+                    : 0.5
```
stackotter, line 298: As far as I can tell, the opacity of a borderless button doesn't change based of being enabled or highlighted. But it does for a plain button (where your opacity values appear to match pretty closely). A plain button's foreground seems to be white by default (rather than the gray of a borderless button).

Both of these buttons have the same background colour, even when the second one is pressed. The only difference seems to be the foreground colour changing from some sort of gray, to white.

```swift
Button { } label: {
    ZStack {
        Color.blue
        Text("Hello")
    }
}
.background(Color.red)
.frame(width: 200, height: 100)
.disabled(true)
.buttonStyle(.borderless)

Button { print("hi") } label: {
    ZStack {
        Color.blue
        Text("Hello")
    }
}
.background(Color.red)
.frame(width: 200, height: 100)
.buttonStyle(.borderless)
```

Sources/WinUIBackend/WinUIBackend+Button.swift
```diff
@@ -0,0 +1,216 @@
...
+        horizontalContentAlignment = HorizontalAlignment.center
+        verticalContentAlignment = VerticalAlignment.center
+
+        click.addHandler { [weak self] _, _ in
+            guard let self else { return }
+            self.action?()
+        }
+    }
+
+    override func onPointerPressed(_ e: PointerRoutedEventArgs!) throws {
```
stackotter, line 100: Rename these `e` parameters to `event`

Sources/WinUIBackend/WinUIBackend+Button.swift
stackotter: I found the button text colour disabled style ([here](https://github.com/microsoft/microsoft-ui-xaml/blob/fc2f821173298e8130fb5b143373ba70793bc251/src/controls/dev/CommonStyles/Common_themeresources_any.xaml#L8)) and it seems like your guess was pretty close! The true value appears to be 36.5% (rounded).

In a similar manner to the other backends, I believe that we should only change the opacity of the foreground colour itself, rather than of the whole button label (except in plain buttons where SwiftUI does seem to modify the opacity of the whole label).

stackotter (changes requested):
I've now looked through all of the code. The only thing left for me to do is testing on Windows, Android, and Linux, but I will likely have to leave that for another day, because I want to finish off the update video tonight.

Sources/SwiftCrossUI/Backend/BackendFeatures/Controls/Buttons.swift
MiaKoring: the difference in looks originates from our adaptive gray being different than swiftuis. SwiftUI appears to match their gray in both dark an light.

Sources/SwiftCrossUI/Views/Button.swift
MiaKoring: As discussed on discord, we will keep it for now.
reason:
```swift
print(type(of: Button("Test"){}))
print(type(of: Button {} label: { Text("Test") }))
```
Button<TupleView1<Text>>
Button<TupleView1<Text>>

Sources/SwiftCrossUI/Environment/EnvironmentValues.swift
MiaKoring: no, its not just for bordered, but the current button style in the environment

MiaKoring: /// The amount of padding that the current backend applies to the labels of buttons with the current ``ButtonStyle``. Is that ok?

Sources/UIKitBackend/UIKitBackend+CustomButton.swift
MiaKoring: not sure, I always do it that way. I can change it and if it disappears we’ll know why

Sources/AndroidBackend/Kotlin/CustomButton.kt
MiaKoring: I didn’t find one, the drawables are something completely different. once we use material it should be a yes though

Sources/AndroidBackend/Kotlin/CustomButton.kt
MiaKoring: this is where we get the ripple effect from. when disabled we obviously don’t want it to react to taps.

Sources/AppKitBackend/AppKitBackend+Button.swift
MiaKoring: The issue with that is, that we currently can’t change the environment for being pressed. Another issue is, that a borderless button would have 0 press effect when a foreground color is set in the environment. We *could*  attempt to mimic SwiftUI there, but honestly I would prefer a more sane way. 

same is true for disabled.

MiaKoring:
I made some improvements to the rendering issues you mentioned. 
The foreground color change on press I would honestly prefer to do as a future improvement. Same for some improvements to feeling even correcter on other platforms (which would involve again a lot more research and on Android likely Material)

The foreground color change will have to trigger a re-layout of the label and have the state of being pressed somewhere in scui.

I honestly need a break from those stupid buttons. I didn’t expect them to take that much time and energy when I started. And I’d say its already a huge improvement to before.

MiaKoring:
Also it might make sense to make ButtonStyle a protocol early on. That could potentially also help with making highly specific adjustments without getting too convoluted. This is not something I would set as a goal of this PR though.

Also the .opacity modifier might be needed to properly do protocol button style, unless its a method on the buttonstyle for now.

Sources/SwiftCrossUI/Backend/BackendFeatures/Controls/Buttons.swift
```diff
@@ -79,62 +79,45 @@ extension BackendFeatures {
...
         /// - Returns: The modified environment.
-        func buttonLabelEnvironment(
+        func computeButtonLabelEnvironment(
             from environment: EnvironmentValues
         ) -> EnvironmentValues
     }
 }
 
-// MARK: - Default Implementations
+// MARK: - Default Implementationsv
```
stackotter: Accidental `v` on the end there

Sources/SwiftCrossUI/Backend/BackendFeatures/Controls/Buttons.swift
```diff
@@ -79,62 +79,45 @@ extension BackendFeatures {
...
-        // iOS appears to dimm even foregroundColors in the environment.
 
         // Set the default foregroundColor for the label unless overridden.
         // Uses the same colors as SwiftUI.
         if
             buttonStyle == .borderless,
             deviceClass == .desktop
         {
+            // Approximately equivalent to Color.secondary in SwiftUI.
+            let opacity = environment.colorScheme == .dark ? 0.7: 0.5
```
stackotter: Put a space before `:`

Sources/UIKitBackend/UIKitBackend+CustomButton.swift
```diff
@@ -98,7 +139,7 @@ final class UICustomButton: UIButton {
     override public var isHighlighted: Bool {
         didSet {
             UIView.animate(
-                withDuration: 0.1,
+                withDuration: isHighlighted ? 0.1: 0.3,
```
stackotter: Put a space before the colon

Sources/SwiftCrossUI/Views/Button.swift
```diff
@@ -90,11 +90,20 @@ extension Button: TypeSafeView {
         environment: EnvironmentValues,
         backend: Backend
     ) -> ViewLayoutResult {
-        var childEnvironment = backend.buttonLabelEnvironment(from: environment)
+        let buttonPadding = backend.buttonPadding(in: environment)
+        var childEnvironment = backend.computeButtonLabelEnvironment(from: environment)
```
stackotter: Child environment should be a `let` (according to the CI warnings)

stackotter:
## GtkBackend

The first image shows CounterExample with this PR, and the second image shows CounterExample with the latest commit to main. It looks like both the horizontal and vertical padding of buttons are incorrect, and the label of the `-` button seems off centre (but I have no idea why that would be).

Screenshot From 2026-06-22 16-33-32

Screenshot From 2026-06-22 16-39-40

stackotter:
## Gtk3Backend

First image is new PR buttons, and second image is buttons from latest commit to main. Seems to have similar issues to GtkBackend, except the minus label appears to be well aligned. Other than that, the buttons from main appear to have no press effect, while the buttons in this PR have a rather obvious dark press effect.

Screenshot From 2026-06-22 16-52-27

Screenshot From 2026-06-22 16-54-22

stackotter:
## WinUIBackend

The padding seems a bit off on WinUI, but in general the interactions seem to match native perfectly, so this should be an easy fix. The first one is under this PR, and the second is with the latest commit of main.

Screenshot From 2026-06-22 16-48-13

Screenshot From 2026-06-22 17-01-17

stackotter (changes requested):
Regarding the code, I've only got small comments this time. I've also had quite a few compilation warnings while testing locally (both in core and in WinUIBackend).

Examples/Sources/ControlsExample/ControlsApp.swift
```diff
@@ -55,8 +57,37 @@ struct ControlsApp: App {
...
+                                    Picker(
+                                        of: [
+                                            ButtonStyle.bordered,
+                                            ButtonStyle.plain,
+                                            ButtonStyle.borderless
+                                        ],
+                                        selection: $buttonStyle
+                                    )
+                                #endif
+                                VStack(spacing: 0) {
```
stackotter: I think this VStack is redundant?

Examples/Sources/ControlsExample/ControlsApp.swift
```diff
@@ -55,8 +57,37 @@ struct ControlsApp: App {
...
+                                        count += 1
+                                    } label: {
+                                        Text("Click me!")
+                                            .padding(.horizontal, 11)
+                                            .padding(.vertical, 4)
+                                            .background {
+                                                Color.purple.opacity(0.5)
+                                                    .cornerRadius(5)
+                                            }
+                                            .foregroundColor(.adaptive(light: .black, dark: .white))
```
stackotter: Make this a text-labelled button instead. I believe that ControlsExample should be showcasing default behaviour instead of customisation behaviour.

stackotter (changes requested):

MiaKoring:
> ## GtkBackend
> 
> 
> 
> The first image shows CounterExample with this PR, and the second image shows CounterExample with the latest commit to main. It looks like both the horizontal and vertical padding of buttons are incorrect, and the label of the `-` button seems off centre (but I have no idea why that would be).
> 
> 
> 
> Screenshot From 2026-06-22 16-33-32
> 
> 
> 
> Screenshot From 2026-06-22 16-39-40
> 
> 

My suspicion is that thats due to me disabling the minsize

MiaKoring:
winui could be these symbols measuring smaller, but I'd have to check

Examples/Sources/CounterExample/CounterApp.swift
```diff
@@ -17,6 +17,7 @@ struct CounterApp: App {
                     Button("-") {
                         count -= 1
                     }
+
```
stackotter: Extra line

Sources/Gtk3Backend/Gtk3Backend+Button.swift
```diff
@@ -62,8 +59,9 @@ extension Gtk3Backend {
 }
 
 fileprivate final class GtkCustomButton: Gtk3.Button {
-    static let horizontalPadding: Double = 12
-    static let verticalPadding: Double = 6
+    // This value is the result of measurements.
+    // Runtime computing it is not viable.
+    static let buttonPadding = SIMD2<Int>(33, 17)
```
stackotter: Are these meant to be odd numbers? How would centering work properly otherwise?

stackotter (changes requested):
Thanks for making those changes. Just have two comments for this latest set of changes. Then after conflicts are resolved, it should be ready for some final testing and finally a merge into main.

MiaKoring:
Sorry about the Gtk3 padding thing. I don’t remember how I measured it, but it was 1 too small on both axes.

Conflicts are resolved now.

Thank you for sticking through this. I can imagine this being an insanely annoying PR to review, as it was very annoying to develop too. I hope my next few PRs will have less friction.

MiaKoring:
ok, the ci fail is an easy fix, looks like I now need to @_spi import SwiftCrossUI

MiaKoring:
Okay, I guess I’ll have to look into android tomorrow. It says somethign about inspection

MiaKoring:
Latest commit fixes the fail on android. It was caused by a missing @_spi(Backends) import SwiftCrossUI in Android Backend’s inspection modifiers file. Also I removed some dead variables I noticed through warnings from the android gradients implementation

Sources/SwiftCrossUI/Views/Button.swift
```diff
@@ -1,10 +1,10 @@
 /// A control that initiates an action.
 public struct Button<Label: View> {
     public typealias Content = TupleView1<Label>
-    /// The label to show on the button.
-    public var label: () -> Label
+    @_spi(Backends) public var label: () -> Label
     /// The action to be performed when the button is clicked.
-    package var action: @MainActor @Sendable () -> Void
+    @_spi(Backends) public var action: @MainActor @Sendable () -> Void
+
```
stackotter, line 8: Extra blank line

Sources/SwiftCrossUI/Views/Button.swift
```diff
@@ -1,10 +1,10 @@
 /// A control that initiates an action.
 public struct Button<Label: View> {
     public typealias Content = TupleView1<Label>
-    /// The label to show on the button.
```
stackotter, line 3: Accidentally removed this doc comment

stackotter (changes requested):

stackotter (changes requested):
I've gone through all of the new changes; just had two formatting/docs things to mention, but other than that it's looking good

stackotter (approved):
Thanks for fixing those up!

stackotter:
While doing one last local test on each platform, I found that the labels are still significantly off-centre with GtkBackend running on my Fedora PC. Gtk3Backend on the other hand, looks fine.

Here is CounterExample running with this PR;

off center labels

And here is CounterExample running with latest main;

current button label centering

Looks like some left padding is missing (because the padding on the other 3 sides looks fine).

MiaKoring:
How do we want to proceed here?
In a newer Gtk version it's fixed and it currently only happens on Buttons with very small labels. Could a fix for that on older Gtk be a future thing?

fbartho:
I think I found an existing gap that has a fix worth folding in: `DummyBackend.Button` never overrides `naturalSize`, so it inherits `.zero` and every button lays out 0×0 — core delegates button sizing entirely to `backend.naturalSize(of:)` (see the TODO in `Button.computeLayout`). It looks like `SimpleButton` and the new wrapping `Button` in this PR carry the same gap forward. A working override for the current shape, as reference: https://github.com/fbartho/swift-cross-ui/commit/aee425a

**LLM Usage:** *I used Claude to find this bug and verify its root cause. I authored this comment and take responsibility for its contents.*

fbartho:
As I putz along working on my attempt at a StaticHTMLBackend for myself, I tripped across exactly the missing APIs that this PR is trying to land!

Is there anything I can do to help get this merged?

Thanks!

stackotter:
> Could a fix for that on older Gtk be a future thing?

Yeah ok, happy for that to be the case as long as it's recorded as a GitHub issue

PR #591, opened 2026-05-25
Fix: `Gtk3Backend` to match `GtkBackend` signal registering behavior.

furbytm:
> [!IMPORTANT]
> `@State` and dynamic properties broke on `Gtk3` after #588 because the
> generated `registerSignals()` overrides were never being called — `Gtk3Backend`'s
> `didMoveToParent` didn't call `removeSignals()`/`registerSignals()` unlike `GtkBackend`.

> [!NOTE]
> This PR aligns `Gtk3Backend` behavior with `GtkBackend` to remove subtle differences
> and prevent unknown behavior.

### Fixes:
- Call `removeSignals()` and `registerSignals()` in `Gtk3`'s `didMoveToParent`, matching
  `GtkBackend`'s existing pattern and restoring dynamic property support on the Gtk3 backend.
- Disconnect all GTK signal handlers before releasing the `GObject` in `deinit` to prevent
  callbacks firing into deallocated Swift wrapper instances — now consistent across both backends.
- Moved `Gtk3` base `Widget` signal registrations out of `didMoveToParent()` and into a new dedicated `registerSignals()` override method.

### Potentially related issues:
- https://github.com/moreSwift/swift-cross-ui/issues/434
- https://github.com/moreSwift/swift-cross-ui/issues/433

### DIFF
--- a/Sources/Gtk3/Widgets/Widget.swift
+++ b/Sources/Gtk3/Widgets/Widget.swift
@@ -25,7 +25,12 @@ open class Widget: GObject {
 
     open func didMoveToParent() {
         removeSignals()
-      
+        registerSignals()
+    }
+
+    open func didMoveFromParent() {}
+
+    open override func registerSignals() {
         // The Gtk3 docs claim that this handler should take GdkEventButton as a
         // value, but that leads to crashes on Rocky Linux. These crashes are
         // fixed by instead taking the event as a pointer. I've confirmed that
@@ -102,12 +107,8 @@ open class Widget: GObject {
             guard let self else { return }
             self.styleUpdated?()
         }
-      
-        registerSignals()
     }
-
-    open func didMoveFromParent() {}
-
+  
     public func queueDraw() {
         gtk_widget_queue_draw(widgetPointer)
     }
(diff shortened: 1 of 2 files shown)


furbytm:
Thanks to @kaascevich for flagging the **Gtk3** backend regression and helping work through the fix.

Sources/Gtk3/Widgets/Widget.swift
```diff
@@ -100,6 +102,8 @@ open class Widget: GObject {
             guard let self else { return }
             self.styleUpdated?()
         }
```
stackotter, line 109: While we're at it, I think it'd be a good idea to move these signal handlers to Widget's (currently non-existent) registerSignals method

stackotter (changes requested):

Sources/Gtk3/Widgets/Widget.swift
furbytm: Good idea! I added a `registerSignals()` override for this base `Widget` class, and moved all the `addSignal()` calls into there.

stackotter (approved):
Thanks for addressing that comment; I'll test locally rq then merge if all goes well

PR #593, opened 2026-05-26
Android Representables

bbrk24:
Implements `AndroidViewRepresentable` and `AndroidxFragmentRepresentable`. As an aside, we currently depend on a very old version of androidx.fragment, and could afford to upgrade.

### DIFF
--- a/Sources/AndroidBackend/AndroidViewRepresentable.swift
+++ b/Sources/AndroidBackend/AndroidViewRepresentable.swift
@@ -1,0 +1,197 @@
+import AndroidKit
+import SwiftCrossUI
+import SwiftJava
+
+public struct AndroidViewRepresentableContext<Representable: AndroidViewRepresentable> {
+    public let coordinator: Representable.Coordinator
+    public internal(set) var environment: EnvironmentValues
+}
+
+public protocol AndroidViewRepresentable: SwiftCrossUI.View where Content == Never {
+    associatedtype ViewType: AndroidKit.View
+    associatedtype Coordinator = Void
+
+    @MainActor
+    func makeCoordinator() -> Coordinator
+
+    @MainActor
+    func makeAndroidView(context: Self.Context) -> ViewType
+
+    @MainActor
+    func updateAndroidView(
+        _ view: ViewType,
+        context: Self.Context
+    )
+
+    @MainActor
+    func sizeThatFits(
+        _ proposal: ProposedViewSize,
+        view: ViewType,
+        context: Self.Context
+    ) -> ViewSize
+}
+
+extension AndroidViewRepresentable {
+    public typealias Context = AndroidViewRepresentableContext<Self>
+
+    @MainActor
+    public func sizeThatFits(
+        _ proposal: ProposedViewSize,
+        view: ViewType,
+        context: Self.Context
+    ) -> ViewSize {
+        let density = context.environment.windowScaleFactor
+
+        // 0x80000000 = View.MeasureSpec.AT_MOST
+        // 0x3FFFFFFF = View.MeasureSpec.makeMeasureSpec(Int32.max, View.MeasureSpec.UNSPECIFIED)
+        let widthMeasureSpec =
+            if
+                let proposedWidth = proposal.width,
+                proposedWidth > 0
+            {
+                Int32(bitPattern: 0x80000000 | UInt32(min(proposedWidth * density, 0x3FFFFFFF)))
+            } else {
+                0x3FFFFFFF as Int32
+            }
+
+        let heightMeasureSpec =
+            if
+                let proposedHeight = proposal.height,
+                proposedHeight > 0
+            {
+                Int32(bitPattern: 0x80000000 | UInt32(min(proposedHeight * density, 0x3FFFFFFF)))
+            } else {
+                0x3FFFFFFF as Int32
+            }
+
+        view.measure(widthMeasureSpec, heightMeasureSpec)
+
+        let width = Double(view.getMeasuredWidth()) / density
+        let height = Double(view.getMeasuredHeight()) / density
+
+        return ViewSize(width, height)
+    }
+}
+
+extension AndroidViewRepresentable where Coordinator == Void {
+    public func makeCoordinator() {}
+}
+
+@JavaClass(
+    "dev.swiftcrossui.androidbackend.RepresentingView",
+    extends: AndroidKit.FrameLayout.self
+)
+class RepresentingView: AndroidKit.FrameLayout {
+    @JavaMethod
+    convenience init(
+        _ activity: AndroidKit.Activity?,
+        environment: JNIEnvironment? = nil
+    )
+
+    @JavaMethod
+    func getSwiftContext() -> SwiftObject?
+
+    @JavaMethod
+    func setSwiftContext(_ swiftContext: SwiftObject?)
+
+    @JavaMethod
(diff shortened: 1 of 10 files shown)


stackotter:
The linter doesn't like the force try in one of your representable implementations. Probably just missing the comment to disable that rule for that file

Sources/AndroidBackend/Kotlin/RepresentingView.kt
```diff
@@ -0,0 +1,29 @@
...
+    var swiftContext: SwiftObject? = null
+
+    private var _child: View? = null
+
+    var child: View?
+        get() = _child
+        set(value) {
+            _child = value
+            value?.let {
+                addView(it)
```
stackotter, line 20: I think we should call `removeAllViews()` before adding the new child in order to allow this setter to be invoked more than once

Sources/AndroidBackend/Kotlin/RepresentingView.kt
```diff
@@ -0,0 +1,29 @@
...
+import android.app.Activity
+import android.view.Gravity
+import android.view.View
+import android.view.ViewGroup
+import android.widget.FrameLayout
+
+class RepresentingView(activity: Activity) : FrameLayout(activity) {
+    var swiftContext: SwiftObject? = null
+
+    private var _child: View? = null
```
stackotter, line 12: What's the reason for underscoring a private-marked field?

Sources/AndroidBackend/AndroidViewRepresentable.swift
```diff
@@ -0,0 +1,219 @@
...
+
+    /// Create the initial `android.view.View` instance.
+    @MainActor
+    func makeAndroidView(context: Self.Context) -> ViewType
+
+    /// Update the view with new values.
+    /// - Parameters:
+    ///   - view: The view to update.
+    ///   - context: The context, including the coordinator and potentially new environment
+    ///   values.
```
stackotter: Indent this hanging line one more level (to match the start of the bullet item content)

Sources/AndroidBackend/AndroidViewRepresentable.swift
```diff
@@ -0,0 +1,219 @@
+import AndroidKit
+import SwiftCrossUI
+import SwiftJava
+
+public struct AndroidViewRepresentableContext<Representable: AndroidViewRepresentable> {
```
stackotter, line 6: Is it possible to make this `@MainActor` to match the other backends' representable context types?

Sources/AndroidBackend/AndroidViewRepresentable.swift
```diff
@@ -0,0 +1,219 @@
...
+    /// - Returns: The view's preferred size.
+    @MainActor
+    func sizeThatFits(
+        _ proposal: ProposedViewSize,
+        view: ViewType,
+        context: Self.Context
+    ) -> ViewSize
+
+    // TODO(bbrk24): Support dismantleAndroidView
+    // View doesn't have the same kind of lifecycle as Fragment, so it's not straightforward
```
stackotter, line 54: Can we support this on the Swift side of things? If we call the dismantle method in the Swift widget's deinit it should get called when the corresponding view graph node gets dropped.

Sources/AndroidBackend/AndroidxFragmentRepresentable.swift
```diff
@@ -0,0 +1,271 @@
+import AndroidKit
+import SwiftCrossUI
+import SwiftJava
+
+public struct AndroidxFragmentRepresentableContext<Representable: AndroidxFragmentRepresentable> {
```
stackotter, line 6: Can this be `@MainActor`?

Sources/AndroidBackend/AndroidxFragmentRepresentable.swift
```diff
@@ -0,0 +1,271 @@
...
+
+    /// Create the initial `androidx.fragment.app.Fragment` instance.
+    @MainActor
+    func makeFragment(context: Self.Context) -> FragmentType
+
+    /// Update the fragment with new values.
+    /// - Parameters:
+    ///   - fragment: The fragment to update.
+    ///   - context: The context, including the coordinator and potentially new environment
+    ///   values.
```
stackotter: Indent this hanging line one level

stackotter (changes requested):

Sources/AndroidBackend/Kotlin/RepresentingView.kt
bbrk24: Just so that the name doesn't conflict with the public (computed) `var child` below. Android itself seems to prefer the prefix `m` for this, so this would be `mChild`, but the projects I've worked on use `_` instead.

stackotter:
The implementations look great. Would you be able to update the AdvancedCustomizationExample app to include an example of these representables in action? You can `#if` out the `inspect` modifiers part of the example, because that's of course not relevant for Android yet

Sources/AndroidBackend/Kotlin/RepresentingView.kt
stackotter: Ohhh, yep, figured it might be a name clash thing but didn't see that the clash was literally right below it hahah, makes sense.

Sources/AndroidBackend/Kotlin/RepresentingView.kt
bbrk24: The Swift code only calls `setChild` once, immediately after calling `makeAndroidView`. Unless the widget is reused in a way I'm not aware of, I don't think this should cause any problems.

Sources/AndroidBackend/AndroidViewRepresentable.swift
bbrk24: I don't see why not, but maybe the compiler will have a different idea

Sources/AndroidBackend/Kotlin/RepresentingView.kt
stackotter: Yeah, I was was more just suggesting this so that the method behaves like a proper setter such that if we do change how things work in future, it's not too easy to use the method wrong. This makes it easier to locally reason about what the code is actually trying to do (currently looks like it's trying to add a child rather than setting the solo child)

Sources/AndroidBackend/AndroidViewRepresentable.swift
stackotter: I'm happy for this to be left as a future direction and just mentioned in a new issue instead

stackotter (changes requested):

Sources/AndroidBackend/AndroidViewRepresentable.swift
bbrk24: Yeah, I didn't want to trust the Swift `deinit` for wrappers around Java objects. But I did try putting it in the Java `finalize`, which is basically the same thing, but I ran into the issue that that means somehow materializing the representable type and casting the view and coordinator for that just to be able to call the dismantle function, which I had issues with.

stackotter (dismissed):
Looks good now! I'll test locally and then merge

stackotter:
For some reason the 'CUSTOM NATIVE BUTTON' label wraps on my Pixel when I run AdvancedCustomizationExample. Does that happen for you too?

stackotter (approved):

PR #598, opened 2026-05-29
Implement ScrollView for Android

bbrk24:
`ScrollView([.horizontal, .vertical])` is still a little janky -- you can't scroll diagonally and sometimes only one scroll bar is visible, but it is technically supported.


stackotter (approved):
Looks good! Will test locally then merge if all seems well

stackotter:
Works great for the single axis scrolling that I tested locally, thanks for tackling this!

PR #599, opened 2026-05-31
Android file open dialog

bbrk24:
This PR adds support for file open dialogs in AndroidBackend. Save dialogs are left for future work, as they require the MIME type to be given.

Because of how Android does sandboxing, the URLs produced are not `file://` URLs, nor do they directly represent the folder structure on disk. They need to be passed into `ContentResolver` rather than being read directly.

### DIFF
--- a/Sources/AndroidBackend/AndroidBackend.swift
+++ b/Sources/AndroidBackend/AndroidBackend.swift
@@ -208,8 +208,8 @@ public final class AndroidBackend: BackendFeatures.BaseStubs {
 
         let matchParent = try! JavaClass<AndroidKit.ViewGroup.LayoutParams>().MATCH_PARENT
 
-        let leftInset = Int(Self.helpers.getSafeAreaLeftInset(Self.activity))
-        let topInset = Int(Self.helpers.getSafeAreaTopInset(Self.activity))
+        let leftInset = Int(helpers.getSafeAreaLeftInset(Self.activity))
+        let topInset = Int(helpers.getSafeAreaTopInset(Self.activity))
         let fullWindowSize = SIMD2(Int(matchParent), Int(matchParent))
         setSize(of: container, to: fullWindowSize)
         setPosition(ofChildAt: 0, in: container, to: SIMD2(leftInset, topInset))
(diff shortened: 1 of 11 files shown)


Sources/AndroidBackend/AndroidBackend.swift
```diff
@@ -112,6 +115,26 @@ public final class AndroidBackend: BackendFeatures.BaseStubs {
 
     public init() {
         helpers = AndroidBackendHelpers(environment: Self.env)
+
+        let fragmentActivity = Self.activity.as(FragmentActivity.self)!
+
+        let filesCallback = FilesActivityCallback(environment: Self.env)
+        let filesAction = SwiftAction(environment: Self.env) {
+            let urls = filesCallback.getUrlStrings()
+            AndroidBackend.fileDialogCallback?(urls.map { Foundation.URL(string: $0)! })
```
stackotter: Might be worth adding a custom fatal error message when the URL fails to parse, just to make such an issue a bit more debuggable. Same with the one below

stackotter (approved):
Looks good, just going to test locally before merging

stackotter (changes requested):
Oops, forgot that I left one requested change

stackotter:
For someone to use the file URL that the dialog returns, do they have to use Android's ContentResolver? Is there any way for us to resolve the content URI to a file URI on behalf of the app? I guess not if Android doesn't have a granular file sandbox

It might be time for us to give in and adopt SwiftUI's 'importer/exporter' based file open/save dialog system (where you generally don't get the URLs of the underlying files, receiving a handle instead)

bbrk24:
No, the only way to use a content URI is to pass it to ContentResolver, which opens the file and returns a file descriptor.

stackotter:
Yeah ok, looks like we'll have to update to an importer/exporter pattern rather than our current methods that just fetch URLs to the requested content

stackotter:
I've created an issue for the future enhancement that I mentioned; https://github.com/moreSwift/swift-cross-ui/issues/605

I'm happy to merge this PR once the comment has been resolved.

stackotter (dismissed):
Thanks for fixing that up (and the WebView one too)

stackotter:
Whoops, this has merge conflicts now

stackotter (approved):

PR #602, opened 2026-05-31
[AndroidBackend] Fix windowing behavior on older API levels

### LINKED ISSUE #579: AndroidBackend crashes immediately pre-API 34

bbrk24:
Fixes #579

This PR changes how the window is measured in older API levels. I also found that the system accounts for the insets for you on API 30-34, so I apply the legacy logic for those API levels, too (not just API 28-29, where the functions we've been using don't exist).

### DIFF
--- a/Examples/Package.resolved
+++ b/Examples/Package.resolved
@@ -208,15 +208,6 @@
       }
     },
     {
-      "identity" : "swift-cwinrt",
-      "kind" : "remoteSourceControl",
-      "location" : "https://github.com/moreSwift/swift-cwinrt",
-      "state" : {
-        "revision" : "601287a2a89e8e56e8aa53782b6054db8be0bce8",
-        "version" : "0.1.2"
-      }
-    },
-    {
       "identity" : "swift-filestreamer",
       "kind" : "remoteSourceControl",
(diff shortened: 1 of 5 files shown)


stackotter (approved):
Thanks for resolving that 🙏

PR #606, opened 2026-06-01
Implement images in AndroidBackend

bbrk24:
It's a bit difficult to test because swift-bundler doesn't handle resources on Android correctly, but you can e.g. load images from a network and it seems to work fine.


stackotter (approved):
Looks good, thanks for implementing that 🙏

PR #611, opened 2026-06-01
Add tooltip support on Android

bbrk24:


stackotter (approved):
Wow, that really was very simple hahah

PR #613, opened 2026-06-01
Fix AppStorage on Android

### LINKED ISSUE #609: [AndroidBackend] `@AppStorage` doesn't persist across app restart
It does persist across activity recreation, but not across the entire app being closed and re-opened.

bbrk24:
Fixes #609

UserDefaults don't appear to work on Android, so this PR adds `SharedPreferencesAppStorageProvider` and makes it the default Android app storage provider.

### DIFF
--- a/Sources/AndroidBackend/SharedPreferencesAppStorageProvider.swift
+++ b/Sources/AndroidBackend/SharedPreferencesAppStorageProvider.swift
@@ -9,6 +9,7 @@ extension AndroidKit.SharedPreferences {
     func getString(key: String, defaultValue: JavaString?) throws -> JavaString?
 }
 
+// swiftlint:disable force_try
 public struct SharedPreferencesAppStorageProvider: AppStorageProvider {
     private let sharedPreferences: AndroidKit.SharedPreferences
     private let encoder = Foundation.JSONEncoder()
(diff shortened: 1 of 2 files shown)


Sources/AndroidBackend/SharedPreferencesAppStorageProvider.swift
```diff
@@ -0,0 +1,58 @@
...
+        editor.apply()
+    }
+
+    public func retrieveValue<Value: Codable>(
+        ofType type: Value.Type,
+        forKey key: String
+    ) -> Value? {
+        var jsonString: String
+        do {
+            guard let javaString = try sharedPreferences.getString(key: key, defaultValue: nil)
```
stackotter: Put the let on the next line (with a hanging indent)

Sources/AndroidBackend/SharedPreferencesAppStorageProvider.swift
```diff
@@ -0,0 +1,58 @@
...
+            else {
+                return nil
+            }
+            jsonString = javaString.toString()
+        } catch {
+            log("Exception thrown by sharedPreferences.getString: \(error)")
+            return nil
+        }
+
+        guard let data = jsonString.data(using: .utf8) else { return nil }
```
stackotter: `let data = Data(jsonString.utf8)` gives you the data directly and avoids the need for `nil` handling

stackotter (changes requested):
Looks great, my feedback is only minor. Other than my feedback there's just a linter CI failure because of the force try

stackotter (approved):
Thanks for addressing those comments, looks great now

Issue #614, opened 2026-06-03
swift package manifest error on unix for an unexpected comma.


furbytm:
When building my own project, [Ubuntu CI is throwing errors about an unexpected comma](https://github.com/wabiverse/SwiftUSD/actions/runs/26860869559/job/79213913874#step:13:204) for the AndroidBackend in the `package.products.append()` call.

PR #615, opened 2026-06-03
Fix trailing comma in Package.swift

furbytm:
Related issue(s) this fix addresses:
- #614

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -400,7 +400,7 @@ if androidBackendSupported {
     ]
 
     package.products.append(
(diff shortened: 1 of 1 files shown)


stackotter (approved):
Thanks for fixing this

furbytm:
> Thanks for fixing this

ayy, no formatting issues in this revision! 😉

furbytm:
@stackotter perhaps we may want to consider tagging and issuing a new release for this reasonably soon, else many unix builds break without this fix.

PR #621, opened 2026-06-05
Fix building SwiftCrossUI from Xcode

### LINKED ISSUE #586: Can't Compile Using Xcode
When I try to compile any of the Examples in the main branch using Xcode 26.5 on macOS 26.3.1, I get the following error:

```
swift-cross-ui/Sources/AndroidBackendShim/include/header.h:1:10: 'android/log.h' file not found
in file included from swift-cross-ui/Sources/AndroidBackendShim/impl.c:1:
```

stackotter:
Xcode ignores the dependency conditions that we use to only include AndroidBackend when targeting Android (even though SwiftPM understands them just fine). I've previously implemented a workaround for the same issue but for xcodebuild. This PR extends the workaround to apply to Xcode as well.

This fixes #586.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -26,7 +26,7 @@ if let version = getGtk4MinorVersion(), version >= 10 {
     gtkSwiftSettings.append(.define("GTK_4_10_PLUS"))
 }
 
-let invokedByXcodebuild: Bool
+let invokedByXcode: Bool
 #if os(macOS)
     import Darwin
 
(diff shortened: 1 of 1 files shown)


unnamedd:
@stackotter yeah, with the change here, it's fixed now. Building from inside Xcode works.

Here's the change I made
```swift
    .package(url: "https://github.com/moreSwift/swift-cross-ui", branch: "stackotter/fix_586"),
    .package(url: "https://github.com/apple/swift-argument-parser", from: "1.3.0"),
```

PR #624, opened 2026-06-06
Add ActivityDelegate

bbrk24:
Discussed in Discord beforehand. Motivating use-case is to allow registering custom activity launchers in `onCreate`, but I included all the other methods for completeness.

### DIFF
--- a/Sources/AndroidBackend/ActivityDelegate.swift
+++ b/Sources/AndroidBackend/ActivityDelegate.swift
@@ -1,0 +1,145 @@
+import AndroidKit
+import SwiftJava
+
+/// A delegate that is notified of the activity's lifecycle changes.
+///
+/// Since it's (currently) not possible to use a custom subclass of `FragmentActivity` in
+/// SwiftCrossUI apps, this allows you to act on lifecycle events instead. All methods have default
+/// implementations that do nothing.
+///
+/// - Important: Due to the different mechanisms Jetpack provides for installing listeners, some of
+/// these methods are called within the corresponding activity methods, while others are called
+/// shortly after the corresponding activity methods.
+///
+/// Methods called during the corresponding activity method are:
+/// - ``onCreate(of:env:)``
+/// - ``onConfigurationChanged(for:to:env:)``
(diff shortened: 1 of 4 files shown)


Sources/AndroidBackend/ActivityDelegate.swift
```diff
@@ -0,0 +1,145 @@
...
+public protocol ActivityDelegate {
+    func onCreate(of activity: FragmentActivity, env: JNIEnvironment?)
+    func onStart(of activity: FragmentActivity, env: JNIEnvironment?)
+    func onResume(of activity: FragmentActivity, env: JNIEnvironment?)
+    func onPause(of activity: FragmentActivity, env: JNIEnvironment?)
+    func onStop(of activity: FragmentActivity, env: JNIEnvironment?)
+    func onDestroy(of activity: FragmentActivity, env: JNIEnvironment?)
+
+    func onConfigurationChanged(
+        for activity: FragmentActivity,
```
stackotter, line 34: I feel like this parameter label would make more sense as `of`, but it's probably fine either way? Just wanted to check you thought about that

stackotter (changes requested):
Generally looks great, just had one slight concern about a parameter label, but other than that no comments

Sources/AndroidBackend/ActivityDelegate.swift
bbrk24: In my mind, "of" with "changed" would be a description of what's changing. Like the English phrase "on configuration changed" could also be worded "on change of configuration" (though it seems a bit odd out of context). I used "for" here because it's the configuration for this activity; other activities could simultaneously have different configurations, e.g. on a multi-display setup.

stackotter: With your explanation I think that makes sense to me now. I was thinking 'on configuration changed (of blah activity) do this', but 'on configuration changed (for blah activity) do this' makes a similar amount of sense. For some reason the of version reads better in my head. But the symmetry with the 'new intent' delegate method also makes sense.

stackotter (approved):

PR #627, opened 2026-06-07
[AndroidBackend] TextField improvements & SecureField/TextEditor support

### LINKED ISSUE #603: AndroidBackend text field unnecessary write error
`05-31 20:19:16.349  5801  6020 E Swift   : 2026-05-31T20:19:16-0400 warning SwiftCrossUI: [SwiftCrossUI] Unnecessary write to text Binding of TextField detected, please open an issue at https://github.com/moreSwift/swift-cross-ui/issues/new so we can fix it for AndroidBackend.`

bbrk24:
Fixes #603 

This PR fleshes out the TextField support for AndroidBackend and adds support for SecureField and TextEditor. I also cleaned up a couple other files in AndroidBackend that I missed in previous PRs.


Sources/AndroidBackend/Kotlin/CustomEditText.kt
```diff
@@ -5,36 +5,64 @@ import android.util.Log
...
-                actionId == EditorInfo.IME_ACTION_SEND ||
-                    actionId == EditorInfo.IME_ACTION_GO ||
-                    actionId == EditorInfo.IME_ACTION_SEARCH
-            ) {
-                Log.i("CustomEditText", "Submit")
+            val action =
+                if (actionId == EditorInfo.IME_NULL) v.imeOptions and EditorInfo.IME_MASK_ACTION
+                else actionId
+
+            Log.v("CustomEditText", "Editor action!")
```
stackotter: I think we should remove the log messages now that things work

Sources/AndroidBackend/Kotlin/SecureEditText.kt
```diff
@@ -0,0 +1,46 @@
...
+        } else {
+            super.setInputType(InputType.TYPE_CLASS_TEXT or InputType.TYPE_TEXT_VARIATION_PASSWORD)
+        }
+
+        if (keyListener !is CustomKeyListener) {
+            keyListener = CustomKeyListener(keyListener)
+        }
+    }
+
+    private inner class CustomKeyListener(private val base: KeyListener?) : KeyListener {
```
stackotter, line 40: What's the purpose of this CustomKeyListener?

stackotter (changes requested):
The implementation generally looks good, and TextField & SecureField appear to be working correctly. TextEditor has some issues though.

To test TextEditor I commented out the NavigationSplitView part of NotesExample and updated `selectedNote` to be a computed property that returns `notes[0].id`.

Image

The TextEditor appears to take up the full height that SwiftCrossUI assigns it, but the actual text area remains at the height of a single line. It then lets you scroll the multiple lines of text within that one-line-tall clipped view of the mutliline text content.

The other issue is that the TextEditor has decorations (such as a background and border etc). SwiftCrossUI's TextEditor is designed to be a reusable text editing component without any padding, backgrounds, or borders. This allows people to use it in a wider range of scenarios (and matches SwiftUI's TextEditor).

In future I think we should introduce a way to request the multiline text editor to have native decorations (for use in forms etc), but for now we should remain consistent.

Sources/AndroidBackend/Kotlin/SecureEditText.kt
bbrk24: Maybe I should add a code comment. It lets you use keyboard types that aren't typically associated with passwords, like the email or URL keyboard, while still obscuring the text. Normally "password" is another type alongside "email", "url" etc, but when you use a KeyListener you can separate the keyboard from the obscured state of the text field.

Sources/AndroidBackend/Kotlin/CustomEditText.kt
bbrk24: Removed

Sources/AndroidBackend/Kotlin/SecureEditText.kt
stackotter: Ah ok, yeah a code comment would be great there, because I think that's a pretty hard rationale to discover without either knowing it already or reading all of the docs (without knowing what the original issue was)

stackotter (changes requested):
Thanks for fixing those TextEditor issues, looks great now. I'm happy to merge once that KeyListener comment is added

stackotter (approved):

PR #631, opened 2026-06-10
Android inspect/inspectWindow modifiers

bbrk24:


stackotter (dismissed):
Looks great! I'll test locally, then merge if all works

stackotter:
At first I thought there might be an issue with the inspect modifier's underlying view, because the button on the right is bigger than the one on the left, but it turns out that SwiftCrossUI leaves a margin around Android buttons for some reason 🤔 It isn't related to this PR but just thought I'd mention it. I've opened an issue for it here; https://github.com/moreSwift/swift-cross-ui/issues/632

Screenshot (Jun 10, 2026 10_43_06 PM)

Here's the same example but with the body of the inspect modifier commented out for the button on the right, and a `.background(Color.blue)` modifier applied to the button on the left.

Screenshot (Jun 10, 2026 10_48_04 PM)

stackotter:
Happy to merge once conflicts have been resolved

stackotter (approved):

PR #633, opened 2026-06-11
Android List support

bbrk24:


Sources/AndroidBackend/Kotlin/lists/CustomListAdapter.kt
```diff
@@ -0,0 +1,54 @@
...
+                lp.width = parent.width
+                lp
+            } else {
+                ViewGroup.LayoutParams(parent.width, height)
+            }
+
+        return view
+    }
+
+    override fun getDropDownView(position: Int, convertView: View?, parent: ViewGroup) =
```
stackotter: Why do we have to override `getDropDownView`?

stackotter (changes requested):
Thanks for implementing this, the implementation looks good. I just had one question about why `getDropDownView` is necessary

stackotter:
Just tested locally, all works well

Sources/AndroidBackend/Kotlin/lists/CustomListAdapter.kt
bbrk24: My bad, we don't. The base class already has this exact implementation: https://cs.android.com/android/platform/superproject/+/android-latest-release:frameworks/base/core/java/android/widget/BaseAdapter.java;drc=ad9254cb1d45844e0a3b7470292b77d34d31e962;l=74

I saw that it was an abstract method on a supertype of `BaseAdapter`, but missed that `BaseAdapter` itself already implements it.

stackotter (approved):
Thanks for that

PR #638, opened 2026-06-13
Re-add corner radius container and implement on Android

### LINKED ISSUE #634: Update corner radius for Android support
Although Android can render rounded rectangles of arbitrary colors natively, cutting off arbitrary views (as would be necessary when applying a corner radius to a ScrollView or Text) requires a custom ViewGroup. We should change the corner radius backend methods to use an optional wrapping widget (like tooltips do currently) and implement such a custom ViewGroup.

bbrk24:
Fixes #634. I also refactored the corner radius modifier slightly to be more similar to the help modifier.


stackotter (approved):
Looks good!

PR #641, opened 2026-06-13
[AndroidBackend] Fix TextField onChange being called on initial render

### LINKED ISSUE #640: [AndroidBackend] TextField etc clear their contents on initial render
This isn't just a rendering bug where it looks empty at first -- it actually calls the binding's `set` with empty string. This is particularly annoying when it's a binding to an `@AppStorage` field.

bbrk24:
Fixes #640


stackotter (approved):
Looks like a sensible enough fix 👍

PR #647, opened 2026-06-22
fix: disfavor ToolbarItem.frame(width:) so View.frame wins on iOS

### LINKED ISSUE #626: `.frame()` modifiers break when no height is specified on `Spacer` & `Button` on iOS.
On iOS only, things fail to compile when you add a `.frame()` modifier to a `Spacer()` or `Button()` without specifying a height:

Image

The error reads:
```bash
Error: type 'any ToolbarItem' cannot conform to 'View'
```

> [!IMPORTANT]
> There is [an overload](https://github.com/moreSwift/swift-cross-ui/blob/809aa89da24c460b13cb2e1daba5431d8398e38b/Sources/UIKitBackend/KeyboardToolbar.swift#L186-L207) that is marked `@available(visionOS, unavailable)` and same for tvOS.

mvanhorn:
## Summary

Adds `@_disfavoredOverload` to the `ToolbarItem.frame(width:)` overload in `Sources/UIKitBackend/KeyboardToolbar.swift` so that `.frame(width:)` on a `Spacer` or `Button` resolves to the backend-independent layout modifier instead of the keyboard-toolbar overload.

## Why this matters

On iOS, `Spacer` and `Button` conform to both `View` and `ToolbarItem`. The `ToolbarItem.frame(width:)` overload at `Sources/UIKitBackend/KeyboardToolbar.swift:193` has no defaulted arguments, so Swift's overload resolver preferred it over `View.frame(width:height:alignment:)` (`Sources/SwiftCrossUI/Views/Modifiers/Layout/FrameModifier.swift:25`) for plain `.frame(width:)` calls. Because that overload returns `any ToolbarItem`, the call failed to compile with `type 'any ToolbarItem' cannot conform to 'View'` (#626).

`@_disfavoredOverload` is the established convention for this in the repo: it is already applied to overloads in `FrameModifier.swift`. With the attribute, view receivers resolve to the layout modifier (fixing the bug), while inside a `@ToolbarBuilder` closure the result type is a `ToolbarItem`, so the toolbar overload is still selected by return-type context. The documented `Spacer().frame(width: _)` fixed-width-spacer usage in keyboard toolbars keeps working unchanged.

This is preferred over adding a new top-level `frame(width:)` overload, which the maintainer noted produced an ambiguous-overload error.

## Testing

- `swift build --target SwiftCrossUI` succeeds.
- `swiftformat --lint` passes on the changed file.
- The change is isolated to the `UIKitBackend` (iOS / macCatalyst) target; desktop and other backends are unaffected. The full iOS build is exercised by CI.

Fixes #626

### DIFF
--- a/Sources/UIKitBackend/KeyboardToolbar.swift
+++ b/Sources/UIKitBackend/KeyboardToolbar.swift
@@ -190,6 +190,7 @@ extension ToolbarItem {
     ///
     /// If `width` is positive, the item will have that exact width. If `width` is zero or
     /// nil, the item will have its natural size.
+    @_disfavoredOverload
     public func frame(width: Double?) -> any ToolbarItem {
         if #available(iOS 14, macCatalyst 14, *),
            self is Spacer || self is FixedWidthSpacerItem


stackotter (approved):
Ah interesting, this works because our builder has a `buildExpression` implementation. If it didn't, then I believe this would've broken compilation of keyboard toolbar builder closures that use the offending frame modifier. This combination of disfavoredOverload and buildExpression is good to keep in mind in future for resolving result builder inference issues.

mvanhorn:
Appreciate the merge, @stackotter. Disfavoring the ToolbarItem.frame(width:) overload so View.frame wins on iOS is a neat use of @_disfavoredOverload.

PR #650, opened 2026-06-23
AndroidBackend: Add Sliders and move off BaseStubs conformance

bbrk24:


bbrk24:
Seems like the CI failure was a transient network error:
> Warning: IO exception while downloading manifest:
java.net.ConnectException: Operation timed out

Sources/AndroidBackend/Kotlin/CustomSlider.kt
```diff
@@ -0,0 +1,47 @@
...
+
+    fun setBounds(min: Double, max: Double, places: Int) {
+        // Slider needs the values to be multiples of the step size, or else it crashes. However,
+        // in my testing, if you set the step size too small, it crashes anyways when it internally
+        // sets its own value due to user interaction. It seems like 0.1 is fine and 0.01 isn't, and
+        // it only displays the value to two decimal places anyways.
+        if (places > 1) {
+            stepSize = 0.0f
+        } else {
+            stepSize = BigDecimal.valueOf(1L, places).toFloat()
```
stackotter: Decimal places is only meant to affect the number of decimal places displayed by the slider (for backends that display the value). SwiftCrossUI always hardcodes it to either `0` (for 'integer' sliders) or `2` (for continuous sliders). But that's only meant to affect text, and step size is something separate (that we should really handle soon, but don't yet).

When I was double checking that, I found that UIKitBackend's slider implementation also uses decimalPlaces to round slider values. It shouldn't be doing that either. Feel free to change that in this PR if you want, or I can create an issue for it otherwise.

Sources/AndroidBackend/AndroidBackend+ScrollViews.swift
```diff
@@ -1,7 +1,7 @@
 import AndroidKit
 import SwiftCrossUI
 
-extension AndroidBackend: BackendFeatures.ScrollContainers {
+extension AndroidBackend {
```
stackotter, line 5: I reckon it could be nice to list the implicit conformances in a comment above the extension so that it's easier to discover what the code relates to (especially in `AndroidBackend+Progress.swift` where we implement multiple conformances).

Maybe just `/// Implements BackendFeatures.ScrollContainers` or something along those lines?

stackotter (changes requested):

Sources/AndroidBackend/Kotlin/CustomSlider.kt
bbrk24: By default, Android's slider shows two decimal places if the step size isn't an integer, no matter what its value actually is. There is a way to control the value that's displayed though (the documentation claims to have an example that uses metric prefixes, but the formatting is broken so the code doesn't show up right), so I'll investigate that.

stackotter: The decimalPlaces value isn't user controllable, so for now you can just set the stepSize to 1 when `decimalPlaces == 0`, and 0 when `decimalPlaces == 2`. I believe that should achieve the desired effect (at least from what you've said). Not great, but the slider system needs revamping anyway (to properly set the step size when choosing integers).

bbrk24: Switching the places over from being step size to label rounding ended up making the code a lot shorter.

Sources/AndroidBackend/AndroidBackend+ScrollViews.swift
bbrk24: Added all of these.

stackotter (approved):
The code looks good now, and the slider in ControlsExample works well. Thanks for implementing this 🙏

PR #651, opened 2026-06-23
Added @FocusState and related modifiers for AppKitBackend

MiaKoring:
This PR adds the first split of #414 

This PR introduces focus management, including state-driven focus, focusability control, and control over visual highlight displaying on AppKitBackend.

Feature Support:
All new features are supported on AppKitBackned

This PR adds support for 3 focus related features:
### `@FocusState`
FocusState is a bidirectional property wrapper. If the focus changes through user action it will be reflected in the state and the other way around setting it will focus the corresponding view.

`View/focused` works hand in hand with `@FocusState`. It needs to be added to a View with a `FocusState<T>.Binding` and a unique value (`any Hashable`) identifying it for a view to set the `@FocusState` to the unique value on focus gain and be selected when set programmatically. 
When multiple views are assigned the same value or multiple leaf views recieve it through the environment, it works reactive, but only the widget last handled by the layout engine will recieve the focus visibly to the user.

There is an overload of `@FocusState` and `View/focused` accepting a non optional Bool. This would be the preferred choice in cases where you only have one view/subtree to focus without needing granular information.

### `View/focusable`
`View/focusable` modifies the focusability for a subtree. Setting it to `Focusability.disabled` excludes the subtree from recieving focus via keyboard navigation/tab press. Just like SwiftUI it doesn’t prevent focus through mouse click, but the next tab press is going to move focus outside of the subtree again.

### `View/focusEffectDisabled`
…  does exactly what it sounds like, disables the focus effect/hightlight/ring when applied to a widget with parameter `true`.

## SwiftCrossUI
- added `BackendFeatures.Focus`
   - setFocusEffectDisabled: backend for `View/focusEffectDisabled`, only called by `ViewGraphNode`
   - registerFocusObservers: registers a widget with `FocusState`, only called by `ViewGraphNode`
- added `BackendFeatures.FocusDisabling`
   - createFocusContainer: marker controlling focusability of its subtree on backends without hierarchical focus control, otherwise normal container. Created by `View/focusable`
   - updateFocusContainer: sets the focusability for the subtree
- added 2 properties to `EnvironmentValues`:
   - focusObservers: contains data of all `View/focused` modifiers applied on the tree above a view
   - focusEffectDisabled: set by `View/focusEffectDisabled`
- added 1 property and 1 function to `ViewLayoutResult`:
   - func with: same as on environment, returns a copy of the result with the given property changed
   - shouldSetFocusData: Views need to opt in to participate in focus shenanigans. this property is set to true by focusable leaf views and otherwise false. 
   - `_BuiltinPickerImplementation` doesn't delegate this to `ViewGraphNode` and therefore doesn't set it, but is still part of this feature
- modified `ViewGraphNode/commit`: `if currentLayout?.shouldSetFocusData == true`, registers observers of `@FocusData` on the widget and handles `View/focusEffectDisabled`
- modified Leaf views layout function to have `ViewLayoutResult/shouldSetFocusData` set to true:
   - Button
   - Checkbox
   - ToggleButton
   - ToggleSwitch
   - DatePicker
   - Menu
   - Slider
   - TextEditor
   - TextField
   - SecureField
   - WebView
- modified `_BuiltinPickerImplementation/commit` to do the same `ViewGraphNode/commit` would, but with a different widget than `ViewGraphNode/commit` recieves.
- Added `FocusChainManager` protocol
   - Simplifies support for`View/focusable(.disabled)` on backends without builtin hierarchical support for disabling views (e.g. `AppKitBackend`)
   - requires implementation of a few functions to gain information about a widget and the framework’s focus chain
   - provides functions to call in a place where the UI framework asks for the next/previous view to focus, doing the heavy lifting.
   - added `FocusabilityContainer` protocol: gets used by `FocusChainManager` to identify a marker view
   - added `FocusChainParticipant` protocol: needs to be implemented by `AppBackend/Widget`, provides information about the visibility of a widget and wether it can be a tab stop
   - hopefully sufficient documentation
- Added `@FocusState`: basically just a bit more fancy `@State` constrained to `(any Hashable)?`, not accepting an initial value and being able to reset to nil or false (if the `Bool` overload is used). The Binding is a specific binding for `FocusState` because it also needs the reset capability.
- Added `FocusData`:
   - A struct erasing the generic of `@FocusState`, containing the type, what the value passed to `View/focused` is, if the value matches the current state and set&reset closures, used by a backend’s `FocusStateManager`
- Added `Focusability`: an enum currently with cases `.unmodified` and `.disabled`, used by `View/focusable`
- Added the mentioned view modifiers

## AppKitBackend
- Added `FocusabilityContainer`:
   - implements `FocusabilityContainer`
- Added new required functions of `AppBackend`
- Added FocusStateManager
   - Observes `NSWindow/firstResponder` and updates `@FocusState` accordingly
   - special handling for `NSObservableTextField` and `NSObservableSecureTextField`, as they pass the focus to a child view.
   - focuses a widget when its `[FocusData]` contains a match in state to identifying value
- Moved `NSCustomWindow` to separated file
- Added logic and properties on `NSCustomWindow` to support focus chain manipulation i.e. `View/focusable`
   - overrides `NSWindow/initialFirstResponder` getter to respect `View/focusable`
   - overrides `NSWindow/selectKeyView(preceding:)` and `NSWindow/selectKeyView(following:)` to use `FocusChainManager` custom logic, respecting `View/focusable`
   - implements `FocusStateManager` protocol requirements

# Additional information
## Examples
- ControlFocusabilityTest was added to test `View/focusable`. I would prefer to keep it around to make adoption in future/other backends easier.

## How does the `FocusChainManager` logic work?
1. currently focused element is supplied to a directional function
2. request suggestion from UI framework
3. checks suggestion for a `FocusabilityContainer` with `FocusabilityContainer/focusability == .disabled` somewhere above it.
   - if no disabled container is found, returns
   - if a container is found asks UI framework for suggestion following the previous suggestion
   - exits when the suggestion is an already visited widget (full circle)

I hope the list and explanation helps :)

Tests/building on other platforms might not succeed, as in this PR only AppKit is supported, like we talked about with making a “stacked” PR.


Sources/SwiftCrossUI/Backend/BackendFeatures/Focus.swift
```diff
@@ -0,0 +1,43 @@
...
+        ///
+        /// Used by ``View/focusable(_:)``.
+        func setFocusEffectDisabled(on widget: Widget, disabled: Bool)
+    }
+
+    /// Backend methods for disabling focusability of a subtree.
+    ///
+    /// Used by ``View/focusable(_:)``.
+    public protocol FocusDisabling: Core {
+        /// Create a container controlling the focusability of the widgets below it.
```
stackotter: I assume 'below' here means widgets that are descendants of the container? If so we should make the wording more clear, because it could sound like it's blocking focusability of widgets 'behind' it. Same for the other doc comment

Sources/SwiftCrossUI/Layout/ViewLayoutResult.swift
```diff
@@ -8,6 +8,9 @@ public struct ViewLayoutResult {
     public var participateInStackLayoutsWhenEmpty: Bool
     /// The preference values produced by the view and its children.
     public var preferences: PreferenceValues
+    /// If a view can ever get focused out of the box, this needs to be set to true,
+    /// to be compatible with `@FocusState`.
```
stackotter: The first sentence of a doc comment should generally be phrased more like a description of the property rather than implementation instructions. Then after a blank line you can put the advanced instructions, like so; (I haven't actually looked at where shouldSetfocusData is used yet, so I've put a placeholder for the description instead)

```swift
/// Controls whether ... blah ... foo.
///
/// If a view can natively receive focus, then this must be set to true,
/// in order for `@FocusState` to work correctly with the view.
```

Sources/SwiftCrossUI/Values/Focusability.swift
```diff
@@ -0,0 +1,10 @@
+/// Enables finer-grained control of ``View/focusable`` than a Bool woul.
```
stackotter: The first sentence should describe what the thing is, like so;

```swift
/// The focusability of a ``View``.
```

But now that I've read through the type some more, I feel like we could rename the type to make its usage more intuitive. If this were truly representing 'focusability' then it'd have "is focusable" and "is not focusable" as its cases (basically a boolean). This feels more like it's representing a "focusability modification", so perhaps we could call it `FocusabilityModification`? Still not a huge fan of that name, so maybe we could keep it as `Focusability`, but I just thought I should mention the idea.

Sources/SwiftCrossUI/ViewGraph/ViewGraphNode.swift
```diff
@@ -253,6 +253,29 @@ public class ViewGraphNode<NodeView: View, Backend: BaseAppBackend>: Sendable {
...
+            if let backend2 = backend as? any BackendFeatures.Focus {
+                setFocusData(on: backend2)
+            } else if
+                !parentEnvironment.focusObservers.isEmpty ||
+                parentEnvironment.focusEffectDisabled
+            {
+                logger.warnOnce("\(Backend.self) doesn't support focus control/tracking.")
+            }
+
+            func setFocusData<Backend2: BackendFeatures.Focus>(on backend: Backend2) {
```
stackotter: imo the parameter should be labelled `with`, because it's not setting the focus data on the backend, but it is setting the focus data with/using the backend.

Also, I think it reads better if the function is defined before the call site. I was very confused when I read the call site without knowing that the function was a locally-scoped helper, and I was wondering why the function wasn't taking a widget parameter.

It might be worth renaming the function to `setWidgetFocusData` as well, because then it's clear what it's setting the focus data on when it doesn't take the target thing as a parameter (otherwise I would assume it's the focus data of the ViewGraphNode or the backend)

Sources/SwiftCrossUI/Views/Modifiers/FocusModifier.swift
```diff
@@ -0,0 +1,137 @@
...
+    public func focusable(_ focusability: Focusability = .unmodified) -> some View {
+        FocusModifier(body: TupleView1(self), focusability: focusability)
+    }
+
+    /// Conditionally disables the focus indicator.
+    public func focusEffectDisabled(_ disabled: Bool = true) -> some View {
+        EnvironmentModifier(self) { environment in
+            environment.with(\.focusEffectDisabled, disabled)
+        }
+    }
```
stackotter: This file should be renamed to `FocusedModifier` and these two modifiers should go in their own files named according to their modifier names (`FocusableModifier` and `FocusEffectDisabledModifier`).

Sources/SwiftCrossUI/Views/_BuiltinPickerImplementation.swift
```diff
@@ -79,6 +79,27 @@ public struct _BuiltinPickerImplementation: TypeSafeView {
...
+        if let backend2 = backend as? any BackendFeatures.Focus {
+            setFocusData(on: backend2)
+        } else if
+            !environment.focusObservers.isEmpty ||
+            environment.focusEffectDisabled
+        {
+            logger.warnOnce("\(Backend.self) doesn't support focus control/tracking.")
+        }
+
+        func setFocusData<Backend2: BackendFeatures.Focus>(on backend: Backend2) {
```
stackotter: This is basically identical to the ViewGraphNode helper of the same name, so both callsites should use the same implementation. I can imagine someone updating the one in ViewGraphNode without realising that there's a copy of it here as well.

Perhaps `BackendHelpers.updateFocusData<Backend: BackendFeatures.Focus>(of:with:environment:)` or something like that? You'd create a `BackendHelpers` enum in `Sources/SwiftCrossUI/Backend` to do so (and make the function a static method)

Sources/SwiftCrossUI/Views/Button.swift
```diff
@@ -120,7 +120,9 @@ extension Button: TypeSafeView {
             Int(childResult.size.height) + buttonPadding.y
         )
 
-        return ViewLayoutResult.leafView(size: ViewSize(size))
+        return ViewLayoutResult
+            .leafView(size: ViewSize(size))
+            .with(\.shouldSetFocusData, true)
```
stackotter: `shouldSetFocusData` doesn't read well or intuitively imo. I assume it's named that way because it's controlling whether ViewGraphNode calls its internal `setFocusData` method, but I think it'd be better to name it after the actual meaning of the property.

Naming this property is definitely tricky, but here are some ideas I have so far (some may be incorrect because I haven't finished reading through all of your code yet). If my understanding is correct, this property is representing whether a view has the possibility of being focusable? So false if the view's widget will never be directly focusable, and true otherwise?

- `neverFocusable` (this would store the opposite value, and true would signify that we can skip focus related operations as an optimisation. the default value would be true, and special views would have to override it. I like this one because 'never' sounds like a strong assertion, and makes it sound like false would be the safe default value, which is correct?)
- `canBeFocusable` (this would store the same value as the original; true would mean that we can't skip focus related operations)
- Joke option: `focusableable`

Sources/SwiftCrossUI/Values/FocusData.swift
```diff
@@ -0,0 +1,45 @@
+/// The information passed by ``View/focused(_:)`` to a backend.
+public struct FocusData: @unchecked Sendable {
```
stackotter: Why is this `@unchecked Sendable`? I feel like we should definitely avoid that (apologies if that was my suggestion originally, I really can't remember, but I don't believe that it was).

You can make the `any` properties sendable by adding `& Sendable` to the protocol requirement iirc, and you can make the closures `Sendable` with `@Sendable` of course. Alternatively, if this is only used on the main thread, you could consider marking it as `@MainActor` instead I think?

stackotter: This struct feels more like a fancy binding with extra metadata rather than just plain 'data'. And given that it's an internal detail of the framework and not just something borrowed from SwiftUI, I think we should rename it to something more descriptive. In my mind this is a kind of FocusState binding tailored specifically towards backends.

Potential name ideas:

- `BackendFocusStateBinding`

Ok I've run out of ideas already

stackotter: If this is public because it needs to be accessible to backends, then this should be marked `@_spi(Backends)` (I understand that you made these changes before Skye's introduction of spi)

Sources/SwiftCrossUI/Values/FocusData.swift
```diff
@@ -0,0 +1,45 @@
...
+    /// Solely used for calculating a hash.
+    private let match: any Hashable
+    /// A function to set the ``FocusState`` to ``match``.
+    public let set: () -> Void
+    /// A function to set the ``FocusState`` to unfocused.
+    public let reset: () -> Void
+
+    /// Whether the backend should remove focus from the view.
+    public let shouldUnfocus: Bool
+    /// Wheter the the ``FocusState``'s value matches ``match``.
```
stackotter: Typo of 'whether', and 'the' twice

Sources/SwiftCrossUI/Values/FocusData.swift
```diff
@@ -0,0 +1,45 @@
...
+    public let matches: Bool
```
stackotter: From my understanding so far, these two properties seem to only get read once by the backend's focus chain manager, generally during `register`, and then after that only the `set` and `reset` properties of this struct are used. And for both of these properties, the focus chain manager only uses them via calling `.contains(where: \.matches)` or `.contains(where: \.shouldUnfocus)`, so I don't see a reason to include the property as part of FocusData. Instead they should be passed to `register` separately.

In fact, I believe that these two properties should never be true at the same time, so perhaps we can represent them as single `Bool?` parameter called `focusOverride`. If nil, the manager would do nothing, if true the manager would focus the view, and if false the manager would unfocus the view. Then you could also consider splitting `register` into two separate methods with separate responsibilities. One would register focus data, and then other would focus/unfocus a view if needed.

Please let me know what you think of these ideas and if I've got any misconceptions about how this all fits together.

Sources/AppKitBackend/NSCustomWindow.swift
```diff
@@ -0,0 +1,122 @@
...
+        view.previousValidKeyView
+    }
+
+    public func makeKey(_ widget: Widget) {
+        makeFirstResponder(widget)
+    }
+
+    public func getParent(of widget: Widget) -> Widget? {
+        widget.superview
+    }
```
stackotter: I found it a bit confusing having this implementation in amongst all of the class inheritance override stuff. It might be neater to move this to an extension dedicated to adding `FocusChainManager` conformance instead

Sources/AppKitBackend/NSCustomWindow.swift
```diff
@@ -0,0 +1,122 @@
...
+            (window as! NSCustomWindow).persistentUndoManager
+        }
+    }
+}
+
+extension NSView: FocusChainParticipant {
+    public var canBeTabStop: Bool {
+        canBecomeKeyView
+    }
+}
```
stackotter: This should probably live in the `AppKitBackend+Focus.swift` file

Sources/AppKitBackend/AppKitBackend+Focus.swift
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    public func setFocusEffectDisabled(on widget: NSView, disabled: Bool) {
+        widget.focusRingType = disabled ? .none : .default
+    }
+}
+
+// MARK: - FocusState
+
+class FocusStateManager: NSObject {
```
stackotter: This is big enough that it'd probably be nice to have it in its own file

Sources/SwiftCrossUI/Backend/FocusChainManager.swift
```diff
@@ -0,0 +1,122 @@
...
+///
+/// To use:
+/// 1. Conform your coordinator to ``FocusChainManager``.
+/// 2. Implement ``makeKey(_:)`` to trigger the framework's focus mechanism.
+/// 3. Call ``selectTabStop(following:)`` or ``selectTabStop(preceding:)`` when
+///    the framework requests a focus change (e.g., via Tab or Shift-Tab).
+public protocol FocusChainManager {
+    associatedtype Widget: FocusChainParticipant
+
+    /// Returns the  widget following a given widget in the focus chain, suggested by the UI framework.
```
stackotter: accidentally included a double space

Sources/SwiftCrossUI/Backend/FocusChainManager.swift
```diff
@@ -0,0 +1,122 @@
...
+    /// This widget will be validated for visibility and focusability by the caller.
+    /// Used by the provided functions for easier ``View/focusable`` compatibility.
```
stackotter: `Used by FocusChainManager's helper methods.` (shouldn't have the 'for easier...' part imo)

same for the method below

Sources/SwiftCrossUI/Backend/FocusChainManager.swift
```diff
@@ -0,0 +1,122 @@
...
+    func closestValidStop(preceding view: Widget) -> Widget?
+
+    /// Makes a widget the "key view" or "first responder".
+    func makeKey(_ widget: Widget)
+
+    /// Returns the immediate parent.
+    func getParent(of widget: Widget) -> Widget?
+}
+
+/// A protocol to mark a widget as a focusability flag.
```
stackotter:
```
/// A protocol implemented by focus containers of backends that can't
/// natively toggle focusability of views.
```

Sources/SwiftCrossUI/Backend/FocusChainManager.swift
```diff
@@ -0,0 +1,122 @@
...
+/// Consumed by ``FocusChainManager``.
```
stackotter: `Respected by ...`

Sources/SwiftCrossUI/Backend/FocusChainManager.swift
```diff
@@ -0,0 +1,122 @@
...
+public protocol FocusabilityContainer {
+    /// Whether the container and its children can gain focus by keyboard navigation.
+    var focusability: Focusability { get }
+}
+
+/// Required by ``FocusChainManager``.
+/// Implement this protocol on widgets you want controlled by ``FocusChainManager``.
+/// Implementing it on the base Widget type of the Backend is recommended. e.g. `NSView`.
```
stackotter:
```swift
/// A protocol which backend widgets must conform to in order for
/// a backend to use them with a ``FocusChainManager``.
```

Sources/SwiftCrossUI/Backend/FocusChainManager.swift
```diff
@@ -0,0 +1,122 @@
...
+public protocol FocusChainParticipant: Equatable {
+    /// Whether a widget participates in the focus chain, i.e. if it can gain focus via keyboard navigation.
+    var canBeTabStop: Bool { get }
+    /// Whether a widget is hidden. Hidden widgets are skipped while selecting the next target.
+    var isHidden: Bool { get }
```
stackotter: Does FocusChainManager ever use this separately from canBeTabStop? Perhaps this can be merged into `canBeTabStop`'s responsibilities? (`canBeTabStop` would be false if the view is hidden)

Sources/SwiftCrossUI/Backend/FocusChainManager.swift
```diff
@@ -0,0 +1,122 @@
...
+}
+
+extension FocusChainManager {
+    public func findNextAllowedFocusTarget(
+        suggestion: Widget,
+        forward: Bool = true
+    ) -> Widget? {
+        var currentOption: Widget? = suggestion
```
stackotter: Maybe `currentCandidate` could be a more intuitive name for this?

Sources/SwiftCrossUI/State/FocusState.swift
```diff
@@ -0,0 +1,111 @@
...
+        }
+
+        /// The stored getter.
+        private let getValue: () -> Value
+        /// The stored setter.
+        private let setValue: (Value) -> Void
+        /// The stored resetter.
+        private let resetValue: () -> Void
+
+        /// Creates a binding with a custom getter and setter. To create a binding from
```
stackotter: `getter, setter, and resetter`

Sources/SwiftCrossUI/State/FocusState.swift
```diff
@@ -0,0 +1,111 @@
...
+        /// an `@FocusState` property use its projected value instead: e.g. `$myFocusStateProperty`
```
stackotter: I generally ignore the `@` when choosing between 'a' and 'an', so I'd write 'a'

Sources/SwiftCrossUI/State/FocusState.swift
```diff
@@ -0,0 +1,111 @@
...
+        implementation = StateImpl(initialStorage: Storage(nil))
+    }
+
+    public func update(with environment: EnvironmentValues, previousValue: FocusState<Value>?) {
+        implementation.update(with: environment, previousValue: previousValue?.implementation)
+    }
+
+    /// A property wrapper type that can read and write a value that indicates the current focus location.
+    @propertyWrapper
+    public class Binding {
```
stackotter: I think that this should be a struct. Is there a reason that it should be a class?

Examples/Sources/ControlFocusabilityTest/ControlFocusabilityApp.swift
```diff
@@ -0,0 +1,322 @@
...
+
+extension View {
+    func focusableIfSupported(_ disabled: Bool) -> some View {
+        #if canImport(AppKitBackend) || canImport(GtkBackend)
+            self
+                .focusable(disabled ? .unmodified : .disabled)
+        #else
+            self
+        #endif
+    }
```
stackotter, line 314: This probably isn't necessary given that support for other platforms isn't missing for a fundamental reason other than just not being merged yet. And the other focus modifiers are used without custom modifiers so it feels inconsistent

Examples/Sources/ControlFocusabilityTest/ControlFocusabilityApp.swift
```diff
@@ -0,0 +1,322 @@
...
+}
+
+struct ChildView: View {
+    var isFocused: FocusState<Int?>.Binding
```
stackotter: Can this be written as `@FocusState<Int?>.Binding var isFocused`?

stackotter: Calling this `isFocused` reads a bit funny. I think `focusState` probably works better?

Examples/Sources/ControlFocusabilityTest/ControlFocusabilityApp.swift
```diff
@@ -0,0 +1,322 @@
...
+    @State var isSliderFocusable = true
+    @State var isProgressSliderFocusable = true
+    @State var isTextFieldFocusable = true
+    @State var isSecureTextFieldFocusable = true
+    @State var isPickerStyleFocusable = true
+    @State var isFlavorPickerFocusable = true
+    @State var isDatePickerStyleFocusable = true
+    @State var isDatePickerFocusable = true
+
+    @FocusState var focused: Int?
```
stackotter: I feel like this makes more sense if it's called `focusState`, because focused sounds more like a boolean value imo

Sources/AppKitBackend/AppKitBackend+Focus.swift
```diff
@@ -0,0 +1,146 @@
...
+                if let lastFocused = windowFocusState.lastFocused {
+                    handleFocusChange(of: ObjectIdentifier(lastFocused), toState: false)
+                }
+                windowFocusState.lastFocused = responder
+            } else if !windowFocusState.shouldSkip {
+                if let lastFocused = windowFocusState.lastFocused {
+                    handleFocusChange(of: ObjectIdentifier(lastFocused), toState: false)
+                }
+                windowFocusState.lastFocused = responder
+            } else if windowFocusState.shouldSkip {
```
stackotter: Even though it works this way, I feel like the `shouldSkip` check should be a guard. Perhaps even above this whole if statement? Then the meaning of it would also be much more clear. Imo the comments about FooTextField should be attached to the line where we set shouldSkip to true, instead of here, because in theory the reason that we're skipping here is just because we were told to, and the mechanism could be used by other workarounds in future.

It could also make sense to rename `shouldSkip` to `shouldSkipNextFocusUpdate` for more clarity?

Sources/SwiftCrossUI/Views/Modifiers/FocusModifier.swift
```diff
@@ -0,0 +1,137 @@
...
+    /// Setting to `nil` on ``WinUIBackend`` causes the first focusable widget to gain focus
+    /// due to WinUI not supporting setting an "unfocused" state.
+    public func focused<Value: Hashable>(
+        _ focusBinding: FocusState<Value?>.Binding,
+        equals match: Value
+    ) -> some View {
+        EnvironmentModifier(self) { environment in
+            environment.with(
+                \.focusObservers,
+                environment.focusObservers + [
```
stackotter: These aren't really observers imo. If you go with my suggestion of renaming and refactoring FocusData to be BackendFocusStateBinding, then you could rename the environment property to `focusBindings`. Otherwise, if you really want these to be observers, then they should just hold `notifyFocused` and `notifyUnfocused` closures and nothing else; and then when you create the observers you would implement the closures such that they perform this specific setting and resetting behaviour.

Actually, now that I think about it, that might be even nicer than my `BackendFocusStateBinding` idea.

Sources/SwiftCrossUI/Views/Modifiers/FocusModifier.swift
```diff
@@ -0,0 +1,137 @@
+extension View {
+    /// Controls the focusability of a view.
+    /// Only affects out of the box interactable Views.
```
stackotter: Is this part about 'out of the box' interactable views still true? Once we introduce support for adding custom tab stops I feel like this will become false, and then we'd forget to update the docs. And without support for custom tab stops, that behaviour you've described is the only possible behaviour I think?

stackotter (changes requested):
Thank you for your patience on this PR. The implementation is for the most part really clean; I love the FocusChainManager helper and the way that you've architected AppKitBackend's implementation of the protocol. I said "that's so cool" out loud while reading through the FocusChainManager implementation hahah (probably referring to how useful it is given its minimal implementation requirements).

---

As I reviewed the changes and my understanding of the code changed, my opinions on FocusData and focusObservers have changed as well (which you'll see through the comments), so I'll detail my final thoughts here;

I think that FocusData should be renamed to WidgetFocusObserver and made `@_spi(Backends)`. It should only store a `didFocus` closure and a `didUnfocus` closure. When creating the WidgetFocusObserver instances in the relevant focus modifier, you can then implement the "set on focus" and "reset on focus" logic by implementing those two closures.

The part that I haven't changed my mind on is that `matches` and `shouldUnfocus` should not be part of WidgetFocusObserver. I've detailed my reasoning in one of the comments, but essentially, these two properties have nothing to do with observation, and are never used on individual WidgetFocusObserver instances (i.e. they're only used when doing `.contains(where: \.matches)` on arrays of observers). And from my understanding, they should never both be true at the same time, so we can replace them with a single `Bool?` property. This property could be propagated in the environment alongside `focusObservers` and could be called something like `focusOverride` or `forcedFocusOverride`? It could be make sense to make an enum called `Focus` that has two cases `focused` and `unfocused` so that the property can be `Focus?` instead of `Bool?`, because `focusOverride = false` reads poorly imo.

Then, with those two changes made, you can create a separate backend method called `setFocus(of:to:)` that sets the focus of a given widget to the given focus state. If `focusOverride` is nil then you can just skip calling the method, so the method's second parameter should just be of type `Focus`.

If you're unsure about whether I've changed my mind on a particular suggestion that might relate to this FocusData stuff, please let me know. Apologies if they cause any confusion, but I have to wrap up reviewing this for now so I don't have time to go back and update all of the comments.

I think I mistakenly referred to a `register` method a few times when I meant `registerFocusObservers`, hopefully that's not too confusing.

Sources/SwiftCrossUI/Values/Focusability.swift
MiaKoring: The issue with SwiftUI’s bool approach it, that a true adds another focus stop, even if the inner view is focusable oob and false disables it’s focusability, that’s the reason for making it an enum with unmodified. (the enabled state can be added in the future, but aside from AppKitBackend iirc it’s a bit complicated (and weird))

I would like to keep it as Focusability. (especially since the true case wouldn’t be a focusability modification)

Sources/SwiftCrossUI/Values/FocusData.swift
MiaKoring: I believe I remember (it’s been some time for me too) an issue with mainactor for either gtk or WinUI, but as this is only for AppKit I’ll just make it mainactor isolated and if there’s now an issue with the other backends I’ll deal with it then 👍

Sources/SwiftCrossUI/Backend/FocusChainManager.swift
MiaKoring: Valid point, though I believe having isHidden separately makes it easier to use FocusChainManager, cause you don’t have to know how it works for using it. And since it is available for third party backends to use, I believe that can matter.

Sources/SwiftCrossUI/State/FocusState.swift
MiaKoring: If it’s a struct, swift complains when I want to write to it in the `View/focused(_:)` implementation. even though it just passes them through. If you can think of a way to make that work with struct, sure. (Maybe I got tunnel view rn)

Examples/Sources/ControlFocusabilityTest/ControlFocusabilityApp.swift
MiaKoring: it is necessary, because WinUI doesn’t support it. Because WinUI is a bitch :) 
The function I would use in WinUI to implement FocusChainManager doesn’t work.

Examples/Sources/ControlFocusabilityTest/ControlFocusabilityApp.swift
MiaKoring: made it can

Sources/AppKitBackend/AppKitBackend+Focus.swift
```diff
@@ -70,6 +39,15 @@ class FocusStateManager: NSObject {
         {
             widget.window?.makeFirstResponder(widget)
         }
+
+        if
```
stackotter: I think this can be an `else if`?

Sources/AppKitBackend/AppKitBackend+Focus.swift
```diff
@@ -70,6 +39,15 @@ class FocusStateManager: NSObject {
...
+            focus == .unfocused,
+            let window = widget.window,
+            window.firstResponder == widget || textFieldsTextViewIsFocused(field: widget)
+        {
+            _ = window.makeFirstResponder(nil)
+            return
```
stackotter: Unnecessary early return

Sources/AppKitBackend/FocusStateManager.swift
```diff
@@ -0,0 +1,73 @@
...
+        }
+    }
+
+    private func handleFocusChange(of identifier: ObjectIdentifier, toState isFocused: Bool) {
+        guard let data = focusData[identifier] else { return }
+
+        if isFocused {
+            data.forEach { binding in
+                binding.didGainFocus()
+            }
```
stackotter: I'm not sure why I missed this last time, but these should be for loops instead of `forEach`. Also, rename the loop variable to `observer` (instead of `binding`)

Sources/SwiftCrossUI/Backend/BackendFeatures/FocusHandling.swift
```diff
@@ -13,26 +13,29 @@ extension BackendFeatures {
...
+        /// Makes a widget gain or lose focus.
+        func setFocus(of widget: Widget, to focus: SwiftCrossUI.Focus)
     }
 
     /// Backend methods for disabling focusability of a subtree.
     ///
     /// Used by ``View/focusable(_:)``.
     public protocol FocusDisabling: Core {
-        /// Create a container controlling the focusability of the widgets below it.
+        /// Create a container controlling the focusability of the widget's children.
```
stackotter: `of its children` (and same for the doc comment below)

Sources/SwiftCrossUI/Backend/BackendFeatures/FocusHandling.swift
stackotter: You've renamed the file to `FocusHandling` but the protocol is still called `Focus`. I realise that it was to avoid a naming clash, but maybe we can also rename the protocol to match in order to avoid confusion?

Sources/SwiftCrossUI/Backend/BackendHelpers.swift
```diff
@@ -0,0 +1,28 @@
+enum BackendHelpers {
+    /// Sets the ``WidgetFocusObserver``s from the environment on a widget.
+    static func setWidgetFocusObservers<Backend: BackendFeatures.Focus>(
```
stackotter: This method does more than just set the focus observers, so it should be renamed to reflect that. Maybe `updateWidgetFocusProperties`?

Sources/SwiftCrossUI/Backend/BackendHelpers.swift
```diff
@@ -0,0 +1,28 @@
...
+            on: widget.into()
+        )
+
+        backend.setFocusEffectDisabled(
+            on: widget.into(),
+            disabled: environment.focusEffectDisabled
+        )
+    }
+
+    /// Makes a widget gain or lose focus.
```
stackotter, line 19: Can also leave the widget's focus unchanged if `nil`

Sources/SwiftCrossUI/State/FocusState.swift
stackotter: What does Swift complain about?

Sources/SwiftCrossUI/Backend/FocusChainManager.swift
stackotter: My main issue with having it separate is that `canBeTabStop` should really be `false` when `isHidden` is true anyway. In AppKit's case I believe that `canBecomeKeyView` already takes visibility into account. Also, I don't think SwiftCrossUI ever instructs backends to hide widgets, so hidden widgets would be a very backend-specific thing. The only case where I can see AppKitBackend setting `isHidden` is in button style application (which might actually mean that we have incorrect behaviour for plain buttons atm?)

Sources/SwiftCrossUI/Environment/EnvironmentValues.swift
```diff
@@ -479,7 +479,10 @@ extension EnvironmentValues {
     public var deviceClass: DeviceClass { backend.deviceClass }
 
     /// All observers set by ``View/focused(_:)`` in the environment.
-    @Entry public var focusObservers: [FocusData] = []
+    @Entry public var widgetFocusObservers: [WidgetFocusObserver] = []
+
+    /// A value used to make widgets programmatically gain or lose focus.
+    @Entry public var focusOverride: Focus?
```
stackotter: Should these two maybe be `@_spi(Backends)`?

Sources/SwiftCrossUI/Layout/ViewLayoutResult.swift
```diff
@@ -10,7 +10,7 @@ public struct ViewLayoutResult {
     public var preferences: PreferenceValues
     /// If a view can ever get focused out of the box, this needs to be set to true,
     /// to be compatible with `@FocusState`.
```
stackotter:
```swift
/// If `true` then `ViewGraphNode` skips a bunch of focus-related housekeeping.
/// If a view can ever become focusable (even if it's not focusable right now)
/// then this must be set to `false`.
```

Sources/SwiftCrossUI/Values/Focus.swift
```diff
@@ -0,0 +1,16 @@
+/// An enum, stored in ``EnvironmentValues``, indicating
+/// whether a widget should programmatically gain or lose focus.
+public enum Focus: Sendable {
+    case focused
+    case unfocused
+}
+
+extension Optional where Wrapped == Focus {
+    func modify<Value: Hashable>(with state: Value?, match: Value) -> Self {
```
stackotter, line 12: Document this because it's quite a strange method out of context, perhaps;

```swift
/// Modifies a focus override based on a focus state and the focus value
/// associated with a widget. If the value and the state match, then the
/// focus override gets set to `focused`, 
```

Sources/SwiftCrossUI/Values/Focus.swift
```diff
@@ -0,0 +1,16 @@
...
+        guard self != .focused else { return self }
+
+        if state == match { return .focused }
+        if state != nil { return .unfocused }
+        return nil
```
stackotter: Why does a `nil` state mean that the focus doesn't get overridden? In my mental model, a nil focus state means that all widgets linked to that focus state should have their focus set to `unfocused`

Sources/SwiftCrossUI/Values/Focusability.swift
stackotter: I meant that it feels like it's specifically designed for the modifier, while its name makes it feel like it would be representing the actual focusability of a view (which is always true or false). It kind of feels like an overlay

Sources/SwiftCrossUI/Values/WidgetFocusObserver.swift
```diff
@@ -0,0 +1,16 @@
+/// The information passed by ``View/focused(_:)`` to a backend.
```
stackotter: This doc comment should get updated to reflect the type's new role, maybe;

```swift
/// A focus state observer passed by ``View/focused(_:)`` to backends.
```

Sources/SwiftCrossUI/Values/WidgetFocusObserver.swift
```diff
@@ -0,0 +1,16 @@
...
+@MainActor
+public struct WidgetFocusObserver: Sendable {
+    /// A function to set the ``FocusState`` to ``match``.
```
stackotter: `/// A function called when the target widget gains focus.`

Sources/SwiftCrossUI/Values/WidgetFocusObserver.swift
```diff
@@ -0,0 +1,16 @@
...
+    public let didGainFocus: () -> Void
+    /// A function to set the ``FocusState`` to unfocused.
```
stackotter: `/// A function called when the target widget loses focus.`

Sources/SwiftCrossUI/ViewGraph/ViewGraphNode.swift
```diff
@@ -253,27 +253,25 @@ public class ViewGraphNode<NodeView: View, Backend: BaseAppBackend>: Sendable {
     /// - Returns: The most recently computed layout. Guaranteed to match the
     ///   result of the last call to ``computeLayout(with:proposedSize:environment:)``.
     public func commit() -> ViewLayoutResult {
-        if currentLayout?.shouldSetFocusData == true {
+        if currentLayout?.isNeverFocusable == false {
```
stackotter: Move this if after the `guard let currentLayout` so that we don't have to unwrap the `currentLayout`. It should never be `nil` in commit, so the first thing in the commit method should be the `guard let currentLayout`.

Then you can also use `!currentLayout.isNeverFocusable`

Sources/SwiftCrossUI/Views/Modifiers/FocusedModifier.swift
```diff
@@ -0,0 +1,67 @@
...
+                            focusBinding.reset()
+                        }
+                    )
+                ]
+            )
+            .with(
+                \.focusOverride,
+                environment.focusOverride != .focused
+                    ? focusBinding.wrappedValue ? .focused : .unfocused
+                    :.focused
```
stackotter: Put a space after the colon

Sources/SwiftCrossUI/Views/Modifiers/FocusedModifier.swift
```diff
@@ -0,0 +1,67 @@
...
+                        didLoseFocus: {
+                            focusBinding.reset()
+                        }
+                    )
+                ]
+            )
+            .with(
+                \.focusOverride,
+                environment.focusOverride != .focused
+                    ? focusBinding.wrappedValue ? .focused : .unfocused
```
stackotter: Maybe put parentheses around this inner ternary expression to make things extra clear

Sources/SwiftCrossUI/Views/_BuiltinPickerImplementation.swift
```diff
@@ -80,26 +80,24 @@ public struct _BuiltinPickerImplementation: TypeSafeView {
...
+            )
         } else if
-            !environment.focusObservers.isEmpty ||
-            environment.focusEffectDisabled
+            !environment.widgetFocusObservers.isEmpty ||
+            environment.focusEffectDisabled ||
+            environment.focusOverride != nil
         {
             logger.warnOnce("\(Backend.self) doesn't support focus control/tracking.")
         }
```
stackotter: Even with the helpers this ends up being a moderately complicated duplicate of the code in ViewGraphNode. Perhaps you could add another helper to BackendHelpers that encapsulates "applying focus related properties from the environment to a widget"? Within that method it'd also handle casting to `BackendFeatures.Focus` and warning on no support

stackotter (changes requested):
Thanks for those resolutions; I think most of my requests this time are relatively minor in comparison

Sources/SwiftCrossUI/State/FocusState.swift
MiaKoring: that it's a let constant. even though it's just passing it through.

Sources/SwiftCrossUI/Backend/FocusChainManager.swift
MiaKoring: why would a plain button be hidden just by applying the plain button style?

Sources/SwiftCrossUI/State/FocusState.swift
MiaKoring: I'm gonna try making it a struct with using the closures instead.

And see if it still works as a property wrapper

Sources/SwiftCrossUI/Values/Focus.swift
MiaKoring: you’re right. not sure what I was thinking

Sources/SwiftCrossUI/Backend/FocusChainManager.swift
MiaKoring: if you’re referring to the button background appkit uses, the background is never focusable. 
But yes, I guess we could remove it entirely for styles not using it. I wouldn’t expect the case of switching between bordered and unbordered button styles to be too common xD. I’ll make a PR for it later or tomorrow (if I remember lol, should I forget, feel free to poke me rq xD)

Sources/SwiftCrossUI/State/FocusState.swift
MiaKoring: just to have the context here too: I plain forgot the nonmutating keyword exists xD

Sources/SwiftCrossUI/ViewGraph/ViewGraphNode.swift
```diff
@@ -253,32 +253,19 @@ public class ViewGraphNode<NodeView: View, Backend: BaseAppBackend>: Sendable {
...
-                logger.warnOnce("\(Backend.self) doesn't support focus control/tracking.")
-            }
-        }
-
         guard let currentLayout else {
             logger.warning("layout committed before being computed, ignoring")
             return .leafView(size: .zero)
         }
 
+        if currentLayout.isNeverFocusable == false {
```
stackotter: `!currentLayout.isNeverFocusable`

stackotter (changes requested):
Just one tiny request this time, everything else looks good now!

stackotter (approved):

PR #655, opened 2026-06-26
Bump to swift-winui v0.2.1 to fix main actor hiccups

stackotter:
Before swift-winui v0.2.1 (which I just released), the MainRunLoopTickler would often end up waiting for a full second before servicing the main run loop again, because Swift's RunLoop would return the next deadline as `Date.distantFuture` (and the tickler then limited that to its maximum delay of 1 second).

The swift-winui fix was performed in https://github.com/moreSwift/swift-winui/pull/3

### DIFF
--- a/Examples/Package.resolved
+++ b/Examples/Package.resolved
@@ -356,8 +356,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/moreSwift/swift-winui",
       "state" : {
-        "revision" : "7cce3fcd0b55c48e6d8993fd2c7c392e8bc945c4",
-        "version" : "0.2.0"
+        "revision" : "df7642fbb88e23a9cc1bbd366c7d6d3780fd0251",
+        "version" : "0.2.1"
       }
     },
     {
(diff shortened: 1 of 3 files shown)


ktraunmueller:
@stackotter  I haven't looked into the details of this merge request, but the mention of `MainRunLoopTickler` caught my attention: I had been running into [similar problems myself a while ago](https://github.com/thebrowsercompany/swift-winrt/issues/213) and the solution was to replace the `MainRunLoopTickler` altogether with a [SwiftApplication](https://github.com/ktraunmueller/winrt-monorepo/blob/main/swift-winui/Sources/WinUI/Application/SwiftApplication.swift) app base class.

stackotter:
Yeah... after diagnosing the issue and making this fix, I discovered your thebrowserco/swift-winrt issue about the same problem 😅 At some point I will probably merge in the solution that thebrowserco unarchived swift-winui to publish. Their "archived" repos keep catching me out hahah; if I knew they'd unarchive it to publish changes then I probably would've checked it for related changes first.

PR #656, opened 2026-06-26
Android date pickers

bbrk24:
This PR implements preliminary DatePickers on Android. Currently, only the `.compact` style is universally available -- `.graphical` is too big for many devices, and `.wheel` requires XML. Currently, the date picker widget is a FrameLayout that contains an instance of AbstractDatePicker (whose concrete subclasses include GraphicalDatePicker and CompactDatePicker); when #491 is resolved, the FrameLayout can be removed, but for now a single widget needs to be able to handle changing its style on the fly.

Since neither TimePicker nor TimePickerDialog can set a finer resolution than minutes, the "second" bit of DatePickerComponents is currently ignored.

### DIFF
--- a/Sources/AndroidBackend/AndroidBackend.swift
+++ b/Sources/AndroidBackend/AndroidBackend.swift
@@ -4,6 +4,7 @@ import SwiftCrossUI
 import AndroidKit
 import AndroidGraphics
 import AndroidBackendShim
+import Mutex
 
 // Many force tries are required for the Android backend but we don't really want them
 // anywhere else so just disable the lint rule at a file level.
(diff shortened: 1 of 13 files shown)


Sources/AndroidBackend/AndroidBackend.swift
```diff
@@ -90,20 +90,23 @@ public final class AndroidBackend: BackendFeatures.BaseStubs {
...
-            case 3: .tv
-            case 4: .watch
-            case let x: fatalError("helpers.getDeviceClass returned unexpected value \(x)")
-        }
+    // Thread-safety note: `_supportedDatePickerStyles` is only set in `computeRootEnvironment`;
+    // `supportedDatePickerStyles` must be `nonisolated` to satisfy the DatePickers protocol.
+    private nonisolated(unsafe) var _supportedDatePickerStyles: [DatePickerStyle] = [
+        .automatic,
+        .compact
+    ]
```
stackotter: I think we should wrap this in a Mutex to avoid the `unsafe`, because to my knowledge this variable shouldn't be a performance bottleneck anyway, and eliminating avoidable `unsafe` usage makes it easier to audit the codebase

stackotter (changes requested):
The code all looks great, and the compact date picker style works well functionally.

Unfortunately there appear to be some internal alignment/sizing issues with the compact date picker style's buttons.

The buttons start out aligned with each other, but misaligned with the general SwiftCrossUI layout;

Image

Then when I changed the selected date, the buttons lost their vertical alignment with each other (which got fixed after a state update):

Image

bbrk24:
The misalignment comes from the platform-independent implementation of DatePicker.

https://github.com/moreSwift/swift-cross-ui/blob/4ad7142af868b5ab19acbc44deb66bf52cdedf7d/Sources/SwiftCrossUI/Views/DatePicker.swift#L104-L110

Notice there's no `(spacing: 0)` on that HStack. When you provide a label, usually you want spacing between the label and the widget. But we aren't providing a label in the ControlsExample, so there's extraneous spacing on the left side of the date picker widget.

Sources/AndroidBackend/AndroidBackend.swift
bbrk24: Added a Mutex here.

stackotter (approved):
Looks good now 👍

PR #664, opened 2026-07-03
Set participateInStackLayoutsWhenEmpty to false for EmptyView

### LINKED ISSUE #416: Extra spacing in VStack with inactive if
If you comment out the `if false` block here, the amount of padding at the bottom of the VStack is reduced:

```swift
VStack {
  Text("foo")

  if false {
    Text("bar")
  }
}
```

The presence of an `if false` should not affect layout at all

bbrk24:
Fixes #416 

Turns out EmptyView and empty OptionalView set `participateInStackLayoutsWhenEmpty` to true. Setting this to false removes the extra padding observed in #416 and #656.

For the record, I tested empty ForEach as well, and that seems to already work correctly.

### DIFF
--- a/Sources/SwiftCrossUI/Views/EmptyView.swift
+++ b/Sources/SwiftCrossUI/Views/EmptyView.swift
@@ -51,7 +51,7 @@ public struct EmptyView: View, Sendable {
         environment: EnvironmentValues,
         backend: Backend
     ) -> ViewLayoutResult {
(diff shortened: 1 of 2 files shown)


bbrk24:
CI failed for Android because it failed to install the NDK. I can't exactly make heads or tails of the error — it seems to be complaining something isn't a valid ZIP file — but I can't imagine it's my fault.

stackotter (approved):

PR #666, opened 2026-07-08
Gtk3Backend: Implement sheet support

stackotter:
This PR introduces sheet support for Gtk3Backend, and fixes a correctness issue with GtkBackend's sheet implementation (which I discovered while testing the Gtk3Backend implementation).

The Gtk3Backend sheet implementation was copied largely from the GtkBackend implementation with some minor changes required to get things compiling under Gtk 3. In addition to API changes, I had to switch the `Sheet` type from a `Window` subclass to a `Dialog` subclass to get proper modal behaviour (otherwise the sheet was just treated as a regular window by the window manager), and I had to fix an issue in the GObject `deinit` that was spamming warnings about trying to disconnected handlers that have already been disconnected.

The correctness issue was that `Open Sheet` -> `Present a nested sheet` -> `Dismiss` -> `Dismiss` would result in `Presented a nested sheet` being printed (because the nested sheet's handler for non-programmatic dismisses would get invoked when the root sheet got dismissed). The underlying cause was that dismissing a nested sheet didn't remove the parent's reference to the nested sheet, so the parent would think that it still had a nested sheet.


MiaKoring:
The fail on android here is caused by a missing @_spi(Backends) import SwiftCrossUI in Android Backend’s inspection modifiers file, I had the same issue in my Buttons PR

PR #668, opened 2026-07-09
Android menu buttons

bbrk24:
I'll have to double-check BackendFeatures, but I think this is the last thing Android can definitely do without swift-bundler work.


stackotter (approved):
Code looks great, and it works well on-device; thanks for implementing this 🙏

PR #670, opened 2026-07-10
Don't run platforms' CI on PRs it's not needed

bbrk24:
Testing this may unfortunately require multiple PRs, due to the nature of CI.

### DIFF
--- a/.github/workflows/build-test-and-docs.yml
+++ b/.github/workflows/build-test-and-docs.yml
@@ -37,7 +37,7 @@ jobs:
           if [ '${{ github.ref }}' == 'refs/heads/main' ]; then
             echo '${{ matrix.platform }}=true' >>"$GITHUB_OUTPUT"
           elif
-            git diff --exit-code --name-only HEAD '${{ github.base_ref }}' -- \
+            git diff --exit-code --name-only HEAD "origin/$GITHUB_BASE_REF" -- \
               Examples/ .github/actions/ .github/workflows/build-test-and-docs.yml Package.swift \
               Package.resolved Sources/SwiftCrossUI*/ Sources/DefaultBackend/ Sources/Bundler.toml \
               ${{ matrix.paths }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build-test-and-docs.yml
```diff
@@ -10,13 +10,58 @@ on:
...
+    strategy:
+      matrix:
+        platform: [android, linux, macos, uikit, windows]
+        include:
+          - platform: android
+            paths: 'Sources/AndroidBackend/ Sources/AndroidBackendShim/'
+          - platform: linux
+            paths: 'Sources/*Gtk*/ Tests/ Scripts/test.sh vcpkg.json'
+          - platform: macos
+            paths: 'Sources/AppKitBackend/ Sources/*Gtk*/ Tests/ Scripts/test.sh vcpkg.json'
```
stackotter: Do we have to track vcpkg.json on macOS?

stackotter (changes requested):
Looks good, thanks for working on this. Just left one question, but happy to merge other than that (pending CI of course).

Oh and, what happens to the assemble-docs step when a job that it depends on gets skipped? Does it get skipped itself? Or does it perhaps count as failed/cancelled?

.github/workflows/build-test-and-docs.yml
bbrk24: I'm not actually sure. I just saw that macOS CI installs GTK and added it here.

bbrk24:
> Oh and, what happens to the assemble-docs step when a job that it depends on gets skipped? Does it get skipped itself? Or does it perhaps count as failed/cancelled?

If all the dependent jobs aren't run, I think it never gets queued in the first place. But if we wanted to be sure we'd have to open a second PR after this one to test it.

.github/workflows/build-test-and-docs.yml
stackotter: vcpkg is a Windows package manager thing, so it should only be needed on Windows

stackotter (approved):

stackotter:
I'm going to merge this and then keep an eye on CI for the next few PRs to make sure that it's actually working correctly

PR #672, opened 2026-07-11
Add public accessor for getting ResolvedMenu from View

xwxfox:
All this does is simply add a public way to get ResolvedMenu from a view.

i specifically needed it for something i was working on, but couldnt find a way to do it "correctly"

I was unsure whether it was best to make the current resolve func public, or add a new one prefixed with _ to signal it isnt a "normal feature everyone will use" - so please tell me if it should be put somewhere else / if its completely out of scope :3

(The other files changed were just formatting from swiftformat)

I'd be surprised if this would break anything, but i tested anyways uwu

Tested on:
Arch Linux x64 (7.0.3-zen)
macOS 27.0 arn64 (m2)
Winslop 11 25H2 (26200)

### DIFF
--- a/Sources/SwiftCrossUI/Views/View.swift
+++ b/Sources/SwiftCrossUI/Views/View.swift
@@ -245,4 +245,13 @@ extension View {
     }
 
     public var _asMenuItems: [MenuItem] { body._asMenuItems }
+
+    /// Resolves this view's menu content to the representation used by backends.
+    ///
+    /// This is the same resolution applied to ``Menu`` content and scene ``Commands``.
+    /// - Returns: The resolved menu.
+    @MainActor
+    public func _resolvedMenuContent() -> ResolvedMenu {
+        Menu.resolve(items: _asMenuItems)
+    }
 }
--- a/Sources/SwiftCrossUI/Views/View.swift
+++ b/Sources/SwiftCrossUI/Views/View.swift
@@ -251,7 +251,7 @@ extension View {
     /// This is the same resolution applied to ``Menu`` content and scene ``Commands``.
     /// - Returns: The resolved menu.
     @MainActor
-    public func _resolvedMenuContent() -> ResolvedMenu {
+    @_spi(Backends) public func resolvedMenuContent() -> ResolvedMenu {
         Menu.resolve(items: _asMenuItems)
     }
 }
--- a/Sources/Gtk/GObject.swift
+++ b/Sources/Gtk/GObject.swift
@@ -23,7 +23,7 @@ open class GObject: GObjectRepresentable {
         for (id, _) in signals {
             g_signal_handler_disconnect(gobjectPointer, id)
         }
-  
+
         g_object_unref(gobjectPointer)
     }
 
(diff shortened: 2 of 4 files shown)


bbrk24:
> i specifically needed it for something i was working on, but couldnt find a way to do it "correctly"

I'm curious what it is that you need it for. ResolvedMenu is in the Backend folder, and I'm honestly surprised it's not `@_spi(Backends)`.

xwxfox:
> I'm curious what it is that you need it for

I was working on getting cross platform tray icons to work specifically allowing to use the patterns from SCUI but without trying to implement it directly in the scui codebase as im not familiar enough with it yet - so i wouldnt really feel comfy throwing up a PR for something like it when i dont know the ins n outs yet

so i decided to hack on it with an external package - but that ofc requires me to somehow yoink ResolvedMenu - but i couldnt find any other "happy paths" for it

```swift
/// Sets the tray icon's context menu using SCUI views
///
/// - Parameter content: A SCUI view builder describing the menu.
public func setMenu<Content: View>(
    @ViewBuilder _ content: @escaping @MainActor () -> Content
) {
    menuBuilder = { content()._resolvedMenuContent() }
    backend.setHasMenu(true)
    backend.invalidateMenu()
}
```

and then ofc consumer-facing like:
```swift
tray.setMenu {
    Button("hi") {
        print("hiii")
    }
    Toggle("Mute", isOn: bindingForMuted())
    Divider()
    Menu("Advanced") {
        Button("meow") {
            print(":3")
        }
    }
    Divider()
    Button("Quit") {
    exit(0)
    }
}
```


My janky stuff does seem to work on the platforms i tested on (listed in main PR body) - but it isnt at the level where id be comfy with throwing it into a PR yet

Sources/SwiftCrossUI/Views/View.swift
```diff
@@ -245,4 +245,13 @@ extension View {
     }
 
     public var _asMenuItems: [MenuItem] { body._asMenuItems }
+
+    /// Resolves this view's menu content to the representation used by backends.
+    ///
+    /// This is the same resolution applied to ``Menu`` content and scene ``Commands``.
+    /// - Returns: The resolved menu.
+    @MainActor
+    public func _resolvedMenuContent() -> ResolvedMenu {
```
stackotter: We've recently started using SPI to hide internal details that are intended for use by backends. If you annotate this function with `@_spi(Backends)` then the API will be hidden for anyone using SwiftCrossUI the usual way, but backend implementations and custom native view implementations can access the API by annotating their `SwiftCrossUI` import with `@_spi(Backends)`.

stackotter (changes requested):
This seems like a sensible thing to want to do, especially when implementing custom native views. We use the `Backends` SPI to hide internal details (even those that aren't directly used by backends). If you mark your new API with the spi annotation that I suggested, then I'd be happy to merge.

Sources/SwiftCrossUI/Views/View.swift
xwxfox: Done ;3

Also finally got my cross-platform tray lib working-ish
image
image
image

stackotter (approved):
Looks good now; I'll merge after the CI passes

PR #678, opened 2026-07-18
WinUIBackend: queue alert presentation asynchronously (#493)

raliclo:
Using queue to solve 
- #493 (Fixed): WinUIBackend may crash when an environment action is called too early

### DIFF
--- a/Sources/WinUIBackend/WinUIBackend.swift
+++ b/Sources/WinUIBackend/WinUIBackend.swift
@@ -113,6 +113,10 @@ public final class WinUIBackend:
 
     var internalState: InternalState
     nonisolated(unsafe) private var dispatcherQueue: WinAppSDK.DispatcherQueue?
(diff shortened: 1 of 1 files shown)


Sources/WinUIBackend/WinUIBackend.swift
```diff
@@ -1502,6 +1506,120 @@ public final class WinUIBackend:
         }
     }
 
+    public func createAlert() -> Alert {
+        ContentDialog()
+    }
+
```
stackotter, line 1512: All of WinUIBackend's alert logic has been moved to `Sources/WinUIBackend/WinUIBackend+Alerts.swift` since your original PR, and the concurrent alerts deadlock has been solved. You'll have to reconcile your changes with the ones I've made since your original PR. I believe that this PR will only have to solve the issue about alerts being shown too early now

stackotter (changes requested):

Issue #204, opened 2025-06-27
WinUIBackend: Update to latest stable WinUI


stackotter:
We should eventually update WinUIBackend to use the latest version of WinUI 3. If I remember correctly, it's currently using the WinUI from WindowsAppSDK 1.5 preview 1. This probably involves just regenerating the 4 or 5 WinUI-related Swift packages against a newer WindowsAppSDK version.

NSBrianWard:
+1 to this. It looks like 1.5 doesn't run on Windows 11? Is that correct?

stackotter:
I have tested SwiftCrossUI on Windows 11, and it worked for me. But let me know if it doesn't work on your end

melgu:
I have to manually install the WinUI components (and related ones) in Visual Studio, but for me it works on Windows 11.

clarkezone:
Would be great to see an update to winui3 stable

PR #679, opened 2026-07-18
WinUIBackend: filter backdrop console noise (#204)

raliclo:
Temp fixes for WinSDK1.5 console noise issue.
It might not be necessary for WinSDK2.


stackotter:
Do you have an example of some code that would produce such noise to be produced? I don't recall seeing much console noise with WinUIBackend myself

PR #680, opened 2026-07-18
Implement folder selection in open dialogs for GTK, Gtk3, and WinUI b…

raliclo:
Implement folder selection in open dialogs for GTK, Gtk3, and WinUI backends

Add OpenDialogOptions.singleKindSelectionMode to express whether an open
dialog should select files or directories on backends that only support
one kind of item per dialog. Files are preferred when both are allowed,
preserving the existing behavior of the GTK and WinUI backends.

- GtkBackend/Gtk3Backend: use FileChooserAction.selectFolder instead of
  .open when the dialog is directories-only
- WinUIBackend: split showOpenDialog into showFileOpenDialog and a new
  FolderPicker-based showFolderOpenDialog for directories-only dialogs;
  selecting multiple folders is unsupported by FolderPicker and guarded
  by a precondition
- Add OpenDialogOptionsTests covering the selection mode logic

Also fix the Gtk3 module build on Windows:
- Use gulong directly for signal handler IDs instead of converting
  through UInt
- Rename Gtk3CHelpers' max() to static scui_max() to avoid symbol
  clashes

Note: Used Claude Code as LLM Coding support

### DIFF
--- a/Sources/WinUIBackend/WinUIBackend.swift
+++ b/Sources/WinUIBackend/WinUIBackend.swift
@@ -1508,6 +1508,30 @@ public final class WinUIBackend:
         window: Window?,
         resultHandler handleResult: @escaping (DialogResult<[URL]>) -> Void
     ) {
+        switch openDialogOptions.singleKindSelectionMode {
+            case .files:
+                showFileOpenDialog(
+                    fileDialogOptions: fileDialogOptions,
+                    openDialogOptions: openDialogOptions,
+                    window: window,
+                    resultHandler: handleResult
+                )
+            case .directories:
+                showFolderOpenDialog(
+                    fileDialogOptions: fileDialogOptions,
+                    openDialogOptions: openDialogOptions,
+                    window: window,
+                    resultHandler: handleResult
+                )
+        }
+    }
+
+    private func showFileOpenDialog(
+        fileDialogOptions: FileDialogOptions,
+        openDialogOptions: OpenDialogOptions,
+        window: Window?,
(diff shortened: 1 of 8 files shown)


Sources/WinUIBackend/WinUIBackend.swift
```diff
@@ -1550,6 +1574,42 @@ public final class WinUIBackend:
...
+
+        let picker = FolderPicker()
+
+        let window = window ?? windows[0]
+        let hwnd = window.getHWND()!
+        let interface: SwiftIInitializeWithWindow = try! picker.thisPtr.QueryInterface()
+        try! interface.initialize(with: hwnd)
+
+        picker.commitButtonText = fileDialogOptions.defaultButtonLabel
+        picker.fileTypeFilter.append("*")
```
stackotter, line 1596: Is this necessary when selecting folders?

stackotter (changes requested):
I just have one question about the code, but in general this seems like a good fix!

raliclo:
I will do some validation today

stackotter ***@***.***>於 2026年8月8日週六，上午10:48寫道：

Sources/WinUIBackend/WinUIBackend.swift
raliclo: Yes, this is required by FolderPicker. The WinRT/UWP FolderPicker docs say FileTypeFilter must be specified before PickSingleFolderAsync, otherwise it throws; using "*" lists all folders. I kept the line and added a short comment to make that requirement explicit.

PR #683, opened 2026-07-18
WinUIBackend: enforce window size limits

raliclo:
- No dedicated issue : Fix WinUIBackend `setSizeLimits` unimplemented log

### DIFF
--- a/Sources/WinUIBackend/WinUIBackend.swift
+++ b/Sources/WinUIBackend/WinUIBackend.swift
@@ -209,31 +209,7 @@ public final class WinUIBackend:
             self.dispatcherQueue = window.dispatcherQueue
         }
 
-        // import WinSDK
-        // import CWinRT
-        // @_spi(WinRTInternal) import WindowsFoundation
-        // let minSizeHook: HOOKPROC = { (nCode: Int32, wParam: WPARAM, lParam: LPARAM) in
-        //     if nCode >= 0 {
-        //         let ptr = UnsafeRawPointer(bitPattern: Int(lParam))?
-        //             .assumingMemoryBound(to: CWPRETSTRUCT.self)
-        //         if let msgInfo = ptr?.pointee, msgInfo.message == WM_GETMINMAXINFO {
-        //             print("Received WM_GETMINMAXINFO")
-
-        //             // var value: HWND = .init(0)
-        //             _ = try! window._inner.perform(
-        //                 as: __x_ABI_CMicrosoft_CUI_CXaml_CIWindowNative.self
-        //             ) { pThis in
-        //                 try! CHECKED(pThis.pointee.lpVtbl.pointee.get_WindowHandle(pThis, nil))
-        //             }
-        //         }
-        //     }
-        //     return CallNextHookEx(nil, nCode, wParam, lParam)
-        // }
-
-        // _ = SetWindowsHookExW(WH_CALLWNDPROCRET, minSizeHook, nil, GetCurrentThreadId())
-        // print("Registered")
-
-        // print(GetDpiForWindow(nil))
+        window.installSizeLimitHandler()
 
         if let size {
             setSize(ofWindow: window, to: size)
@@ -289,7 +265,8 @@ public final class WinUIBackend:
         minimum minimumSize: SIMD2<Int>,
         maximum maximumSize: SIMD2<Int>?
     ) {
-        debugLogOnce("\(#function) unimplemented")
+        window.minimumContentSize = minimumSize
+        window.maximumContentSize = maximumSize
     }
 
     public func setResizeHandler(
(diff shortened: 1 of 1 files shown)


stackotter (changes requested):
The code looks good, but when I tried it out in my Windows 11 VM I was getting some odd behaviour. It seems to allow the window to become about 20 pixels thinner than it should, and maybe 20-30 pixels shorter than it should? I haven't actually done any measurements because I had to head out before I got the chance

raliclo:
The ⁠ptMinTrackSize⁠ / ⁠ptMaxTrackSize⁠ in ⁠WM_GETMINMAXINFO⁠ expects the tracking size of the entire native window—that is, the outer window size.
The original implementation directly passed the SwiftCrossUI content/client size instead, failing to account for the Windows non-client frame (including the space for borders, captions, and title bars). As a result, the window could shrink approximately 20px narrower and 20–30px shorter than intended, which matches the behavior reported.

I will make a new commit for this soon.

raliclo:
This PR should be working now.

Sources/WinUIBackend/WinUIBackend.swift
```diff
@@ -211,31 +211,7 @@ public final class WinUIBackend:
...
-        //         }
-        //     }
-        //     return CallNextHookEx(nil, nCode, wParam, lParam)
-        // }
-
-        // _ = SetWindowsHookExW(WH_CALLWNDPROCRET, minSizeHook, nil, GetCurrentThreadId())
-        // print("Registered")
-
-        // print(GetDpiForWindow(nil))
+        window.installSizeLimitHandler()
```
stackotter, line 214: I believe that you're calling this twice. Once here, and once in CustomWindow's init. This call should be removed, as the init is the more sensible place to do this setup.

Sources/WinUIBackend/WinUIBackend.swift
```diff
@@ -2284,11 +2265,103 @@ public class CustomWindow: WinUI.Window {
...
+        guard let hwnd = cachedAppWindow.getHWND() else {
+            let scaleFactor = scaleFactor
+            return POINT(
+                x: LONG((Double(contentSize.x) * scaleFactor).rounded(.awayFromZero)),
+                y: LONG(
+                    (Double(contentSize.y + contentHeightAdjustment) * scaleFactor)
+                        .rounded(.awayFromZero)
+                )
+            )
+        }
```
stackotter, line 2334: I don't think there's a point in having this fallback, we should just error out, because this code is unreachable (if we get here we already had a valid HWND, and there's no reason that we should actually fail to get the HWND of a valid window. It's better to just fatal error with a message rather than return a likely incorrect and untested result.

Sources/WinUIBackend/WinUIBackend.swift
```diff
@@ -2284,11 +2265,103 @@ public class CustomWindow: WinUI.Window {
...
+
+        let dpi = GetDpiForWindow(hwnd)
+        let effectiveDPI = dpi == 0 ? UINT(USER_DEFAULT_SCREEN_DPI) : dpi
```
stackotter, line 2337: `dpi` should only be `0` if hwnd is invalid. We trust the hwnd so there's no reason for us to complicate the code with fallbacks (especially given that falling back would only act to hide the underlying error with a value that looks vaguely correct but is actually just arbitrary)

Sources/WinUIBackend/WinUIBackend.swift
```diff
@@ -2284,11 +2265,103 @@ public class CustomWindow: WinUI.Window {
...
+        )
+
+        var rect = RECT(left: 0, top: 0, right: clientWidth, bottom: clientHeight)
+        let style = DWORD(bitPattern: Int32(GetWindowLongPtrW(hwnd, GWL_STYLE)))
+        let extendedStyle = DWORD(bitPattern: Int32(GetWindowLongPtrW(hwnd, GWL_EXSTYLE)))
+
+        guard AdjustWindowRectExForDpi(
+            &rect,
+            style,
+            false,
```
stackotter, line 2351: Add a comment to the end of this line; `// Our windows don't have Win32 menus`. The `false` just looks like a magic value. The comment should help people who end up debugging window sizing issues (which do come up every so often).

Sources/WinUIBackend/WinUIBackend.swift
```diff
@@ -2284,11 +2265,103 @@ public class CustomWindow: WinUI.Window {
...
+            extendedStyle,
+            effectiveDPI
+        ) else {
+            logger.warning("failed to adjust WinUI size limit for window frame")
+            return POINT(x: clientWidth, y: clientHeight)
```
stackotter, line 2356: This should be an error instead of a warning. If this function fails, then we're probably running on an unsupported windows version (before Windows 10 version 1607), and we wouldn't get here in the first place. Returning the client size as a fallback would also just lead to incorrect results, best to error (in this case probably a fatal error because we don't really have any other options, but if you want, you could make this function return an optional, and then update `windowProc` to fallback to a minimum size of 0x0 and a maximum size of nil when it fails to compute the tracking size).

stackotter (changes requested):
Thanks for fixing up the implementation. It does appear to work correctly now (from my brief local testing). I've got some concerns about the code, but they shouldn't be too tricky to address.

PR #685, opened 2026-08-01
Tests: Option 1 — derive AppKitBackend dimensions expectation from backend metrics

### LINKED ISSUE #684: AppKitBackend dimensions test fails on macOS 26+: hardcoded expectation pins NSButton intrinsic metrics
A test case fails on a clean checkout of swift-cross-ui on my system due to a reoccurring issue: OS level metrics for SwiftUI that are being captured in a test case's expectation. [SwiftCrossUITests.swift:194](https://github.com/moreSwift/swift-cross-ui/blob/199a85614e3b2346aa10736b12f969af14a1f1ea/Tests/SwiftCrossUITests/SwiftCrossUITests.swift#L194-L198)

**Failure:** `Expectation failed: result.size == ViewSize(92, 96)` — the view actually lays out to 102×104 on this OS.
**How to Repro:** On macOS 27 call the test suite `./Scripts/test.sh` on a clean checkout of main (199a856). 

**Environment:** macOS 27.0 beta (Build 26A5388g), Xcode 27 Beta 3 (Swift 6.4). 

**Root cause:** NSButton intrinsic content size changed on newer macOS (button content is now hosted via `_NSCoreHostingView<AppKitButton>`): in-graph buttons report 82×24 / 76×24 vs the 72×16 the hardcoded expectation implies. System font metrics are unchanged (13pt / 16pt line height). 

**History:** this expectation has been seen twice before (5e4042d) → 88×95 (9313474, #112) → 92×96 (5e7b274).
CI pins `runs-on: macos-15`, which is why CI stays green.

**Proposed fix:** derive the expectation from the in-graph child widgets' natural sizes plus `defaultPaddingAmount` and `VStack.defaultSpacing`, so the test keeps catching layout-composition regressions (mutation-verified) without pinning AppKit's control metrics.

**LLM Usage:** *I used claude to root cause the issue I observed, to collect system level metadata, to search for past issues/commits. I manually authored the rest of the content. I take responsibility for everything written above. I stand ready to adjust my issue and work with maintainers on a fix*

fbartho:
Fixes #684

The AppKitBackend dimensions test hardcodes `ViewSize(92, 96)`, which pins AppKit's NSButton intrinsic metrics; those changed on macOS 26+/27 (button content now hosted via `_NSCoreHostingView`), so the test fails on newer machines while CI (macos-15) stays green.

This change derives the expectation from the same inputs the layout system consumes — the in-graph child widgets' natural sizes, `backend.defaultPaddingAmount`, and `VStack.defaultSpacing` — via two test helpers. Test-only; no `Sources/` changes.

Notes for review:
-  The assertion still catches padding/spacing/composition regressions - injecting `+3` into `PaddingModifier.horizontalPadding` fails the test
- I considered fuzzy math with a tolerance, but that seemed messier.
- Weakness of this solution: a backend whose `naturalSize` itself goes wrong won't be caught (both sides read the same API).
- In-graph measurement matters: fresh probe buttons report 81 wide where in-graph ones report 82.

### DIFF
--- a/Tests/SwiftCrossUITests/SwiftCrossUITests.swift
+++ b/Tests/SwiftCrossUITests/SwiftCrossUITests.swift
@@ -191,8 +191,20 @@ struct SwiftCrossUITests {
                 environment: environment
             )
 
+            // The intrinsic sizes of AppKit's controls change between macOS releases
+            // (post-Sequoia, NSButton hosts its content in a SwiftUI view and reports a
+            // wider intrinsic size), so hardcoding the expected size makes this test fail
+            // on any OS version other than the one the numbers were recorded on.
+            // Instead, derive the expectation from the same metrics the layout system
+            // consumes. That still exercises the layout algorithm — how padding, stack
+            // spacing and child sizes compose — without pinning AppKit's own metrics.
+            let expectedSize = Self.expectedCounterViewSize(
+                ofViewGraph: viewGraph,
+                backend: backend
+            )
+
             #expect(
-                result.size == ViewSize(92, 96),
+                result.size == expectedSize,
                 "View update result mismatch"
             )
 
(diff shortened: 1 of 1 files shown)


stackotter:
I think this test is probably not testing anything super useful related to AppKitBackend anymore. It'd probably be more suited as a DummyBackend test so that we can run it on every platform. Then we could hardcode an expected result size as well

Issue #687, opened 2026-08-01
Feature Proposal: render swift-cross-ui components in Xcode Previews


fbartho:
It would be great if we could author swift-cross-ui components with `#Preview` SwiftUI-Preview support in Xcode.

**Suggested mechanism** (working — PR incoming): a SwiftUI `NSViewRepresentable` whose coordinator owns a `ViewGraph`, rendered through `AppKitBackend`. Uses only the existing `@_spi(Backends)` surface — zero changes to `SwiftCrossUI` or `AppKitBackend` source.
- A `#Preview` overload handles swift-cross-ui views with no `#if` platform gate — files using it build unchanged on Linux/Windows. 

**One limitation**, Apple-corroborated: SwiftPM has no way to set the build setting Xcode's preview pipeline needs to dynamically load package products, so previews that reach across a package's modules need an app-target scheme as the host (Apple's host-selection model, WWDC23 session 10252; DTS thread [812649](https://developer.apple.com/forums/thread/812649) confirms no manifest-level lever exists).

<details><summary>What we tested and eliminated along the way</summary>

- A differently-spelled macro (`#SCUIPreview`): never shown — the canvas scans source for the literal `#Preview` token
- Delegating our macro to SwiftUI's `#Preview`: detected, but "Missing Preview" — nested macro expansion produces a different registry identity
- Emitting SwiftUI's registration directly (public `DeveloperToolsSupport.PreviewRegistry`): registry identity byte-identical to SwiftUI's — still "Missing Preview"
- Unified-log forensics: the real failure is the preview agent failing to JIT-link package symbols; SwiftUI-only bodies work because system frameworks are already loaded
- The lever is `ENABLE_DEBUG_DYLIB` (a per-product-type default, settable on Xcode targets) — SwiftPM has no API to set it
- App-target test: the ungated overload renders and state updates work
- Also established: `PreviewProvider` gets no canvas tabs on current Xcode; a declaration macro can't nest inside another macro's closure; attached macros can't rewrite the `#Preview` they annotate; a same-module macro overload captures the spelling absolutely

</details>

**Questions**
- Naming: `SCUIPreview` / `SwiftCrossUIPreviews`? (The PR uses these; happy to rename.)
- Is a UIKit variant wanted? It would have to be a parallel implementation.
- I don't think this affects the GTK/Win/Other backend targets in any negative way, though obviously Xcode can only run previews for AppKitBackend or UIKitBackend

[...]

PR #689, opened 2026-08-01
Add Xcode Previews support for SwiftCrossUI views

fbartho:
Implements #687

- `SCUIPreview`: a SwiftUI `NSViewRepresentable` hosting AppKitBackend's rendering of a SwiftCrossUI view in Xcode's Previews canvas. Uses only the existing `@_spi(Backends)` surface — no changes to `SwiftCrossUI` or `AppKitBackend` source.
- A `#Preview` overload, so previewing a SwiftCrossUI view is spelled exactly like previewing a SwiftUI one — no `#if` gates, no SwiftUI import; the file builds unchanged on Linux/Windows. The closure's return type selects between SwiftUI's macro and this one, so one file can preview both kinds of view.
- `Examples/PreviewsExample`: an in-repo app project that hosts previews of package views (Xcode only renders a package view's preview from an app-target scheme). The `.xcodeproj` is generated and gitignored; the committed `project.yml` regenerates it via xcodegen — an optional tool dependency, only there to avoid maintaining a hand-written xcodeproj (see CONTRIBUTING §11):

  ```sh
  xcodegen generate --spec Examples/PreviewsExample/project.yml --project Examples/PreviewsExample
  ```

- `SCUIPreviewSnapshot`: a small testing utility that renders a view to a reproducible 1x PNG.
- Tests: hosting, registration, and macro-expansion suites; `./Scripts/test.sh` at baseline.

Verified working in the Xcode Previews canvas:

previews-example-canvas-2026-08-01

*I used claude to implement this PR. I reviewed every line of work, and take responsibility for any mistakes, and I'm thrilled to help collaborate with maintainers. I iterated on a bunch of different approaches with claude, and then deleted all the noise to and excess to keep what felt like a minimal set*


bbrk24 (changes requested):
CounterExample doesn't compile.

fbartho:
Sorry about that @bbrk24 I'm investigating!

I upgraded to the macOS 27 Beta / Xcode 27 Beta on my daily driver computer, so this could entirely be my fault, or I goofed up when cleaning this up when pushing to a PR.

I'll circle back once I understand what broke.

fbartho:
Found it! Sorry about that @bbrk24, this wasn't anything to do with the beta, I just got overeager after getting it working and forgot that this sample code couldn't work in that target!

I've fixed the build in 380ecc1 and otherwise improved the documentation around the edge-case. Same module previews *are possible* just you have to be in an Xcodeproj-based target so you can set the necessary build-var.

stackotter:
I do believe that previews would provide value to users of SwiftCrossUI, but in their current form I don't think they're a good fit for the project. There are a few things that I would like to see solved/addressed before considering merging this into SwiftCrossUI, and some of the issues may unfortunately be rather difficult to solve;

1. People shouldn't need to have a SwiftUI app entrypoint to gain access to previews (as this would mean that previews can't be used within an actual SwiftCrossUI target).
2. People shouldn't need to have a xcodeproj duplicating their cross platform SwiftPM package manifest in order to access previews. That would cause too much friction for any moderately sized project with multiple targets, products, and dependencies. If a developer made changes to the project's Package.swift on Linux then there'd be no way for them to update the xcodeproj to avoid breaking previews for people working on the same app on macOS. You may be able to tackle this by generating Xcode build schemes in `.swiftpm` (which could be automated by Swift Bundler to make previews a proper first-class feature of the moreSwift ecosystem).

Also, please make sure that you're abiding by our [LLM policy](https://github.com/moreSwift/swift-cross-ui/blob/main/LLM_POLICY.md). Your PR description feels a bit LLM, but I'm open to being incorrect about that.

fbartho:
>@stackotter: People shouldn't need to have a SwiftUI app entrypoint to gain access to previews

I agree with this point, however as-I-understand-it, this is a limitation of Xcode. Xcode discovers #Preview either way, but it only successfully runs previews when the hosting target is built with `ENABLE_DEBUG_DYLIB` — without it, the preview host fails to link the package's symbols. And unfortunately we cannot set that flag on a pure Package.swift-based library target ([I tested a bunch of approaches](https://github.com/moreSwift/swift-cross-ui/issues/687)), and Apple confirms there is no way to set any Xcode build setting from a package manifest ([DTS thread](https://developer.apple.com/forums/thread/812649)).

A cross-platform target *can* however include the #Preview macro (as implemented by SwiftCrossUI, we're tricking Xcode's detection, and then properly implementing a shell), it just will do nothing on unsupported targets.

**Regarding your .swiftpm/ recommendation, could you, please, elaborate more?**

I considered some 3rd party xcodeproj generators as a way to accomplish this special target, but they felt like a step too far to propose in this project. If that's what you prefer, could you elaborate on the .swiftpm / automation approach that you'd like me to consider? Happy to take a stab at it.

<details><summary>Before replying I also tested a .swiftpm scheme directly. Details hidden (click here)</summary>

1. `.swiftpm/xcode/xcshareddata/xcschemes/` -- schemes defined here do build. 
2. Unfortunately, nothing I tried to configure there reaches build settings.
 - ❌ `<BuildSettings>` element is silently ignored
 - ❌ launch environment variables don't help because they're runtime only
 - ❌ a custom `buildConfiguration` name is accepted, but I couldn't figure out how a plain package could leverage that to help in this case.

</details>
 
In summary: xcodeproj/xcconfig files seem to be the blessed way to set these settings, and I couldn't figure out a different way. I'm open to any recommendations that this project would approve of!

Re: the LLM-feedback. I realized my misunderstanding when I got feedback on a different PR, and commit to manually authoring the full PR description going forwards -- I did not however reauthor this PR's description after updating my understanding. _And it goes without saying, this comment was manually authored, including the html-details-disclosure tag._

PR #690, opened 2026-08-01
Port dimensions test to DummyBackend with hardcoded expectation

### LINKED ISSUE #684: AppKitBackend dimensions test fails on macOS 26+: hardcoded expectation pins NSButton intrinsic metrics

fbartho:
Fixes #684. Alternative to #685 and #686, prompted by [stackotter's 2026-08-01
comment on #685](https://github.com/moreSwift/swift-cross-ui/pull/685#issuecomment-5149681461).

## What changed

1. `DummyBackend: Give buttons a natural size` (095dee5) — DummyBackend's text-label button never overrode `Widget.naturalSize`, so it inherited `.zero` and rendered zero-sized. Adds a `naturalSize` override that estimates from the label using the same character metrics `size(of:)` uses, and stores the resolved font so the estimate has a font to work from. After merging #590 this override lives on `SimpleButton`; its consumer is Menu's `naturalSize(of:)` sizing path, since core now sizes `Button` from its label plus `buttonPadding`.

2. `Tests: Port basic-dimensions test from AppKitBackend to DummyBackend` (ffa79cf) — replaces `testBasicLayout`'s `AppKitBackend` instance with `DummyBackend`, removes the `#if canImport(AppKitBackend)` gate around the test (it now runs on every platform), and hardcodes the expected size (84, 88) with an inline comment deriving it from DummyBackend's own constants (`size(of:)` character metrics, `buttonPadding`, `VStack.defaultSpacing`, `DummyBackend.defaultPaddingAmount`) rather than from any native control's intrinsic metrics.

3. `DummyBackend: Centralise text sizing in a single helper` (55cc6f0) — addresses the review comment: `size(of:)` and `SimpleButton.naturalSize` both rely on one private `textSize` helper.

The branch also merges `upstream/main` (ac0bc73) to resolve the conflict with #590; the merge commit message documents the expectation re-derivation, since #590 changed how core sizes buttons.

## Other 

- The hardcoded (84, 88) expectation was derived by hand from the backend's documented constants and cross-checked against the test's actual passing run — the arithmetic is in the test's inline comment, not just a captured value.

### DIFF
--- a/Sources/DummyBackend/DummyBackend.swift
+++ b/Sources/DummyBackend/DummyBackend.swift
@@ -78,6 +78,21 @@ public final class DummyBackend:
         public var label = ""
         public var font: Font.Resolved?
         public var action: (() -> Void)?
+
+        /// Buttons get their intrinsic size from the backend (see the TODO in
+        /// `Button.computeLayout`), so a backend that leaves this at zero renders
+        /// zero-sized buttons. Estimate from the label the same way `size(of:)` does.
+        override public var naturalSize: SIMD2<Int> {
+            guard let resolved = font else { return .zero }
+            let characterHeight = Int(resolved.pointSize)
+            let characterWidth = characterHeight * 2 / 3
+            let horizontalPadding = 10
+            let verticalPadding = 5
+            return SIMD2(
+                characterWidth * label.count + horizontalPadding * 2,
+                Int(resolved.lineHeight) + verticalPadding * 2
+            )
+        }
     }
 
     public class ToggleButton: Widget {
@@ -614,6 +629,7 @@ public final class DummyBackend:
         let button = button as! Button
         button.label = label
         button.action = action
+        button.font = environment.resolvedFont
     }
 
     public func createToggle() -> Widget {
--- a/Sources/DummyBackend/DummyBackend.swift
+++ b/Sources/DummyBackend/DummyBackend.swift
@@ -79,18 +79,21 @@ public final class DummyBackend:
         public var font: Font.Resolved?
         public var action: (() -> Void)?
 
-        /// Buttons get their intrinsic size from the backend (see the TODO in
-        /// `Button.computeLayout`), so a backend that leaves this at zero renders
-        /// zero-sized buttons. Estimate from the label the same way `size(of:)` does.
+        /// Menu sizes its button widget through `naturalSize(of:)`, so leaving
+        /// this at zero renders zero-sized menu buttons.
         override public var naturalSize: SIMD2<Int> {
-            guard let resolved = font else { return .zero }
-            let characterHeight = Int(resolved.pointSize)
-            let characterWidth = characterHeight * 2 / 3
+            guard let font else { return .zero }
+            let labelSize = DummyBackend.textSize(
+                of: label,
+                displayedWith: font,
+                proposedWidth: nil,
+                proposedHeight: nil
+            )
             let horizontalPadding = 10
             let verticalPadding = 5
             return SIMD2(
-                characterWidth * label.count + horizontalPadding * 2,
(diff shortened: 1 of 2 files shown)


Sources/DummyBackend/DummyBackend.swift
```diff
@@ -78,6 +78,21 @@ public final class DummyBackend:
         public var label = ""
         public var font: Font.Resolved?
         public var action: (() -> Void)?
+
+        /// Buttons get their intrinsic size from the backend (see the TODO in
+        /// `Button.computeLayout`), so a backend that leaves this at zero renders
+        /// zero-sized buttons. Estimate from the label the same way `size(of:)` does.
+        override public var naturalSize: SIMD2<Int> {
```
stackotter, line 84: This duplicates `size(of:...)`'s logic. It'd be best to keep that logic centralised so that we have a single source of truth for text sizing. You can either call `size(of:...)` directly here with a dummy `Widget` (the widget doesn't get used by DummyBackend), or if you want a nicer solution, you could extract the core implementation of `size(of:...)` into a private method that only takes parameters that it strictly needs, and then both `size(of:...)` and this `naturalSize` implementation can rely on that single private method for text sizing.

stackotter (changes requested):

stackotter:
In future please don't open one PR per potential approach. I'd prefer if we could discuss potential approaches in a GitHub issue thread or PR review thread and then create a new PR/replace the existing PR once we've decided on a potential solution.

Sources/DummyBackend/DummyBackend.swift
fbartho: Addressed in 55cc6f0. Let me know if this wasn't quite what you meant!

fbartho:
Re: multiple PRs for the same topic, before discussion in an issue by a maintainer: I understand, and commit to not doing so again!

Thanks for the feedback, and sorry my exuberance was noisy @stackotter.

stackotter (approved):

PR #696, opened 2026-08-05
Exclude WinUI targets from the package graph when built by Xcode

fbartho:
Building the package under Xcode or xcodebuild on macOS dies during the build phase because WinUIInterop unconditionally includes ShObjIdl.h, a Windows-only header, and the WinUI product, targets, and swift-winui dependency are unconditional in Package.swift. This gates them behind a `winuiBackendSupported` check, keyed by `invokedByXcode`, the same shape as the existing Android block.

Options: 
a. accept this non-Windows resolution
b. drop the lockfile hunk and let each platform resolve locally
c. or prefer a different mechanism? *I need advice here*

**Verified on macOS:** on the plain SwiftPM path, `swift package dump-package` shows the full target list with both WinUI targets present (matching upstream). Under `xcodebuild`, the scheme list contains no WinUIBackend, and a full build log contains no mention of WinUIBackend, WinUIInterop, or ShObjIdl — the ShObjIdl.h failure does not occur. A fully green xcodebuild run was blocked only by missing gtk3 headers on my machine, unrelated to this change.

**Not verified:** I don't have a Windows box set up for development at this time, so I didn't try it on Windows.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -47,14 +47,10 @@ let androidBackendSupported: Bool
 
 // xcodebuild can't handle non-Apple platform conditional dependencies for some weird
 // reason, so we have to remove WinUIBackend/WinUIInterop and the swift-winui
-// dependency when we detect that we're being built by xcodebuild, mirroring
-// androidBackendSupported above.
-let winuiBackendSupported: Bool
-#if compiler(>=6.2)
-    winuiBackendSupported = !invokedByXcode
-#else
-    winuiBackendSupported = false
-#endif
+// dependency when we detect that we're being built by xcodebuild. Unlike
+// AndroidBackend there is no compiler-version floor: WinUIBackend builds with
+// every Swift version this package supports (Windows CI uses Swift 6.1).
+let winuiBackendSupported = !invokedByXcode
 
 var defaultBackendDependencies: [Target.Dependency]
 if let backend = env["SCUI_DEFAULT_BACKEND"] {
@@ -373,7 +369,7 @@ let package = Package(
     )
 #endif
 
-// Add WinUIBackend if the Swift version is new enough and we're not using xcodebuild
+// Add WinUIBackend unless we're being built by xcodebuild
 if winuiBackendSupported {
     package.dependencies += [
         .package(
--- a/Package.swift
+++ b/Package.swift
@@ -45,6 +45,16 @@ let androidBackendSupported: Bool
     androidBackendSupported = false
 #endif
 
+// swift-winui only resolves on Windows, so WinUIBackend/WinUIInterop must be excluded
+// from the package graph entirely on other hosts (resolution fails otherwise, since
+// SwiftPM resolves dependencies for every target regardless of platform conditions).
+let winuiBackendSupported: Bool
+#if os(Windows)
+    winuiBackendSupported = true
+#else
+    winuiBackendSupported = false
+#endif
+
 var defaultBackendDependencies: [Target.Dependency]
 if let backend = env["SCUI_DEFAULT_BACKEND"] {
     defaultBackendDependencies = [.target(name: backend)]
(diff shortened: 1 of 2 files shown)


bbrk24:
Don't we already have `Scripts/test.sh` to run the tests without having to build other platforms' backends?

fbartho:
Sorry, to clarify, are you saying that Package.swift should have worked unmodified? 

What I experienced was when opening the project in Xcode, and told it to build the swift-cross-ui-package, I immediately get "'ShObjIdl.h' file not found". (after of course installing GTK)

Screenshot of compilation errors

`Scripts/test.sh` is perhaps good for replacing direct `swift test` invocations, but I assumed I should be able to build the package target since it said for "My Mac"

Since Android already had an exclusion for android targets, I figured that was the standard pattern here.

Thoughts @bbrk24 ?

bbrk24:
Sorry, I thought we were talking about Linux. Yes, there are some known issues with building/running the tests with Xcode. The exclusion for Android came from the fact that an iOS-only project that _imported_ SwiftCrossUI was unable to compile at all, because `xcodebuild` tried to compile Android dependencies for some reason.

fbartho:
Right, so you agree @bbrk24 that this code is useful? I had to disable this for working with swift-cross-ui on my system, I assumed we would want the same kind of fix

bbrk24:
No; I never try to run the tests in Xcode. I always have a terminal window open anyways (e.g. for git commands) so I just run the test script there.

stackotter (changes requested):
Rather than gate by host platform, I think it'd be best to only remove the Windows related targets/products/dependencies when building under Xcode or xcodebuild, as we do for Android. You should be able to use the `invokedByXcode` variable to conditionalize our Windows related manifest sections. This approach should minimise the Package.resolved thrashing that you mentioned (i.e. only Xcode users would cause the large Package.resolved changes, rather than every contributor).

I like to be able to have all targets/products/dependencies available on macOS so that I can run swift package commands such as `swift package resolve` or `swift package update` and see if all of our Windows related dependencies are being included correctly.

fbartho:
Thanks for the feedback @stackotter, working on this!

fbartho:
Okay, I believe I've now properly implemented this to be symmetric with the Android pattern.

Please let me know if you need anything more @stackotter, I'm happy to take another pass in a day or two.

Package.swift
```diff
@@ -45,6 +45,17 @@ let androidBackendSupported: Bool
...
 
+// xcodebuild can't handle non-Apple platform conditional dependencies for some weird
+// reason, so we have to remove WinUIBackend/WinUIInterop and the swift-winui
+// dependency when we detect that we're being built by xcodebuild, mirroring
+// androidBackendSupported above.
+let winuiBackendSupported: Bool
+#if compiler(>=6.2)
+    winuiBackendSupported = !invokedByXcode
+#else
+    winuiBackendSupported = false
```
bbrk24: WinUIBackend is very much supported in Swift 6.1, at least for now. If we want to not support it, we'd also have to edit WinUIBackend itself, e.g:

https://github.com/moreSwift/swift-cross-ui/blob/f8bdf05729ae1cd1b7f3ec4e57177a6965052146/Sources/WinUIBackend/WinUIBackend.swift#L2532-L2534

I have Swift 6.1 on my own Windows machine.

fbartho: @bbrk24 I don’t think I understand your comment. Are you proposing a specific change to this PR?

I was not trying to change the support-level of WinUI. After my PR it should still be “as supported” as it was before my PR.

I was trying to make the build work on a machine that doesn’t have access to the Windows-only headers.

Do you have a different change you’d propose instead?

fbartho: Oh! I get it now: I misread your comment. I didn't realize the emphasis was on the swift version number, and that I was cutting support for Swift 6.1.

My only excuse is that I must have been tired, and I interpreted feedback as "make it exactly like the Android one".

Looking at this PR with fresh eyes today, and digging into the CI failure, I realize now what you were saying.

I've pushed a new commit, and we'll see if CI passes this one.

Sorry about that, and thanks for your review @bbrk24

PR #697, opened 2026-08-06
GtkBackend: opt-in bottom anchoring for ScrollView

epsilver:
## Problem

Chat/log-style views have no way to stay pinned to the newest content: the vertical adjustment's `value` stays at an absolute offset while `upper` grows, so a streaming transcript visibly jumps as entries arrive. There is no scroll API in the framework at all today (the ScrollView docs acknowledge it).

## What this adds

- `ScrolledWindow.anchorsToBottom` (GTK backend), off by default.
- A `scrollAnchorsToBottom` environment value so individual scroll views opt in — the rail/lists keep their normal behavior.

On each adjustment `changed` signal, the anchor snaps `value = upper - pageSize`, **but only when the pre-growth geometry was within 120pt of the bottom**:

| Situation | Behavior |
|---|---|
| Content opens taller than the viewport | Lands on the newest entries |
| Growth while at bottom | Follows every delta |
| Growth while scrolled up | Stays put — no fighting the user |
| Back within 120pt of the bottom | Re-engages on the next growth |

The geometry is read **before** updating `lastUpper`: GTK coalesces multiple growths into one `changed` signal, so writing first would make a scrolled-up user look at-bottom and get yanked down (caught by the included reasoning in tests downstream).

## Notes

- 120pt slack reads well at default body sizes; easy to expose later if wanted.
- No behavior change for existing users — the anchor installs only when the environment flag is set.

Dogfooded in a transcript-heavy GTK4 app (streaming sessions over a websocket): pinned open, follow-stream, and release-without-drift all verified.

### DIFF
--- a/Sources/Gtk/Widgets/ScrolledWindow.swift
+++ b/Sources/Gtk/Widgets/ScrolledWindow.swift
@@ -3,6 +3,55 @@ import CGtk
 public class ScrolledWindow: Widget {
     var child: Widget?
 
+    /// Stick-to-bottom anchoring for streaming content. While enabled, the
+    /// view stays pinned to the newest content as it grows when the scrollbar
+    /// was within range of the bottom before the growth; scrolling up
+    /// releases the pin, and scrolling back near the bottom re-engages it on
+    /// the next growth. Off by default; opt in per scroll view via the
+    /// `scrollAnchorsToBottom` environment value.
+    public var anchorsToBottom = false {
+        didSet { installBottomAnchor() }
+    }
+    /// The adjustment's `upper` as of the last change; compared against the
+    /// pre-change scroll geometry to decide whether the user was at the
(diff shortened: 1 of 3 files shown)


bbrk24:
I would rather see an equivalent to [`defaultScrollAnchor(_:)`](https://developer.apple.com/documentation/swiftui/view/defaultscrollanchor(_:)) than something GTK-specific like this. We've discussed it on the Discord before but it looks like we never actually created an issue for it.

Now, you could make the skeleton of `defaultScrollAnchor` and just have it only work with GtkBackend, and then we can fill in the other backends later. But I don't like having an ad-hoc boolean environment value that only one backend uses.

verticalrectangle:
That's a good point and a much more elegant solution. I'll close this one out for now.

PR #698, opened 2026-08-06
GtkBackend: memoize text measurement and skip redundant CSS re-parses

epsilver:
## Problem

On text-dense views (hundreds of `Text` nodes), the Gtk backend spends most of every layout pass on two repeated costs:

1. **`size(of:whenDisplayedIn:...)`** builds a fresh Pango context — fontconfig work — for *every* Text view on *every* pass.
2. Every `update*` for a styled widget runs `css.clear(); css.set(...)` unconditionally, forcing two `gtk_css_provider_load_from_data` parses per widget per pass. Each parse invalidates the widget's style and re-signals, cascading more layout work.

Together they pin the main thread at ~100% CPU on dense views while the user is just looking at a screen.

## Changes

- **Memoize text measurement** in a widget-keyed cache (`TextSizeCacheKey`: widget, text, proposed width/height, ellipsizing, font, line limit). Measurements are deterministic for identical inputs. The table is capped at 2048 entries so streaming content can't grow it unboundedly.
- **Skip redundant CSS writes**: build the candidate CSS first and only assign it when it differs, in `updateTextView`, `updateButton`, `updateToggle`, `updateTextField`, `updateTextEditor`, and `updateCalendarWidget`.

No behavior change beyond performance: same measurements, same styles, same widgets.

## Evidence

Dogfooded on a transcript view with ~200 text nodes (streaming session view in a GTK4 app): before the change, layout passes drove the main thread to 100% CPU during streaming; after, the view idles between updates.

### DIFF
--- a/Sources/GtkBackend/GtkBackend.swift
+++ b/Sources/GtkBackend/GtkBackend.swift
@@ -814,9 +814,34 @@ public final class GtkBackend:
             }
 
         textView.selectable = environment.isTextSelectionEnabled
-        textView.css.clear()
-        textView.css.set(properties: Self.cssProperties(for: environment))
+        // Apply CSS only on a real change: clear()+set() forces two
+        // gtk_css_provider_load_from_data parses per update, and every parse
+        // invalidates the widget's style — on dense views (hundreds of text
+        // nodes) the style/layout signal cascade pins the main thread.
+        var newCss = textView.css
+        newCss.clear()
+        newCss.set(properties: Self.cssProperties(for: environment))
+        if newCss != textView.css {
+            textView.css = newCss
+        }
     }
+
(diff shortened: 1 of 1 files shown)


mba6comma1:
If you haven’t already, you might want to review our [LLM policy](https://github.com/moreSwift/swift-cross-ui/blob/main/LLM_POLICY.md). In particular, we request that contributors disclose any use of LLMs to produce code, and that PR descriptions be written without using LLMs for anything other than machine translation.

epsilver:
Apologies for the lack of disclosure. This code was produced as part of an automated task run in Omperator (an agent-driven coding workspace), which includes LLM-generated code and description. Closing this PR out of respect for the policy. Thank you for pointing it out.

Issue #595, opened 2026-05-27
Texts inside scrollview getting unnecessarily cut off?


bbrk24:
<img width="877" height="135" alt="Image" src="https://github.com/user-attachments/assets/2444f725-97c6-4474-93bb-97c858fa587f" />

`.fixedSize()` fixes this. Was introduced between 0.3.0 and 0.4.0, probably in the layout priority PR.

PR #700, opened 2026-08-06
GtkBackend: fix ellipsized wrapped labels clipping leading text

epsilver:
## Problem

Every `Text` label in the Gtk backend is created with `ellipsize = .end`. On a *wrapped* label, ellipsizing makes Pango lay the text out wider than the allocation (the ellipsized layout does not honor the wrap width), and `GtkLabel`'s default `xalign` of 0.5 then draws that oversized layout **centered** — long paragraphs lose a slice of text off their *leading* edge inside scroll views. On text-dense, transcript-style views this shows up as the first few characters of long lines being cut off.

The `CustomLabel` layout-height override (added to make multi-line ellipsizing work, since a plain ellipsized GtkLabel only shows one line) makes it worse when applied to plain wrapping labels: it leaves Pango holding a stale line layout wider than the allocation.

## Changes

- **Default `ellipsize` to `.none`** in `createTextView`, and drive the mode from `environment.lineLimitSettings` in `updateTextView` — single-line headlines etc. keep ellipsizing; body text wraps.
- **`xalign = 0.0`** so even if a layout ever ends up wider than its allocation, it can only overflow the trailing edge, never clip the leading one.
- **Gate the layout-height override to ellipsized labels only** — it has no effect to give plain wrapping text.
- Skip no-op size requests in `CustomLabel.setSizeRequest` while in the area (each one queues a resize).

## Evidence

Pixel-measured in a GTK4 transcript view: a 2000px-long paragraph rendered 2194px wide in a 946px allocation, drawn centered, losing ~15 characters off the left. After the change the same paragraphs wrap at the allocation width and start flush at the content edge, in both single- and multi-line cases; `lineLimit(1)` headlines still ellipsize correctly.

### DIFF
--- a/Sources/GtkBackend/GtkBackend.swift
+++ b/Sources/GtkBackend/GtkBackend.swift
@@ -791,7 +791,15 @@ public final class GtkBackend:
         textView.horizontalAlignment = .start
         textView.wrap = true
         textView.lineWrapMode = .wordCharacter
-        textView.ellipsize = .end
+        // Do not ellipsize by default: ellipsizing a wrapped label makes
+        // Pango lay the text out wider than the allocation, and GTK then
+        // draws the oversized layout centered (default xalign 0.5), clipping
+        // the leading edge of long paragraphs. Ellipsize only when a line
+        // limit is set (see updateTextView).
+        textView.ellipsize = .none
+        // Draw the text block from the label's left edge so an oversized
+        // layout can never clip text off the leading edge.
+        textView.xalign = 0.0
         textView.yalign = 0.0
         return textView
     }
@@ -814,6 +822,9 @@ public final class GtkBackend:
             }
 
         textView.selectable = environment.isTextSelectionEnabled
+        // Ellipsize only for line-limited text (single-line headlines etc.);
+        // body text must wrap, never ellipsize-clip.
+        textView.ellipsize = environment.lineLimitSettings != nil ? .end : .none
         textView.css.clear()
         textView.css.set(properties: Self.cssProperties(for: environment))
     }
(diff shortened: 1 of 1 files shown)


bbrk24:
Curiously, this seems to make #595 worse, and I honestly don't understand how.

epsilver:
Good catch, and I think I can explain the mechanism — this made the interaction click for me too.

**What changes for #595's case.** Those cells have no `lineLimit`. Before this PR, a label whose allocation is narrower than its text hits the `ellipsize = .end` default and truncates to `7…`. With this PR, that same label gets `ellipsize = .none` and **wraps at the allocation** instead — and in a one-line-high strip like that weather row, the wrapped second line is vertically clipped, which reads strictly worse than the truncation.

**The shared root.** I'd argue #595 and the clip this PR fixes are the same allocation-vs-content mismatch presenting through the ellipsize knob in two directions:

- **#595:** allocation *narrower* than the text → ellipsize truncates to `7…` (pre-PR) or the wrap clips vertically (post-PR).
- **This PR's case:** the *layout* ends up wider than the allocation (ellipsized wrapped Pango layout + the CustomLabel height override leaving it stale) → GtkLabel draws it centered → leading characters cut.

The ellipsize mode doesn't *cause* either one — it decides how the misallocation presents. `fixedSize()` fixing #595 points the same way: the cell isn't being allocated its natural width in the first place, which smells like the 0.4.0 layout-priority change you mentioned, independent of ellipsizing.

**On which default is "right".** Worth noting SwiftUI wraps unbounded `Text` at the available width — it doesn't truncate without a `lineLimit`. So post-PR behavior is actually *closer* to SwiftUI semantics for #595's cells (wrap), and the reason it looks worse is that the strip is fixed-height: the wrapped content has nowhere to go. That said, I don't want this PR to make an existing bug more visible in a bad way.

**Proposals, in my order of preference:**

1. Keep #700 as-is (line-limit-driven ellipsize = SwiftUI-consistent), and I dig into #595's allocation squeeze as a separate root-cause fix — that's the change that makes the weather strip correct under *both* ellipsize modes. Happy to take it if you can point me at the view code that produced that row.
2. If you'd rather not change the default for unbounded text, I can rescope #700 to keep `ellipsize = .end` as the default and only take the two non-behavioral pieces (`xalign = 0` so oversized layouts clip trailing, never leading; height override gated to line-limited labels). The tradeoff: plain wrapping body text keeps the pango-ignores-wrap pathology — it fixes the *leading* clip but long paragraphs still overflow trailing. That's strictly better than today but doesn't resolve the multiline case properly.

My vote is 1. What do you think?

PR #701, opened 2026-08-06
GtkBackend: stop scroll views propagating content natural width

epsilver:
## Problem

Wrapped text reports its *unwrapped* one-line width as its natural width. With `propagate-natural-width` on (the `GtkScrolledWindow` default), that natural width bubbles up through every ancestor's natural size, and any container that honors naturals — split panes, fixed-position containers — ends up allocating the scroll view **wider than the window**. The overflow then clips content off the window's trailing edge: long lines render past the visible area instead of wrapping at the viewport width.

In SwiftCrossUI this bites any app with long text inside a ScrollView beside other content (rail + detail layouts): the detail pane can end up laid out wider than the window itself.

## Change

Disable natural-width propagation in `createScrollContainer`. The layout system already decides scroll view sizes itself — propagating the content's natural width can only inflate them.

## Evidence

Measured on a rail+detail GTK4 workspace: the detail pane's natural width reached ~2700px from unwrapped label widths and the split paned allocated it past the window edge, right-clipping transcript lines. With propagation off, content wraps at the viewport width and nothing crosses the window edge (verified pixel-precisely).

### DIFF
--- a/Sources/GtkBackend/GtkBackend.swift
+++ b/Sources/GtkBackend/GtkBackend.swift
@@ -676,6 +676,13 @@ public final class GtkBackend:
 
     public func createScrollContainer(for child: Widget) -> Widget {
         let scrollView = ScrolledWindow()
+        // Never propagate the content's natural width. Wrapped text reports
+        // its unwrapped one-line width as its natural width, which inflates
+        // every ancestor's natural width until a container that honors
+        // naturals allocates the scroll view wider than the window — and the
+        // overflow clips content off the window's trailing edge. The scroll
+        // view's own size comes from the layout, not from its content.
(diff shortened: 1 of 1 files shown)


fbartho:
I'm no expert at this, but I have tripped across this problem in other UI frameworks.

Are you sure the right answer is to disconnect the "natural size" signal across the ScrollView boundary? Isn't it still correctly a natural width?

Instead, wouldn't a better choice be to restrict the max-width of the scrollview?

This is a key factor in allowing scrollviews to automatically disable their scrollbars if the content is small enough?

Alternatively, if you constrain a Text-label's max-width, then I would hope it automatically wraps the text and tells you its natural size is the max-width and the height of the number of wrapped lines?

PR #703, opened 2026-08-08
Implement Environment.locale correctly on Android & Windows

bbrk24:
`Locale.current` is hardcoded as `en_001` on non-Apple platforms. This PR changes the root environment handlers in AndroidBackend and WinUIBackend to set `Environment.locale` based on their respective platforms' i18n APIs.

One idea that occurred to me is that `Locale.windowsCurrent` should go in a separate module, which both WinUIBackend and GtkBackend import (with the latter guarded by `#if os(Windows)`). I haven't done that here, but it is an option.

### DIFF
--- a/Sources/AndroidBackend/AndroidBackend.swift
+++ b/Sources/AndroidBackend/AndroidBackend.swift
@@ -305,9 +305,14 @@ public final class AndroidBackend: BaseAppBackend {
            let timeZone = Foundation.TimeZone(identifier: identifier)
         {
             environment.timeZone = timeZone
-            environment.calendar = getCurrentCalendar(timeZone: timeZone)
+
+            let (calendar, locale) = getCurrentCalendarAndLocale(timeZone: timeZone)
+            environment.calendar = calendar
+            environment.locale = locale
         } else {
-            environment.calendar = getCurrentCalendar(timeZone: nil)
+            let (calendar, locale) = getCurrentCalendarAndLocale(timeZone: nil)
+            environment.calendar = calendar
+            environment.locale = locale
         }
 
         environment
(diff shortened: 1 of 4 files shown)


bbrk24:
CI seems to be failing due to an unrelated issue with menu buttons.

Kyle-Ye:
There is a upstream issue and PR tracking this. Would you help go upstream this? 

https://github.com/swiftlang/swift-foundation/pull/2057

bbrk24:
Marking this as draft until I fix the calendar on Windows. (Currently it's always Gregorian.)

bbrk24:
@Kyle-Ye Please, go ahead and copy the Windows code into your PR. I don't mind. (The `@_spi(Backends) import SwiftCrossUI` is only used for the logger.) As discussed on Discord, the Android code I have wouldn't work in Foundation, as it uses Java APIs.

stackotter:
I believe that the CI should begin working again if you merge main into this PR

Sources/AndroidBackend/AndroidBackend.swift
```diff
@@ -305,9 +305,14 @@ public final class AndroidBackend: BaseAppBackend {
...
-            environment.calendar = getCurrentCalendar(timeZone: timeZone)
+
+            let (calendar, locale) = getCurrentCalendarAndLocale(timeZone: timeZone)
+            environment.calendar = calendar
+            environment.locale = locale
         } else {
-            environment.calendar = getCurrentCalendar(timeZone: nil)
+            let (calendar, locale) = getCurrentCalendarAndLocale(timeZone: nil)
+            environment.calendar = calendar
+            environment.locale = locale
```
stackotter: If you store the fetched timezone in a local variable before this if statement, then you could extract this repeated locale+calendar logic out of the if statement to deduplicate it

stackotter (changes requested):
Generally looks great, just requested one small code change

stackotter (approved):
Thanks for making those changes! I'll merge once CI finishes

PR #704, opened 2026-08-08
Add versioned multi-module documentation

Kyle-Ye:
## Summary

- add a VersionedDocC configuration for `main`, `0.8.0`, and `0.7.0`
- reuse the existing platform jobs to produce AppKit, UIKit, Gtk, Gtk3, and WinUI symbol-graph artifacts
- assemble those artifacts with SwiftCrossUI through VersionedDocC 0.0.15
- publish a multi-version site with version-correct source links and adjacent-version API Changes

## Why

The existing DocC pipeline already has the correct platform-specific build setup, but the final merged archive only represents one unversioned deployment. VersionedDocC can now consume the symbol graphs emitted by those existing jobs, so the platform builds do not need to move into a separate documentation pipeline.

## Implementation

The platform jobs continue compiling their own backends and upload symbol graphs as before. `assemble-docs` downloads those artifacts, builds the main SwiftCrossUI target, converts each backend independently, and merges the archives into the existing Pages artifact.

The backend modules are currently enabled for `main`; the tagged `0.8.0` and `0.7.0` snapshots retain the SwiftCrossUI module and participate in the same version selector and API Changes history.

## Validation

- validated `.vdc.json` with VersionedDocC 0.0.15
- JSON and workflow YAML parsing pass
- local merged preview verified with SwiftCrossUI, AppKitBackend, UIKitBackend, GtkBackend, and Gtk3Backend
- the previous fork CI run successfully produced every platform artifact, including WinUI; its final assembly started only after the temporary VDC test tag had been removed, which this update replaces with the released `0.0.15` tag

Preview: https://docclab.github.io/swift-cross-ui/main/documentation/swiftcrossui/
VersionedDocC: https://github.com/DocCLab/VersionedDocC


.github/workflows/build-test-and-docs.yml
```diff
@@ -512,72 +512,46 @@ jobs:
...
+
+    - name: Restore VersionedDocC cache
+      uses: actions/cache@v5
       with:
-        target: Gtk3Backend
-        use-symbol-graph-artifact: true
+        path: .docs/cache/versioned-docc
+        key: versioned-docc-multi-module-${{ runner.os }}-${{ hashFiles('.vdc.json', 'Package.resolved') }}-${{ github.sha }}
+        restore-keys: |
+          versioned-docc-multi-module-${{ runner.os }}-${{ hashFiles('.vdc.json', 'Package.resolved') }}-
```
stackotter, line 801: did you mean to end this with `${{ github.sha }}`? It seems cut short

.github/workflows/build-test-and-docs.yml
```diff
@@ -512,72 +512,46 @@ jobs:
...
-          WinUIBackend.doccarchive \
-          --output-path DocsBuild
-
-    - name: Assemble GitHub Pages artifact
-      run: |
-        set -eux
-
-        cp Pages/google5d79e0866e2627dd.html DocsBuild/google5d79e0866e2627dd.html
-        cp Pages/CNAME DocsBuild/CNAME
-        cp Pages/index.html DocsBuild/index.html
```
stackotter, line 590: Would you be able to make sure that these hardcoded files are included in the documentation artifact? It seems that they're not atm.

stackotter (changes requested):

stackotter:
It appears that VDC isn't generating a multi-target index page, and instead returns a 404 if you try to visit the URL where it would usually be hosted. Here's [SCUI's current multi-target index page](https://docs.swiftcrossui.dev/documentation/). Is that something that VDC's known to not support?

.github/workflows/build-test-and-docs.yml
Kyle-Ye: Yep, this is intentional. `github.sha` gives each commit an immutable cache key, while restore-keys restores the latest compatible cache from the prefix.

.github/workflows/build-test-and-docs.yml
Kyle-Ye: Good catch. I’ll add the CNAME and Google verification files back to the uploaded artifact. VDC generates its own version-aware root index, so I’ll keep that instead of the old redirect to /documentation.

Kyle-Ye:
> It appears that VDC isn't generating a multi-target index page, and instead returns a 404 if you try to visit the URL where it would usually be hosted. Here's [SCUI's current multi-target index page](https://docs.swiftcrossui.dev/documentation/). Is that something that VDC's known to not support?

We do support multi-target. I'll fix the missing index.html issue in VDC next release. And update the PR and preview page.

Kyle-Ye:
> It appears that VDC isn't generating a multi-target index page, and instead returns a 404 if you try to visit the URL where it would usually be hosted.

Fixed.

PR #705, opened 2026-08-08
[AndroidBackend] Fix menu buttons

bbrk24:
This was somehow missed in #590.


kaascevich (approved):
Looks good to me! (Not that that really requires saying.)

stackotter (approved):
Thanks for that fix

PR #708, opened 2026-08-15
GTK: preserve nil list selection (fixes #476)

### LINKED ISSUE #476: The List control starts with the first item already selected on the GTK Backend
When displaying a List using the GTK Backend, if the array used to populate the list already has data, the first element is selected automatically. 

Image

Here is a test case:

```swift
import SwiftCrossUI
import Foundation

struct TestView: View {

    struct Item: Identifiable {
        var id: String = UUID().uuidString
        var title: String = ""
    }

    @State var items: [Item] = [
        Item(title: "Item 1"),
        Item(title: "Item 2"),
        Item(title: "Item 3"),
        Item(title: "Item 4"),
    ]

    @State var selected: String?
    @State var selectionDisabled: Bool = false
    
    var body: some View {
        VStack {            
            List(items, id: \.id, selection: $selected) { item in
                HStack {                
                    Text(item.title)
                }
            }
            
            Text("SELECTED = \(self.selected ?? "Nothing selected")")            
        }      
    }
}
```

My workaround was to leave the array empty and load the items asynchronously once the List control appears (doesn't work without  DispatchQueue async):
```swift
...
@State var items: [Item] = []
...
List {
    ...
}.onAppear { 
    DispatchQueue.main.async {
        items = [
            Item(title: "Item 1"),
            Item(title: "Item 2"),
            Item(title: "Item 3"),
            Item(title: "Item 4"),     
        ]                
    }
}
```
This time the list doesn't have anything selected and I can select any item without any issue:

Image

This was tested on Ubuntu 24.04.4, with GTK 4.14.5
Swift version 6.2.4

raliclo:
Keep GTK list selection state in the backend so GTK's initial row selection does not update a nil SwiftCrossUI selection binding.

GTK4 verified with P7 on WSLg.

GTK3 verified with P7 and Gtk3Backend build on WSLg.

### DIFF
--- a/Sources/Gtk3Backend/Gtk3Backend.swift
+++ b/Sources/Gtk3Backend/Gtk3Backend.swift
@@ -51,6 +51,7 @@ public final class Gtk3Backend:
     public let restoresWindowFrames = false
 
     var gtkApp: Application
+    private var selectableListStates: [ObjectIdentifier: SelectableListState] = [:]
 
     /// A window to be returned on the next call to ``Gtk3Backend/createWindow``.
     /// This is necessary because Gtk creates a root window no matter what, and
@@ -755,10 +756,12 @@ public final class Gtk3Backend:
         withRowHeights rowHeights: [Int]
     ) {
         let listView = listView as! ListBox
+        let state = state(for: listView)
         listView.removeAll()
         for item in items {
             listView.add(item)
         }
+        preserveNilSelection(of: listView, state: state)
     }
 
     public func setSelectionHandler(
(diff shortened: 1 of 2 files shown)


Sources/Gtk3Backend/Gtk3Backend.swift
```diff
@@ -51,6 +51,7 @@ public final class Gtk3Backend:
     public let restoresWindowFrames = false
 
     var gtkApp: Application
+    private var selectableListStates: [ObjectIdentifier: SelectableListState] = [:]
```
stackotter, line 54: You should subclass `ListBox` and store the state in a stored property of that subclass instead of storing states in a global dictionary. I know that WinUIBackend currently makes use of the state dictionary pattern quite heavily, but I dislike it and there's an open issue related to removing that pattern.

Sources/Gtk3Backend/Gtk3Backend.swift
```diff
@@ -755,35 +756,85 @@ public final class Gtk3Backend:
...
+        state.isProgrammaticSelectionUpdate = true
+        defer { state.isProgrammaticSelectionUpdate = false }
         if let index {
             listView.selectRow(at: index)
         } else {
+            preserveNilSelection(of: listView, state: state)
+        }
+    }
+
+    private func preserveNilSelection(of listView: ListBox, state: SelectableListState) {
```
stackotter, line 806: What is this method even doing? It appears to do almost exactly the same logic two times in a row, but the second time delayed on the main thread. If this is actually doing something, then it seems like it'd be pretty flakey, and I feel like it has a good chance of sporadically failing. It also seems overcomplicated with all of the boolean state variables. If I think back to times that I've solved similar issues, I can envision there being a significantly simpler and easier to reason about solution.

stackotter (changes requested):
I've requested a couple of architectural changes to your solution. I assume that these changes were generated with an LLM, so I haven't put much effort into actually understanding how your implementation is attempting to solve your stated problem. I'd suggest either attempting to solve this issue from scratch yourself, or if you don't want to do that ask the LLM to rethink the issue to find a simpler approach, and to reason about potential race conditions in solutions that it considers. I'd then also ask you to attempt to understand the generated solution, as that will make the review process more efficient.

PR #710, opened 2026-08-16
CI: Fix typo in docs workflow

stackotter:


.github/workflows/build-test-and-docs.yml
```diff
@@ -211,12 +211,12 @@ jobs:
 
     - name: Build AndroidBackend and extract symbol graphs
       uses: ./.github/actions/extract-symbol-graphs
-      working-directory: Examples
       with:
         target: AndroidBackend
         upload: true
         swiftpm-exe: ~/swiftly/.bin/swift
         swifptm-args: --swift-sdk aarch64-unknown-linux-android28
```
Kyle-Ye: Also this line swiftpm-args

Kyle-Ye: Image

stackotter: Oops, thanks for catching that 🤦‍♂️

Issue #717, opened 2026-08-19
Create dummy implementations of remaining unimplemented SwiftUI features


stackotter:
The initial stage of porting an app from SwiftUI to SwiftCrossUI can be rather tedious. It often involves hunting down and commenting out all of the various modifiers that SwiftCrossUI doesn't support yet.

I'm suggesting that we should ease the pain by providing a SwiftCrossUIPortingKit module (name not yet set in stone) that provides dummy implementations of all of the various modifiers and views that SwiftCrossUI doesn't yet support. These dummy implementations should all be marked with some sort of deprecation warning that states that SwiftCrossUI doesn't implement those features yet.

If possible, we should automate the process by parsing the SwiftUI Swift interface files and diffing them against SwiftCrossUI's symbol graph. We'd then need some heuristics for figuring out what the dummy implementation should look like (e.g. for an unsupported `View`, the dummy implementation should implement the `View` protocol with an empty body, and if we detect that a method on `View` is a modifier, we should implement it by simply returning `self`).

I believe that this would be a very interesting process. It'd also be very valuable once completed.

PR #724, opened 2026-08-20
Introduce APITool and PortingKit

stackotter:
This PR introduces APITool: a tool specifically built for diffing the API surfaces of SwiftUI and SwiftCrossUI to help track the progress of SwiftCrossUI's development. Additionally, APITool can generate stubs for missing SwiftUI APIs. APITool's stub generation is very limited for now, and only supports enums. I hope that it can eventually also be used to generate subs for all of the view modifiers that SwiftCrossUI is currently missing, because missing view moddifiers are currently the biggest pain point when porting from SwiftUI to SwiftCrossUI, and having them stubbed out automatically would mean that you can get a compiling app faster, and worry about the details once you already have a tangible proof of concept for your porting effort. The generated stubs live in _SwiftCrossUIPortingKit, which porters can import into their project to access no-op stubs for missing SwiftCrossUI APIs.

This PR is a step towards #717

---

This PR also updates `Scripts/format.sh` to evaluate its first argument (the path to format) relative to the directory that the script was run in (instead of always being relative to the SCUI root directory). I've also updated it to only run SwiftFormat when there are swift files to format, and to only run ktfmt when there are Kotlin files to format (ktfmt produces a failure error status when it can't find any Kotlin files to format).

### DIFF
--- a/APITool/Sources/APITool/Commands/CommonArguments.swift
+++ b/APITool/Sources/APITool/Commands/CommonArguments.swift
@@ -1,0 +1,25 @@
+import ArgumentParser
+import Foundation
+
+struct CommonArguments: ParsableArguments {
+    @Option(
+        name: .customLong("scui-checkout"),
+        help: """
+            The location of the SwiftCrossUI checkout to operate on. Defaults \
+            to the current directory
+            """,
+        transform: { (path: String) in
+            URL(fileURLWithPath: path)
+        }
+    )
+    var swiftCrossUICheckout: URL?
+
+    @Option(
+        name: .customLong("xcode-app"),
+        help: "The location of Xcode.app. Defaults to /Applications/Xcode.app",
+        transform: { (path: String) in
+            URL(fileURLWithPath: path)
+        }
+    )
+    var xcodeApp: URL?
+}
--- a/APITool/Sources/APITool/Commands/CommonArguments.swift
+++ b/APITool/Sources/APITool/Commands/CommonArguments.swift
@@ -15,11 +15,15 @@ struct CommonArguments: ParsableArguments {
     var swiftCrossUICheckout: URL?
 
     @Option(
-        name: .customLong("xcode-app"),
-        help: "The location of Xcode.app. Defaults to /Applications/Xcode.app",
+        name: .customLong("developer-dir"),
+        help: """
+            The location of your Xcode or CLT developer dir. The default \
+            behavior is to fetch the active developer dir using \
+            'xcode-select --print-path'
+            """,
         transform: { (path: String) in
             URL(fileURLWithPath: path)
         }
     )
-    var xcodeApp: URL?
+    var developerDirectory: URL?
 }
--- a/APITool/Sources/APITool/Commands/GenerateCommand.swift
+++ b/APITool/Sources/APITool/Commands/GenerateCommand.swift
@@ -28,7 +28,7 @@ struct GenerateCommand: ParsableCommand {
 
             // Analyze SCUI and generate stubs
             let context = try Context.load(common)
-            try StubGenerator.generateStub(
+            try StubGenerator.generateStubs(
                 context: context,
                 outputDirectory: outputDirectory
             )
--- a/APITool/Sources/APITool/Commands/GenerateCommand.swift
+++ b/APITool/Sources/APITool/Commands/GenerateCommand.swift
@@ -1,0 +1,42 @@
+import ArgumentParser
+import Foundation
+
+struct GenerateCommand: ParsableCommand {
+    static let configuration = CommandConfiguration(
+        commandName: "generate"
+    )
+
+    @OptionGroup
+    var common: CommonArguments
+
+    @Argument
+    var outputDirectory: String
+
+    func run() {
+        do {
+            // Prepare output directory
+            let outputDirectory = URL(fileURLWithPath: outputDirectory)
+            if outputDirectory.exists() {
+                try FileManager.default.removeItem(at: outputDirectory)
+            }
+            try FileManager.default.createDirectory(
+                at: outputDirectory,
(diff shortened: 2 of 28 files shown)


APITool/Sources/APITool/Commands/CommonArguments.swift
```diff
@@ -0,0 +1,25 @@
...
+            """,
+        transform: { (path: String) in
+            URL(fileURLWithPath: path)
+        }
+    )
+    var swiftCrossUICheckout: URL?
+
+    @Option(
+        name: .customLong("xcode-app"),
+        help: "The location of Xcode.app. Defaults to /Applications/Xcode.app",
```
bbrk24: Could we respect `xcode-select` instead? I don't always have an `Xcode.app`, e.g. I might have `Xcode-26.app` and `Xcode-27.app`.

APITool/Sources/APITool/Commands/GenerateCommand.swift
```diff
@@ -0,0 +1,42 @@
...
+    var outputDirectory: String
+
+    func run() {
+        do {
+            // Prepare output directory
+            let outputDirectory = URL(fileURLWithPath: outputDirectory)
+            if outputDirectory.exists() {
+                try FileManager.default.removeItem(at: outputDirectory)
+            }
+            try FileManager.default.createDirectory(
```
bbrk24, line 22: Remove it if it exists, only to immediately create it again?

Sources/_SwiftCrossUIPortingKit/Generated/Enums.swift
bbrk24, line 1: Why are some of them
```swift
@available(...)public enum ...
```
and others are
```swift
@available(...)
public enum ...
```
?

Sources/_SwiftCrossUIPortingKit/Extensions.swift
```diff
@@ -0,0 +1,40 @@
...
+    public static func == (_ lhs: Self, _ rhs: Self) -> Bool {
+        lhs.x == rhs.x && lhs.y == rhs.y
+    }
+}
+
+extension CoreFoundation.CGSize: Swift.Equatable {
+    public static func == (_ lhs: Self, _ rhs: Self) -> Bool {
+        lhs.width == rhs.width && lhs.height == rhs.height
+    }
+}
```
bbrk24: I don't love the retroactive conformances...

bbrk24 (commented):
Just a few drive-by comments. Haven't had a chance to review the meat of this PR.

APITool/Sources/APITool/Commands/CommonArguments.swift
stackotter: Yeah true, I should've thought of that. I originally had it hardcoded for convenience and then just converted it to an option to make it configurable, but that'd be a much better default

APITool/Sources/APITool/Commands/GenerateCommand.swift
stackotter: Just to clear the directory, in case there are additional files that a previous generator generated but the new one won't etc

Sources/_SwiftCrossUIPortingKit/Extensions.swift
stackotter: Yeah, it's not the nicest, but imo it's fine if the porting kit is a bit janky because it's just stubs anyway. People aren't expected to rely on it as a solid API. I added these conformances to avoid having to generate or hand write equatable conformances for structs and enums that depend on them

Sources/_SwiftCrossUIPortingKit/Generated/Enums.swift
stackotter: Oh whoops, I meant to format the generated code. I'll have to update the scripts to do so. When we filter out attributes, the trivia sometimes gets a bit messed up depending on where in the attribute list the excluded attributes originally were.

APITool/Sources/APITool/Commands/GenerateCommand.swift
bbrk24: Ah, I was thinking of the behavior of `rmdir`/`rm -f` where removing it fails when there are children.

APITool/Sources/APITool/Commands/CommonArguments.swift
stackotter: I've resolved this now

Sources/_SwiftCrossUIPortingKit/Generated/Enums.swift
stackotter: I formatted the code, and that alone wasn't enough to fix some odd formatting, so I've also updated the stub generator to fiddle with the trivia a bit and everything looks normal now afaict

Sources/_SwiftCrossUIPortingKit/Extensions.swift
stackotter: I don't believe that there's a good alternative to this for now (other than making a complicated equatable conformance generator for enums that use CG-types somewhere within their signatures), so I'll resolve this for now. But we can strive to make things less janky over time as the generator matures. The generator itself is actually very small rn (most of this PR is all of the supporting code required to parse/analyse/diff SwiftUI and SCUI)

APITool/Sources/APITool/Commands/GenerateCommand.swift
```diff
@@ -0,0 +1,42 @@
...
+                at: outputDirectory,
+                // It's a little more likely for us to catch typos if we don't
+                // automatically create intermediate directories.
+                withIntermediateDirectories: false
+            )
+
+            // Analyze SCUI and generate stubs
+            let context = try Context.load(common)
+            try StubGenerator.generateStub(
```
stackotter: generateStubs

PR #726, opened 2026-08-21
Support view-body `return` expressions by dispatching to expected view lifecycle hooks

### LINKED ISSUE #725: View: View-bodies with explicit `return` render nothing & breaks content.
The following two snippets of code should produce identical output. 

```swift
struct Works: View {
    var body: some View {
        Button("tap") {}
    }
}

struct Broken: View {
    var body: some View {
        return Button("tap") {}
    }
}
```

**Observed:** `Broken` renders an empty container and a zero-sized text view instead of the button.

I discovered this while working on my attempt at a Web backend, but it applies to any backend. Root-causing under `DummyBackend`, `Works` produces a correctly sized `Button` widget with an action; `Broken` produces an inactive zero-sized tree `Container > Container > TextView` - no `Button` widget exists anywhere.

An explicit `return` is the trigger, including with `let`-then-`return`. Single-expression bodies, multi-statement builder bodies, `if`/`else` bodies, and computed-property bodies all work; only the root is affected - nested children work fine. Users can hit this by accident: adding a `let` or a `guard` to an existing body forces the explicit `return` and silently breaks the view.

## Root cause

An explicit `return` inside a view-body opts out of `@ViewBuilder`, so `body`'s compile-time type is the bare view rather than a builder-wrapper. Three default lifecycle implementations in core wrap that body in a fresh `VStack` and forward calls to that, instead of calling into `body`'s own lifecycle method:

- [`defaultAsWidget`](https://github.com/moreSwift/swift-cross-ui/blob/f8bdf05729ae1cd1b7f3ec4e57177a6965052146/Sources/SwiftCrossUI/Views/View.swift#L171-L177)
- [`defaultComputeLayout`](https://github.com/moreSwift/swift-cross-ui/blob/f8bdf05729ae1cd1b7f3ec4e57177a6965052146/Sources/SwiftCrossUI/Views/View.swift#L197-L212)
- [`defaultCommit`](https://github.com/moreSwift/swift-cross-ui/blob/f8bdf05729ae1cd1b7f3ec4e57177a6965052146/Sources/SwiftCrossUI/Views/View.swift#L230-L245)

When `body` is a builder-wrapper, the wrapper's rewrapping is harmless, because it still ends up at the right view. When `body` is bare, the lifecycle events are skipped. ([`defaultChildren`](https://github.com/moreSwift/swift-cross-ui/blob/f8bdf05729ae1cd1b7f3ec4e57177a6965052146/Sources/SwiftCrossUI/Views/View.swift#L137-L143) already forwards to `body` without rewrapping, so it has children without producing a working widget.)

I tested forwarding each lifecycle method to `body` separately, one at a time; omitting any of them leads to ending up in the default implementations, on the wrong view-type:

- `defaultAsWidget`: forwarding to this one leads to the button appearing
- `defaultComputeLayout`:  [`Button.computeLayout`](https://github.com/moreSwift/swift-cross-ui/blob/f8bdf05729ae1cd1b7f3ec4e57177a6965052146/Sources/SwiftCrossUI/Views/Button.swift#L86-L114) is what calls `backend.updateButton(…)` to install the action, so when skipped/mis-delegated, the button stays empty-framed with `action == nil`.
- `defaultCommit`: dispatches commit-phase effects; any other custom `commit` implementations rely on `body.commit(…)` being called as well.

Fixing only one of these calls is insufficient. We need to fix all of them.

A quick scan of the codebase found the following View implementations that would likely be broken if used at the root of a body with an explicit `return`: [`OnTapGestureModifier.commit`](https://github.com/moreSwift/swift-cross-ui/blob/f8bdf05729ae1cd1b7f3ec4e57177a6965052146/Sources/SwiftCrossUI/Views/Modifiers/Handlers/OnTapGestureModifier.swift#L90-L103), which provides `.onTapGesture`, and the `commit` implementations for `NSViewRepresentable` / `UIViewRepresentable` / `GtkWidgetRepresentable` / `WinUIElementRepresentable` / `AndroidViewRepresentable` protocols, which set the widget's size (_I did not test these, just pattern-matched_).

**Scope:** no in-tree view has an explicit `return` today, so nothing in the repository currently misbehaves. This is a coverage gap that users (aka me) will trip into.

## Speculative Reason for the Wrapping

The primary reason this wrap exists seems to be to provide implicit vertical stacking for multi-child View-bodies.

<details><summary>My tracing of why / when</summary>

They get implicitly wrapped in `TupleView2` at the root, and [the `TupleView` extension's `Group(content: self)`](https://github.com/moreSwift/swift-cross-ui/blob/f8bdf05729ae1cd1b7f3ec4e57177a6965052146/Sources/SwiftCrossUI/Views/TupleView.swift#L30-L75) inherits orientation and spacing from environment. The `VStack` wrap in `View`'s defaults is what makes them vertical with default spacing; `b57bd61` introduced that wrap with no explanation.

</details>

The second effect is that a composite-view's body always lays out with a fresh stacking context. Naively forwarding to `body.<lifecycleMethod>` would let the outer stacking context leak: a two-child body inside `HStack(alignment: .top, ...)` would flip from vertical to horizontal stacking (no existing test cases would observe this). The wrap accomplishes the reset by overriding `layoutOrientation`, `layoutAlignment` and `layoutSpacing` before recursing (in [`computeLayout`](https://github.com/moreSwift/swift-cross-ui/blob/f8bdf05729ae1cd1b7f3ec4e57177a6965052146/Sources/SwiftCrossUI/Views/VStack.swift#L74-L84) and again in [`commit`](https://github.com/moreSwift/swift-cross-ui/blob/f8bdf05729ae1cd1b7f3ec4e57177a6965052146/Sources/SwiftCrossUI/Views/VStack.swift#L90-L110)).

## Proposed fix

Forward all three defaults to `body`, and re-supply the neutral stack layout context in each place where Environment is in play. [`GeometryReader.computeLayout`](https://github.com/moreSwift/swift-cross-ui/blob/f8bdf05729ae1cd1b7f3ec4e57177a6965052146/Sources/SwiftCrossUI/Views/GeometryReader.swift#L60-L75) is a precedent for this same pattern.

```diff
-        let vStack = VStack(content: body)
-        return vStack.asWidget(children, backend: backend)
+        body.asWidget(children, backend: backend)
```

```diff
-        let vStack = VStack(content: body)
-        return vStack.computeLayout(
+        body.computeLayout(
             widget,
             children: children,
             proposedSize: proposedSize,
-            environment: environment,
+            environment: environment.with(.default),
             backend: backend
         )
```

…and the identical change in `defaultCommit`.

Adding an internal type to centralize the definition of these environment-key defaults will also clean up distributed duplicate definitions (the environment-key defaults, `VStack.defaultSpacing` plus its alignment default, and `HStack` reaching across to `VStack<EmptyView>.defaultSpacing`) that feel like boilerplate as currently defined.

In my testing, supplying a subset of the three values produced wrong geometry, so I believe that's further evidence for combining these fields into a single default.

Proposed new type name & shape:
```swift
/// The orientation, alignment and spacing that a stack imposes on its children.
struct StackLayoutContext: Sendable {
    var orientation: Orientation
    var alignment: StackAlignment
    var spacing: Int

    /// The context a view lays out in when no stack has established one.
    static let `default` = StackLayoutContext(orientation: .vertical, alignment: .center, spacing: 10)
}
```

With that default implementation, I would make all existing definition-sites forward to `StackLayoutContext.default` -- no more duplicates, but no consumer side changes required.

_(I'm happy to expose `StackLayoutContext` instead of keeping it internal, but I didn't want to bloat the impact to the public area without input from maintainers)_.

I explored a few other approaches not discussed here; I'm happy to go into them if useful.

**LLM Usage:** *I used Claude to root cause the issue I observed (a broken set of views), to build the minimal reproduction, to search for past issues/commits, and to prototype a proposed fix that I analyzed. I manually authored the entire Issue Description. I used Claude to find and insert deep URLs to specific commits/line-references. I take responsibility for everything written above. I stand ready to adjust my issue and work with maintainers on a fix*

fbartho:
Fixes #725.

```swift
struct Works: View {
    var body: some View {
        Button("tap") {}
    }
}

struct FixedWithThisPR: View {
    var body: some View {
        return Button("tap") {}
    }
}
```
I did a deeper dive on the root issue in #725.

This PR:
1. Adds a (currently-_internal_) `StackLayoutContext` to capture the defaults that were distributed. Backwards-compatible shims ensure no user-side code-changes.
2. The lifecycle methods that previously had been pointed at the wrapper now correctly delegate to the body's implementation(s).

The most interesting changes are in `Sources/SwiftCrossUI/Views/View.swift`

**LLM Usage:** *I used Claude to root cause the issue, build the reproduction, and prepare the patch, which I analyzed and directed. I entirely manually authored this description. I take responsibility for everything in this PR.*

### DIFF
--- a/Sources/SwiftCrossUI/Layout/StackLayoutContext.swift
+++ b/Sources/SwiftCrossUI/Layout/StackLayoutContext.swift
@@ -1,0 +1,50 @@
+/// The orientation, alignment and spacing that a stack imposes on its children.
+///
+/// These three values travel together through the environment; grouping them
+/// keeps their defaults in one place and lets a view establish a complete stack
+/// context in a single step.
+public struct StackLayoutContext: Sendable {
+    /// The orientation children are laid out along.
+    public var orientation: Orientation
+    /// The alignment of children perpendicular to ``orientation``.
+    public var alignment: StackAlignment
+    /// The amount of spacing to apply between children.
+    public var spacing: Int
+
+    /// The context a view lays out in when no stack has established one.
+    ///
+    /// Also the context that implicit stacks adopt: ``VStack`` and ``HStack``
+    /// fall back to its spacing and alignment, and a view body composed of
+    /// several views is stacked with it rather than inheriting the surrounding
+    /// stack's context.
+    public static let `default` = StackLayoutContext(
+        orientation: .vertical,
+        alignment: .center,
+        spacing: 10
+    )
+
+    /// Creates a stack layout context.
+    ///
+    /// - Parameters:
+    ///   - orientation: The orientation children are laid out along.
+    ///   - alignment: The alignment of children perpendicular to `orientation`.
+    ///   - spacing: The amount of spacing to apply between children.
+    public init(orientation: Orientation, alignment: StackAlignment, spacing: Int) {
+        self.orientation = orientation
+        self.alignment = alignment
+        self.spacing = spacing
+    }
+}
+
+extension EnvironmentValues {
+    /// Returns a copy of the environment with the given stack layout context
+    /// applied.
+    ///
+    /// - Parameter context: The context to impose on children.
+    /// - Returns: A copy of the environment laying children out with `context`.
+    public func with(_ context: StackLayoutContext) -> Self {
+        with(\.layoutOrientation, context.orientation)
+            .with(\.layoutAlignment, context.alignment)
+            .with(\.layoutSpacing, context.spacing)
+    }
+}
--- a/Sources/SwiftCrossUI/Layout/StackLayoutContext.swift
+++ b/Sources/SwiftCrossUI/Layout/StackLayoutContext.swift
@@ -3,13 +3,13 @@
 /// These three values travel together through the environment; grouping them
 /// keeps their defaults in one place and lets a view establish a complete stack
 /// context in a single step.
-public struct StackLayoutContext: Sendable {
+struct StackLayoutContext: Sendable {
     /// The orientation children are laid out along.
-    public var orientation: Orientation
+    var orientation: Orientation
     /// The alignment of children perpendicular to ``orientation``.
-    public var alignment: StackAlignment
+    var alignment: StackAlignment
     /// The amount of spacing to apply between children.
-    public var spacing: Int
+    var spacing: Int
 
     /// The context a view lays out in when no stack has established one.
     ///
@@ -17,7 +17,7 @@ public struct StackLayoutContext: Sendable {
     /// fall back to its spacing and alignment, and a view body composed of
     /// several views is stacked with it rather than inheriting the surrounding
     /// stack's context.
-    public static let `default` = StackLayoutContext(
+    static let `default` = StackLayoutContext(
         orientation: .vertical,
         alignment: .center,
         spacing: 10
@@ -29,7 +29,7 @@ public struct StackLayoutContext: Sendable {
     ///   - orientation: The orientation children are laid out along.
     ///   - alignment: The alignment of children perpendicular to `orientation`.
     ///   - spacing: The amount of spacing to apply between children.
-    public init(orientation: Orientation, alignment: StackAlignment, spacing: Int) {
+    init(orientation: Orientation, alignment: StackAlignment, spacing: Int) {
         self.orientation = orientation
         self.alignment = alignment
         self.spacing = spacing
@@ -42,7 +42,7 @@ extension EnvironmentValues {
     ///
     /// - Parameter context: The context to impose on children.
     /// - Returns: A copy of the environment laying children out with `context`.
-    public func with(_ context: StackLayoutContext) -> Self {
+    func with(_ context: StackLayoutContext) -> Self {
         with(\.layoutOrientation, context.orientation)
             .with(\.layoutAlignment, context.alignment)
             .with(\.layoutSpacing, context.spacing)
--- a/Sources/SwiftCrossUI/Environment/EnvironmentValues.swift
+++ b/Sources/SwiftCrossUI/Environment/EnvironmentValues.swift
@@ -240,19 +240,19 @@ extension EnvironmentValues {
     ///
     /// Inherited by ``ForEach`` and ``Group`` so that they can be used without
     /// affecting layout.
-    @Entry public var layoutOrientation: Orientation = .vertical
+    @Entry public var layoutOrientation: Orientation = StackLayoutContext.default.orientation
 
     /// The current stack alignment.
     ///
     /// Inherited by ``ForEach`` and ``Group`` so that they can be used without
     /// affecting layout.
-    @Entry public var layoutAlignment: StackAlignment = .center
+    @Entry public var layoutAlignment: StackAlignment = StackLayoutContext.default.alignment
 
     /// The current stack spacing.
     ///
     /// Inherited by ``ForEach`` and ``Group`` so that they can be used without
     /// affecting layout.
-    @Entry public var layoutSpacing: Int = 10
+    @Entry public var layoutSpacing: Int = StackLayoutContext.default.spacing
 
     /// The current font.
     @Entry public var font: Font = .body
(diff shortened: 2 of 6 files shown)


Sources/SwiftCrossUI/Layout/StackLayoutContext.swift
```diff
@@ -0,0 +1,50 @@
...
+    var alignment: StackAlignment
+    /// The amount of spacing to apply between children.
+    var spacing: Int
+
+    /// The context a view lays out in when no stack has established one.
+    ///
+    /// Also the context that implicit stacks adopt: ``VStack`` and ``HStack``
+    /// fall back to its spacing and alignment, and a view body composed of
+    /// several views is stacked with it rather than inheriting the surrounding
+    /// stack's context.
```
mba6comma1, line 19: The current behavior regarding inheriting the surrounding stack's orientation and alignment is the intended one.

fbartho: @kaascevich Just doublechecking your comment / feedback. What are you trying to say?

The current behavior was *discarding* the stack orientation and alignment sometimes in a weird way. The side effect was *not rendering / not-properly rendering* components sometimes (see my example code snippet).

I'd be thrilled to receive a test case if you can explain what behavior you're worried this PR is breaking?
(To my knowledge, this PR preserves the stacking behavior and avoids breaking any existing test cases in the repo).

---

@MiaKoring you hearted their comment, so there must be something I'm missing. I'm not trying to break anybody's code, so if you want to jump in to share what I missed, I'd appreciate you doing so!

MiaKoring: @fbartho My comment would've been on line 199 of View.swift

Rendering the same everywhere, regardless of surrounding context is not what it should be.

It is intentionally rendering differently depending on HStack, VStack or ZStack.

If this broke in some cases that should be reproduced and fixed. Changing the layout behavior to something completely wrong is not it.

fbartho: As I read it, line 199 is the comment line. Are you saying that my explanation in the comment is confusing or wrong? I totally can understand that. I'd be happy to remove the comment if my extra sentence made it more confusing.

You'll notice that I didn't break (and I didn't edit) any existing test cases in the repo, and I added a test file to cover the codepaths related to the bug I observed and the behavior I expected. So I currently believe I didn't break any layout code that was currently being demonstrated in the repo.

MiaKoring: The comment is describing the change correctly as far as I understand it but the change is wrong.

Yes, we may not have a test case covering that, but it's still logically incorrect. Views are supposed to layout differently depending on the innermost enclosing stack.

MiaKoring: @fbartho Sorry, I wasn’t aware this is the current behavior on main. It shouldn’t be this way. I have tests replicating this issue and they all render like a VStack. The issue for that is #728 

So I now understand how you got to the conclusion this was the right behavior.

mba6comma1: So @MiaKoring double-checked the current behavior on `main`, and it *does* appear that we’re doing orientation/alignment inheritance incorrectly (I believe we’re missing a call to `computeStackLayout` somewhere). This is not the intended behavior, either.

It should also be noted, however, that this change appears to me to be wholly unrelated to the stated goal of this PR — I suggest you (*not* your LLM agent) open a separate issue and/or PR so we can discuss this properly, and remove the associated changes from this PR so it’s focused on just one thing.

mba6comma1: Never mind, Mia already opened one in #728.

fbartho: Okay, I'll close this PR in favor of the discussion in #728 -- I'm happy to help get a correct fix across the finish line.

Sorry for the noisiness.

PR #732, opened 2026-08-25
StackLayout now gets propagated to children #728

### LINKED ISSUE #728: Stack Context is not preserved in children
Rendering
```swift
struct TestView: View {
    var body: some View {
        Text("Test")
        Text("Test")
    }
}
```
in an `HStack`, `VStack` or `ZStack` should produce differenty Layouts. This is not the case currently.

I added Test cases in https://github.com/MiaKoring/swift-cross-ui/tree/test/stack-context

HStack and ZStack currently get rendered like a VStack.

MiaKoring:
Resolves #728 

> [!IMPORTANT] 
> This is a breaking change for apps relying on the previous behaviour.

All changes are within the SwiftCrossUI target.

### EnvironmentValues
- added property `usesZStackLayout` to tell `computeStackLayout(_:)`,… to delegate to the ZStack variants
- added property `zStackContentAlignment`. The existing layoutAlignment property cannot be used as `Alignment` is not representable by `StackAlignment` and vice versa.

### LayoutSystem
- `computeStackLayout(_:)` now delegates to `computeZStackLayout(_:)` when `environment.usesZStackLayout` is `true`
- `commitStackLayout(_:)` now delegates to `commitZStackLayout(_:)` when `environment.usesZStackLayout` is `true`
- added `computeZStackLayout(_:)`. It is basically a copy of `computeStackLayout(_:)` with adjustments to make sense for ZStack context.
- added `computeZStackLayouts(_:)`. It is basically a copy of `computeStackLayouts(_:)` with adjustments to make sense for ZStack context.
- added `recomputeZStackCache(_:)`. It is basically a copy of `recomputeCache(_:)` with adjustments to make sense for ZStack context.
- added `commitZStackLayout(_:)`. It is basically a copy of `commitStackLayout(_:)` with adjustments to make sense for ZStack context.

#### How did I get to these changes?
I first copied each method. Then I removed the orientation, spacing and perpendicularOrientation variables as they are irrelevant for a ZStack. 

After seeing highlighted what used it, I removed or changed the rest. (e.g. [component: orientation] was always replaced). Min and max proposed size were replaced with 0,0 and infinity,infinity respectively as there is no axis to constrain that. In computeZStackLayout instead of adding result sizes, max is used (they’re not layed out horizontally or vertically so we only care about the biggest one to determine the size of the stack).

In commit I also replaced the positioning logic a new switch using the zStackContentAlignment.

redistributeSpaceOnCommit is always false inside of the ZStack layout functions, as the function the other ones use is not compatible with ZStack and I don’t know what reason to redistribute it for, as they don’t affect each others placement aside from the bounds of the ZStack they’re placed inside of.

I hope this is all correct, but the tests pass and Examples I tried also look correct to me, so I’m positive.

### HStack & VStack
sets the new `usesZStackLayout` to false on the environment, so the regular functions are used

### View
replaced `vStack.computeLayout(_:)`/`vStack.commit(_:)` with the respective functions on `body`.

### ZStack
replaced the logic in `computeLayout(_:)` and `commit(_:)` with the newly added `LayoutSystem` functions, like seen on `VStack` and `HStack`.

### Tests
- Added `StackContextRendersDifferently` Suite with one test for each stack type all using the namespaced `TestView`.

> [!NOTE]
> The alignment change in `commitZStackLayout(_:)` appears to be quite powerful lol. Only after changing it did the resulting position in the test case change.

### Things to consider:
I did not make any changes to `Spacer` and `Divider`. For `Spacer` it probably doesn’t matter, but `Dividider` will (afaict) layout differently depending on what the first 2D-Stack parent of `ZStack` is. If this is not desirable, please let me know, changing it to a certain orientation or hiding it entirely when `usesZStackLayout` should be no big deal.

### DIFF
--- a/Sources/SwiftCrossUI/Layout/LayoutSystem.swift
+++ b/Sources/SwiftCrossUI/Layout/LayoutSystem.swift
@@ -520,11 +520,6 @@ public enum LayoutSystem {
             cache = StackLayoutCache(
                 priorityGroups: [group],
                 isHidden: results.map(\.participatesInStackLayouts).map(!),
-                // TODO(stackotter): How does SwiftUI handle space reservation during
-                //   relayouts? I feel like it probably doesn't use minimum lengths if
-                //   it didn't already have to during the initial layout pass because
-                //   the alternative would be expensive, but that approach would also
-                //   be a bit inconsistent
                 totalSpacing: 0,
                 totalReservedSpace: 0,
                 minimumLengths: [Double](repeating: 0, count: children.count),
--- a/Sources/SwiftCrossUI/Layout/LayoutSystem.swift
+++ b/Sources/SwiftCrossUI/Layout/LayoutSystem.swift
@@ -132,7 +132,7 @@ public enum LayoutSystem {
         let perpendicularOrientation = orientation.perpendicular
 
         let stackLength = proposedSize[component: orientation]
-        if stackLength == 0 || stackLength == .infinity || stackLength == nil || children.count == 1
+        if stackLength == 0 || stackLength == .infinity || stackLength == nil || children.count <= 1
         {
             var resultLength: Double = 0
             var resultWidth: Double = 0
@@ -494,21 +494,22 @@ public enum LayoutSystem {
         backend: Backend,
         inheritStackLayoutParticipation: Bool = false
     ) -> ViewLayoutResult {
-        if children.count == 1 {
-            var maxWidth: Double = 0
-            var maxHeight: Double = 0
+        let specialSizes: [Double?] = [nil, .infinity]
+        if children.count <= 1 || (specialSizes.contains(proposedSize.width) && specialSizes.contains(proposedSize.height)) {
+            var stackWidth: Double = 0
+            var stackHeight: Double = 0
             var results: [ViewLayoutResult] = []
             for child in children {
                 let result = child.computeLayout(
                     proposedSize: proposedSize,
                     environment: environment
                 )
-                maxWidth = max(maxWidth, result.size.width)
-                maxHeight = max(maxHeight, result.size.height)
+                stackWidth = max(stackWidth, result.size.width)
+                stackHeight = max(stackHeight, result.size.height)
                 results.append(result)
             }
 
-            let size = ViewSize(maxWidth, maxHeight)
+            let size = ViewSize(stackWidth, stackHeight)
 
             // In this case, flexibility and layout priority don't matter. We set
             // the grouping to the trivial grouping so that commitStackLayout
@@ -519,11 +520,11 @@ public enum LayoutSystem {
             )
             cache = StackLayoutCache(
                 priorityGroups: [group],
-                isHidden: results.map(\.participatesInStackLayouts).map(!),
+                isHidden: [false],
                 totalSpacing: 0,
                 totalReservedSpace: 0,
                 minimumLengths: [Double](repeating: 0, count: children.count),
-                redistributeSpaceOnCommit: false
+                redistributeSpaceOnCommit: proposedSize.width == nil || proposedSize.height == nil
             )
 
             return ViewLayoutResult(
@@ -545,8 +546,7 @@ public enum LayoutSystem {
             of: children,
             proposedSize: proposedSize,
             cache: cache,
-            environment: environment,
-            ignoreHiddenChildrenEntirely: false
+            environment: environment
         )
 
         var size = ViewSize.zero
@@ -570,29 +570,18 @@ public enum LayoutSystem {
         let minimumProposedSize = ProposedViewSize(0, 0)
         let maximumProposedSize = ProposedViewSize(.infinity, .infinity)
 
-        var isHidden = [Bool](repeating: false, count: children.count)
         var priorities = [Double](repeating: 0, count: children.count)
 
-        let flexibilities = children.enumerated().map { i, child in
+        for (i, child) in children.enumerated() {
             let minimumResult = child.computeLayout(
                 proposedSize: minimumProposedSize,
                 environment: environment.with(\.allowLayoutCaching, true)
             )
-            let maximumResult = child.computeLayout(
-                proposedSize: maximumProposedSize,
-                environment: environment.with(\.allowLayoutCaching, true)
-            )
 
-            isHidden[i] = !minimumResult.participatesInStackLayouts
             priorities[i] = minimumResult.preferences.layoutPriority
-
-            let widthFlexibility = maximumResult.size.width - minimumResult.size.width
-            let heightFlexibility = maximumResult.size.height - minimumResult.size.height
-
-            return widthFlexibility + heightFlexibility
         }
 
-        let sortedChildren = zip(children.indices, zip(priorities.map(-), flexibilities))
+        let sortedChildren = zip(children.indices, priorities.map(-))
             .sorted { first, second in
                 // Sort by descending priority and then by ascending flexibility
                 first.1 <= second.1
@@ -632,7 +621,7 @@ public enum LayoutSystem {
 
         return StackLayoutCache(
             priorityGroups: priorityGroups,
-            isHidden: isHidden,
+            isHidden: [false],
             totalSpacing: 0,
             totalReservedSpace: 0,
             minimumLengths: [],
@@ -640,63 +629,48 @@ public enum LayoutSystem {
         )
     }
 
-    /// The main ZStack layout space allocation algorithm. Used during computeLayout.
+    /// The main ZStack layout algorithm. Used during computeZStackLayout
     @MainActor
     static func computeZStackLayouts(
         of children: [LayoutableChild],
         proposedSize: ProposedViewSize,
         cache: StackLayoutCache,
-        environment: EnvironmentValues,
-        ignoreHiddenChildrenEntirely: Bool
+        environment: EnvironmentValues
     ) -> [ViewLayoutResult] {
         var renderedChildren = [ViewLayoutResult](
             repeating: .leafView(size: .zero),
             count: children.count
         )
+        
+        var currentProposedSize = proposedSize
 
         for group in cache.priorityGroups {
-            var childrenRemaining = group.children.count { index in
-                !cache.isHidden[index]
-            }
-
+            var maxSize: ProposedViewSize = .zero
+            
             for index in group.children {
                 let child = children[index]
 
-                // No need to render visible children.
-                if cache.isHidden[index] {
-                    if ignoreHiddenChildrenEntirely {
-                        continue
-                    }
-
-                    // Update child in case it has just changed from visible to hidden,
-                    // and to make sure that the view is still hidden (if it's not then
-                    // it's a bug with either the view or the layout system).
-                    let result = child.computeLayout(
-                        proposedSize: .zero,
-                        environment: environment
-                    )
-                    if result.participatesInStackLayouts {
-                        logger.warning(
-                            "hidden view became visible on second update; layout may break",
-                            metadata: [
-                                "view": "\(child.tag ?? "<unknown type>")"
-                            ]
-                        )
-                    }
-                    renderedChildren[index] = result
-                    renderedChildren[index].participateInStackLayoutsWhenEmpty = false
-                    renderedChildren[index].size = .zero
-                    continue
-                }
-
                 let childResult = child.computeLayout(
-                    proposedSize: proposedSize,
+                    proposedSize: currentProposedSize,
                     environment: environment
                 )
+                
+                if let currentMaxWidth = maxSize.width {
+                    maxSize.width = max(childResult.size.width, currentMaxWidth)
+                } else {
+                    maxSize.width = childResult.size.width
+                }
+                
+                if let currentMaxHeight = maxSize.height {
+                    maxSize.height = max(childResult.size.height, currentMaxHeight)
+                } else {
+                    maxSize.height = childResult.size.height
+                }
 
                 renderedChildren[index] = childResult
-                childrenRemaining -= 1
             }
+            
+            currentProposedSize = maxSize
         }
 
         return renderedChildren
@@ -715,13 +689,17 @@ public enum LayoutSystem {
         let alignment = environment.zStackContentAlignment
         backend.setSize(of: container, to: size.vector)
 
-        guard !cache.redistributeSpaceOnCommit else {
-            fatalError("unreachable")
+        if cache.redistributeSpaceOnCommit {
+            _ = computeZStackLayouts(
+                of: children,
+                proposedSize: ProposedViewSize(size.width, size.height),
+                cache: cache,
+                environment: environment
+            )
         }
-
+        
         let renderedChildren = children.map { $0.commit() }
 
-        var position = Position.zero
         for (index, child) in renderedChildren.enumerated() {
             // Avoid the whole iteration if the child is hidden. If there
             // are weird positioning issues for views that do strange things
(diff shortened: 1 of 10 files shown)


Sources/SwiftCrossUI/Views/ZStack.swift
```diff
@@ -85,31 +75,17 @@ public struct ZStack<Content: View>: View {
...
-
-        backend.setSize(of: widget, to: size.vector)
+        var cache = (children as? TupleViewChildren)?.stackLayoutCache ?? StackLayoutCache.initial
+        LayoutSystem.commitZStackLayout(
+            container: widget,
+            children: layoutableChildren(backend: backend, children: children),
+            cache: &cache,
+            layout: layout,
+            environment: environment
+                .with(\.usesZStackLayout, true)
```
stackotter, line 86: Reset the `layoutOrientation` to `ZStack` (as discussed in PR comment)

Sources/SwiftCrossUI/Layout/LayoutSystem.swift
```diff
@@ -459,4 +483,325 @@ public enum LayoutSystem {
...
+        for (index, child) in renderedChildren.enumerated() {
+            // Avoid the whole iteration if the child is hidden. If there
+            // are weird positioning issues for views that do strange things
+            // then this could be the cause.
+            if !child.participatesInStackLayouts {
+                continue
+            }
+
+            // Compute alignment
+            switch alignment {
```
stackotter: We have the `Alignment.position(ofChild:in:)` method for exactly this situation. I think you should be able to copy paste its implementation to create an overload for Doubles and then just use that here. The first parameter is the size of the child, and the second is the size of the parent. The return value is the position that the child should have in the parent given the alignment.

stackotter: You should be able to avoid the `var position` mutable variable and instead just store the return value of the `Alignment.position(ofChild:in:)` call directly to a `let` variable, and then use `LayoutSystem.roundSize` to round each component of the position. Or you could add a new `init` to `Position` that accepts a `SIMD2<Double>` and then do `Position(position).vector` to do the rounding.

Sources/SwiftCrossUI/Layout/LayoutSystem.swift
```diff
@@ -459,4 +483,325 @@ public enum LayoutSystem {
...
+            isHidden[i] = !minimumResult.participatesInStackLayouts
+            priorities[i] = minimumResult.preferences.layoutPriority
+
+            let widthFlexibility = maximumResult.size.width - minimumResult.size.width
+            let heightFlexibility = maximumResult.size.height - minimumResult.size.height
+
+            return widthFlexibility + heightFlexibility
+        }
+
+        let sortedChildren = zip(children.indices, zip(priorities.map(-), flexibilities))
```
stackotter: I believe that we should be able to avoid flexibility sorting in ZStacks, because within each layout priority group the size of one child can't affect the size proposed to another. Our existing ZStack implementation doesn't do any flexibility sorting.

Sources/SwiftCrossUI/Layout/LayoutSystem.swift
```diff
@@ -459,4 +483,325 @@ public enum LayoutSystem {
...
+            let minimumResult = child.computeLayout(
+                proposedSize: minimumProposedSize,
+                environment: environment.with(\.allowLayoutCaching, true)
+            )
+            let maximumResult = child.computeLayout(
+                proposedSize: maximumProposedSize,
+                environment: environment.with(\.allowLayoutCaching, true)
+            )
+
+            isHidden[i] = !minimumResult.participatesInStackLayouts
```
stackotter: I think we should also be able to completely ignore `isHidden` in ZStacks because we don't have to handle spacing. The `isHidden` mechanism is purely for avoiding putting spacing around invisible elements (such as disabled `if` blocks).

Sources/SwiftCrossUI/Layout/LayoutSystem.swift
```diff
@@ -459,4 +483,325 @@ public enum LayoutSystem {
...
+        proposedSize: ProposedViewSize,
+        environment: EnvironmentValues
+    ) -> StackLayoutCache {
+        let minimumProposedSize = ProposedViewSize(0, 0)
+        let maximumProposedSize = ProposedViewSize(.infinity, .infinity)
+
+        var isHidden = [Bool](repeating: false, count: children.count)
+        var priorities = [Double](repeating: 0, count: children.count)
+
+        let flexibilities = children.enumerated().map { i, child in
```
stackotter: Given that we don't have to do flexibility sorting, we should be able to eliminate this entirely.

Sources/SwiftCrossUI/Layout/LayoutSystem.swift
```diff
@@ -459,4 +483,325 @@ public enum LayoutSystem {
...
+    static func computeZStackLayout<Backend: BaseAppBackend>(
+        container: Backend.Widget,
+        children: [LayoutableChild],
+        cache: inout StackLayoutCache,
+        proposedSize: ProposedViewSize,
+        environment: EnvironmentValues,
+        backend: Backend,
+        inheritStackLayoutParticipation: Bool = false
+    ) -> ViewLayoutResult {
+        if children.count == 1 {
```
stackotter: This `if` statement is essentially a fast path for cases where we don't have to care about ordering. In the case of ZStacks that just means cases where layout priority doesn't have an effect (because we ignore flexibility entirely in ZStacks).

I think you're missing the case where the ZStack has no children (which can happen in a ForEach). It looks like I also missed that case in the original computeStackLayout implementation (whoops).

My intuition tells me that order also doesn't matter when both axes of the proposal are `nil`, or `.infinity`. We may also be able to take the fast path during a minimum size proposal, because all we care about during a minimum size proposal is the overall size of the ZStack and not the internal layout, but there are cases where you could get a 0x0 proposal outside of a minimum size proposal, and I can think of some pathological cases where you'd get a different layout with and without layout sorting (or at least I feel like there might be). To get a definitive answer we'd probably have to investigate SwiftUI's ZStack layout algorithm by constructing some really weird examples. But IMO it shouldn't matter much for now.

So in short;

```swift
let specialSizes: [Double?] = [nil, .infinity]
if children.count <= 1 || (specialSizes.contains(proposedSize.width) && specialSizes.contains(proposedSize.height)) {
    ...
}
```

Sources/SwiftCrossUI/Layout/LayoutSystem.swift
```diff
@@ -459,4 +483,325 @@ public enum LayoutSystem {
...
+                let result = child.computeLayout(
+                    proposedSize: proposedSize,
+                    environment: environment
+                )
+                maxWidth = max(maxWidth, result.size.width)
+                maxHeight = max(maxHeight, result.size.height)
+                results.append(result)
+            }
+
+            let size = ViewSize(maxWidth, maxHeight)
```
stackotter: When reading this I associated `maxWidth` and `maxHeight` with maximum size proposals. These variables could be renamed to `stackWidth` and `stackHeight` for clarity imo

Sources/SwiftCrossUI/Layout/LayoutSystem.swift
```diff
@@ -459,4 +483,325 @@ public enum LayoutSystem {
...
+            // In this case, flexibility and layout priority don't matter. We set
+            // the grouping to the trivial grouping so that commitStackLayout
+            // effectively ignores flexibility.
+            let group = LayoutPriorityGroup(
+                children: Array(children.indices)[...],
+                priority: 0
+            )
+            cache = StackLayoutCache(
+                priorityGroups: [group],
+                isHidden: results.map(\.participatesInStackLayouts).map(!),
```
stackotter: Can just be all `false` because we don't care about `isHidden` (computing it in this case would make it look like we did care about it).

Sources/SwiftCrossUI/Layout/LayoutSystem.swift
```diff
@@ -459,4 +483,325 @@ public enum LayoutSystem {
...
+                totalSpacing: 0,
+                totalReservedSpace: 0,
+                minimumLengths: [Double](repeating: 0, count: children.count),
+                redistributeSpaceOnCommit: false
```
stackotter: If either axis of the proposed size is nil, then `redistributeSpaceOnCommit` should be true.

Sources/SwiftCrossUI/Layout/LayoutSystem.swift
```diff
@@ -459,4 +483,325 @@ public enum LayoutSystem {
...
+            priorityGroups: priorityGroups,
+            isHidden: isHidden,
+            totalSpacing: 0,
+            totalReservedSpace: 0,
+            minimumLengths: [],
+            redistributeSpaceOnCommit: false
+        )
+    }
+
+    /// The main ZStack layout space allocation algorithm. Used during computeLayout.
```
stackotter: This isn't really a space allocation algorithm, so we should probably just say `The main ZStack layout algorithm. Used during computeZStackLayout`

Sources/SwiftCrossUI/Layout/LayoutSystem.swift
```diff
@@ -459,4 +483,325 @@ public enum LayoutSystem {
...
+        for group in cache.priorityGroups {
+            var childrenRemaining = group.children.count { index in
+                !cache.isHidden[index]
+            }
+
+            for index in group.children {
+                let child = children[index]
+
+                // No need to render visible children.
+                if cache.isHidden[index] {
```
stackotter: You can remove this whole if because we don't care about `isHidden` in ZStacks

Sources/SwiftCrossUI/Layout/LayoutSystem.swift
```diff
@@ -459,4 +483,325 @@ public enum LayoutSystem {
...
+        environment: EnvironmentValues,
+        ignoreHiddenChildrenEntirely: Bool
+    ) -> [ViewLayoutResult] {
+        var renderedChildren = [ViewLayoutResult](
+            repeating: .leafView(size: .zero),
+            count: children.count
+        )
+
+        for group in cache.priorityGroups {
+            var childrenRemaining = group.children.count { index in
```
stackotter: `childrenRemaining was written to, but never read` (how's my Swift compiler larping?)

Sources/SwiftCrossUI/Layout/LayoutSystem.swift
```diff
@@ -459,4 +483,325 @@ public enum LayoutSystem {
...
+                }
+
+                let childResult = child.computeLayout(
+                    proposedSize: proposedSize,
+                    environment: environment
+                )
+
+                renderedChildren[index] = childResult
+                childrenRemaining -= 1
+            }
```
stackotter, line 673: Each priority group's proposed size should be whatever the size of the stack was after the previous priority group got processed. So at first `proposedSize` will be the input size, and then at the end of the outer loop it should get reassigned to the current size of the stack (the max of all child sizes so far). The best way to compute that new proposed size would be to keep a running maximum in the inner loop so that you've already computed it by the time the inner loop finishes for a given layout priority group.

Sources/SwiftCrossUI/Layout/LayoutSystem.swift
```diff
@@ -459,4 +483,325 @@ public enum LayoutSystem {
...
+        cache: inout StackLayoutCache,
+        layout: ViewLayoutResult,
+        environment: EnvironmentValues,
+        backend: Backend
+    ) {
+        let size = layout.size
+        let alignment = environment.zStackContentAlignment
+        backend.setSize(of: container, to: size.vector)
+
+        guard !cache.redistributeSpaceOnCommit else {
```
stackotter: If `redistributeSpaceOnCommit` is true, just call `computeZStackLayouts` one more time. You should be able to just discard the results like we do in `commitStackLayout`, because you get the results after the if block when you commit all of the children.

Tests/SwiftCrossUITests/StackContextRendersDifferently.swift
```diff
@@ -0,0 +1,105 @@
+import Testing
+import DummyBackend
+@testable @_spi(Backends) import SwiftCrossUI
+
+@MainActor
+@Suite("View renders differently depending on the surrounding stack context")
+struct StackContextRendersDifferently {
+    struct TestView: View {
+        // The view body contains the same text two times, cause that is a simple
+        // way to check the expected behaviour. The bounds are expected to be 1:1
```
stackotter: Instead of `The bounds are expected to be 1:1`, it might be better to say `The children are expected to be identical`. I originally read `1:1` as an aspect ratio (meaning that the bounds of TestView are meant to be a square, which clearly isn't true and confused me for a bit).

Tests/SwiftCrossUITests/StackContextRendersDifferently.swift
```diff
@@ -0,0 +1,105 @@
...
+
+        guard let container = lastParent as? DummyBackend.Container else {
+            Issue.record("Parent of text views unexpectedly wasn't a container.")
+            return
+        }
+
+        #expect(container.children.count == 2)
+
+        guard
+            container.children.count == 2,
```
stackotter: You're checking the children count twice

Tests/SwiftCrossUITests/StackContextRendersDifferently.swift
```diff
@@ -0,0 +1,105 @@
...
+            let (_, firstPosition) = container.children.first,
+            let (_, secondPosition) = container.children.last
```
stackotter: I think that I prefer taking `children[0]` and `children[1]` because that's more clearly getting the first two children. We already checked the size of the `children` array doing more conditional checks about the size of the array seems odd to me

stackotter (changes requested):
This is a great initial implementation for a feature in what can be quite a confusing part of SwiftCrossUI! The unit tests are great imo, and should easily catch regressions with being overly restrictive. I've left a bunch of comments requesting changes, but many of them are about things to remove that were copied over from the V/H stack layout algorithm, so there hopefully aren't too many complicated requests.

Sources/SwiftCrossUI/Layout/LayoutSystem.swift
MiaKoring: not sure why I would need a double overload, all parameters are integers already and the result is also consumed as an integer?

If it’s about precision I could also just change the int one to convert them to doubles internally

Sources/SwiftCrossUI/Layout/LayoutSystem.swift
MiaKoring: Does that mean there should be one priority group containing all of them? I’m not entirely sure to what includes “avoiding flexibility sorting"

Sources/SwiftCrossUI/Layout/LayoutSystem.swift
MiaKoring: I updated the regular layout method to also have <=1 instead of == 1 now

Sources/SwiftCrossUI/Layout/LayoutSystem.swift
MiaKoring: oh lol I missed that. It’s only used by propedChildSize, which obviously doesn’t exist in ZStack because they don’t affect each other

Tests/SwiftCrossUITests/StackContextRendersDifferently.swift
MiaKoring: yes, due to #expect not being an early exit and the guard not being specific about which of these cases failed. So if the Issue.record is triggered and we have the failing expect too, we know what’s wrong immediately, without the need to add debug statements.

Tests/SwiftCrossUITests/StackContextRendersDifferently.swift
MiaKoring: nvm, ignore my previous comment lol

Sources/SwiftCrossUI/Layout/LayoutSystem.swift
stackotter: I was suggesting it because your existing code does the logic with doubles before converting them to integers, and if you instead converted to integers first you'd get a slightly less precise result.

Sources/SwiftCrossUI/Layout/LayoutSystem.swift
```diff
@@ -570,29 +570,18 @@ public enum LayoutSystem {
         let minimumProposedSize = ProposedViewSize(0, 0)
         let maximumProposedSize = ProposedViewSize(.infinity, .infinity)
```
stackotter: `maximumProposedSize` is now unused

Sources/SwiftCrossUI/Layout/LayoutSystem.swift
```diff
@@ -570,29 +570,18 @@ public enum LayoutSystem {
...
 
-        var isHidden = [Bool](repeating: false, count: children.count)
         var priorities = [Double](repeating: 0, count: children.count)
 
-        let flexibilities = children.enumerated().map { i, child in
+        for (i, child) in children.enumerated() {
```
stackotter: This can be a `map` that produces `priorities` now, because it's only looking for a single value. e.g. `let priorities = children.enumerated().map ...`

Sources/SwiftCrossUI/Layout/LayoutSystem.swift
```diff
@@ -570,29 +570,18 @@ public enum LayoutSystem {
...
-            let widthFlexibility = maximumResult.size.width - minimumResult.size.width
-            let heightFlexibility = maximumResult.size.height - minimumResult.size.height
-
-            return widthFlexibility + heightFlexibility
         }
 
-        let sortedChildren = zip(children.indices, zip(priorities.map(-), flexibilities))
+        let sortedChildren = zip(children.indices, priorities.map(-))
             .sorted { first, second in
                 // Sort by descending priority and then by ascending flexibility
```
stackotter: `// Sort by descending priority`

Sources/SwiftCrossUI/Layout/LayoutSystem.swift
```diff
@@ -715,93 +689,28 @@ public enum LayoutSystem {
...
+        if cache.redistributeSpaceOnCommit {
+            _ = computeZStackLayouts(
+                of: children,
+                proposedSize: ProposedViewSize(size.width, size.height),
+                cache: cache,
+                environment: environment
+            )
         }
-
+
```
stackotter: Trailing spaces (might break things once I merge your formatting CI job)

Sources/SwiftCrossUI/Layout/LayoutSystem.swift
stackotter: What you've got now is good. You still sort by layout priority but just not by flexibility within each group.

stackotter (changes requested):
Thanks for those resolutions. Basically everything looks great; I just found a few things to fix from your new changes, and commented on the issue about the Alignment.position double overload (regarding precision)

Sources/SwiftCrossUI/Layout/LayoutSystem.swift
MiaKoring: does it? child.size.width and size.width are both integers

stackotter (approved):
Thanks for all of those resolutions!

Sources/SwiftCrossUI/Layout/LayoutSystem.swift
stackotter: I've realised that rounding before the positioning is probably actually what we want to do, because the positioning has to be correct with respect to the displayed sizes of the views, not with respect to the pre-rounding sizes. My bad on that one, your implementation is correct afaict

PR #733, opened 2026-08-25
Update benchmarks and add environment-heavy benchmark

### LINKED ISSUE #731: Investigate using TreeDictionary for EvironmentValues
- Faster CoW, as it doesn't have to copy the entire storage on every write.
- Algorithmically slower lookup (O(1) vs O(log n)), though it may not matter.
- Requires a dependency on swift-collections, if we don't already have one (I haven't checked).

bbrk24:
Closes #731 

Sometimes, a negative result is still a result. On my MacBook Air, the new benchmark takes about 140ms when the backing storage is `Dictionary`, and about 150ms when the backing storage is `TreeDictionary`. The grid benchmark also performs worse with `TreeDictionary` than with `Dictionary`, but by a much smaller margin. Therefore, I am closing #731 without updating `EnvironmentValues`.

I did have to make some minor adjustments to the existing benchmark files, as they were a bit out of date.

### DIFF
--- a/Benchmarks/LayoutPerformanceBenchmark/EnvironmentHeavyView.swift
+++ b/Benchmarks/LayoutPerformanceBenchmark/EnvironmentHeavyView.swift
@@ -1,0 +1,144 @@
+import SwiftCrossUI
+import Foundation
+
+struct V1: View {
+    @Binding var toggleState: Bool
+    @Binding var date: Date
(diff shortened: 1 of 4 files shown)


stackotter (approved):
Looks great, thanks for PRing this work. We'll have to start benchmarking in CI at some point.

PR #734, opened 2026-08-27
Added action and script to check formatting

MiaKoring:
The script is a copy of `format.sh` with `--lint` added to the swiftformat command and `--dry-run --set-exit-if-changed` for ktfmt.

Before running them I added `set -eo pipefail` to make sure CI fails if either of the checks fail.

I’m going to push some changes that I will remove again to  make sure CI fails as expected.


Scripts/check-formatted.sh
```diff
@@ -0,0 +1,69 @@
+#!/bin/bash
```
stackotter: Would it be possible to conditionalise the existing `format.sh` script to understand an environment variable that tells it to lint, rather than making a copy of the script?

I recently made a bunch of changes to the format.sh script, and I'd hate to have to do that to two scripts or have the scripts get out of sync.

.github/workflows/check-formatting.yml
```diff
@@ -0,0 +1,21 @@
+name: Check formatting | Run './Scripts/format.sh' on fail.
```
stackotter: I interpreted this name as meaning that it runs `format.sh` on fail, so it could be good to update the wording to make it clear that it's an instruction for the PR author. Make `Check formatting; run './Scripts/format.sh` if this fails`? Not a huge deal though

.github/workflows/check-formatting.yml
```diff
@@ -0,0 +1,21 @@
...
+  contents: read
+
+jobs:
+  check:
+    name: "check-formatted"
+    runs-on: macos-15
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd #6.0.2
+        with:
+          persist-credentials: false
```
stackotter, line 14: Why should we use a specific commit hash instead of just `actions/checkout@v6` like our other workflows do?

.github/workflows/check-formatting.yml
```diff
@@ -0,0 +1,21 @@
+name: Check formatting | Run './Scripts/format.sh' on fail.
+
+on:
+  pull_request:
+    branches: [main]
+    
+permissions:
+  contents: read
```
stackotter: I think this is unnecessary?

.github/workflows/check-formatting.yml
```diff
@@ -0,0 +1,21 @@
...
+
+jobs:
+  check:
+    name: "check-formatted"
+    runs-on: macos-15
```
stackotter: We should run this on Linux if possible to avoid clogging up our macOS workers, but perhaps that would be too annoying to setup? It'd be worth looking into at least

stackotter (changes requested):

.github/workflows/check-formatting.yml
MiaKoring: I copied this from my vein action.

a pinned commit hash afaik helps prevent supply chain attacks.

.github/workflows/check-formatting.yml
MiaKoring: also a thing I do on vein to make it harder to do weird things. it’s just a permission restriction but I can remove it

.github/workflows/check-formatting.yml
stackotter: Figured it might be something to do with supply chain attacks. I'm happy for us to update all of our actions to do that given the prevalence of supply chain attacks in the node ecosystem recently, but maybe in a separate PR

.github/workflows/check-formatting.yml
stackotter: I just checked and our default workflow permissions are "Read repository contents and packages", so I think we're safe to leave this out, but it is probably a good idea in general

stackotter (approved):
Looks good now

PR #735, opened 2026-08-28
Color pickers

### LINKED ISSUE #723: ColorPicker

bbrk24:
Closes #723

Implemented on all major backends except AndroidBackend, which has no native color picker. Not available on iOS 13 or tvOS.


Sources/UIKitBackend/UIKitBackend+Control.swift
```diff
@@ -652,4 +685,37 @@ extension UIKitBackend: BackendFeatures.TapGestures {
             }
         }
     }
+
+    extension UIKitBackend: BackendFeatures.ColorPickers {
+        public func createColorPicker() -> any WidgetProtocol {
+            if #available(iOS 14, macCatalyst 14, *) {
```
mba6comma1: You could instead mark the entire `ColorPickers` conformance `@available(iOS 14, macCatalyst 14, *)` and then get rid of this check. (It'll be checked for existence at runtime by the `CastBackend` macro.)

bbrk24: Huh, I didn't know that's how that works. Updated

Examples/Sources/ControlsExample/ControlsApp.swift
```diff
@@ -201,6 +202,20 @@ struct ControlsApp: App {
                                 }
                             #endif
                         #endif
+
+                        #if !os(tvOS) && !canImport(AndroidBackend)
+                            if #available(iOS 14, macCatalyst 14, *) {
```
stackotter, line 209: It looks like we'll have to update our ViewBuilder to support `buildLimitedAvailability`? (according to the CI warnings)

Sources/SwiftCrossUI/Views/HorizontalControlSpacer.swift
```diff
@@ -0,0 +1,13 @@
+/// A view that behaves like a spacer on certain device classes, for `HStack`-based controls like
+/// ``ColorPicker`` and switch-style ``Toggle``.
+internal struct HorizontalControlSpacer: View {
```
stackotter: `internal` is redundant

Sources/SwiftCrossUI/Backend/BackendFeatures/ColorPickers.swift
```diff
@@ -0,0 +1,32 @@
...
+        /// Update a color picker.
+        /// - Parameters:
+        ///   - colorPicker: The color picker widget to update.
+        ///   - supportsOpacity: If `false`, do not allow the user to change the alpha value of the
+        ///   color. If the color provided to ``setValue(ofColorPicker:to:)`` has an alpha value
+        ///   less than 1, leave it at that value.
+        ///   - environment: The current environment.
+        ///   - onChange: The action to perform when the selected color changes. This handler
+        ///   replaces any existing change handlers and is called whenever a selection is made, even
+        ///   if the same color is picked again.
```
stackotter: Indent wrapped line by two more spaces (to align with the content of the first line)

Sources/Gtk3Backend/Gtk3Backend+ColorPicker.swift
```diff
@@ -0,0 +1,72 @@
...
+    public func updateColorPicker(
+        _ colorPicker: Widget,
+        supportsOpacity: Bool,
+        environment: EnvironmentValues,
+        onChange: @escaping (SwiftCrossUI.Color.Resolved) -> Void
+    ) {
+        let colorButton = colorPicker as! CustomColorButton
+        colorButton.sensitive = environment.isEnabled
+        colorButton.useAlpha = supportsOpacity
+        colorButton.colorSet = { [unowned colorButton] (_) in
```
stackotter: The first parameter to the closure is the color button itself, so we should be able to avoid the unowned capture here

Sources/GtkBackend/GtkBackend+ColorPicker.swift
```diff
@@ -0,0 +1,63 @@
...
+    public func updateColorPicker(
+        _ colorPicker: Widget,
+        supportsOpacity: Bool,
+        environment: EnvironmentValues,
+        onChange: @escaping (SwiftCrossUI.Color.Resolved) -> Void
+    ) {
+        let colorButton = colorPicker as! CustomColorButton
+        colorButton.sensitive = environment.isEnabled
+        colorButton.useAlpha = supportsOpacity
+        colorButton.colorSet = { [unowned colorButton] (_) in
```
stackotter: Should be able to avoid the unowned capture by using the first parameter of the closure instead (which is a reference to the color button)

Sources/WinUIBackend/WinUIBackend+ColorPicker.swift
```diff
@@ -0,0 +1,114 @@
...
+            dialog.colorPicker.isAlphaEnabled = self.supportsOpacity
+            dialog.xamlRoot = self.xamlRoot
+            self.dialog = dialog
+
+            if let color {
+                dialog.colorPicker.color = color
+            }
+
+            do {
+                let promise = try dialog.showAsync()!
```
stackotter, line 59: I think in their current form ColorPickers might not work from inside sheets (because it's count as showing two dialogs at once, which WinUI doesn't support). We'll eventually need a more general alert/sheet/colorpicker stacking system. Don't have to worry about it in this PR, but it's probably worth opening an issue about (I'll do that when I submit this review)

Sources/WinUIBackend/WinUIBackend+ColorPicker.swift
```diff
@@ -0,0 +1,114 @@
...
+    public func updateColorPicker(
+        _ colorPicker: Widget,
+        supportsOpacity: Bool,
+        environment: EnvironmentValues,
+        onChange: @escaping (SwiftCrossUI.Color.Resolved) -> Void
+    ) {
+        let colorButton = colorPicker as! ColorButton
+        colorButton.supportsOpacity = supportsOpacity
+        colorButton.isEnabled = environment.isEnabled
+        colorButton.onChange = { onChange(.init(uwpColor: $0)) }
```
stackotter: Spell out the `init` fully

stackotter (changes requested):
This will be super nice to have in SCUI. I've just left some comment requesting changes (most of which are small).

Examples/Sources/ControlsExample/ControlsApp.swift
bbrk24: I can add that. Should I `#if` it out on Windows and Linux?

Sources/Gtk3Backend/Gtk3Backend+ColorPicker.swift
bbrk24: It's also a Swift wrapper around the object, not the object itself. At least with SwiftJava, just because the Java object is an instance of the subclass doesn't mean the Swift wrapper is, and I didn't bother to check whether the `as!` cast would succeed. I'll try that.

Sources/SwiftCrossUI/Backend/BackendFeatures/ColorPickers.swift
bbrk24: Updated

Sources/Gtk3Backend/Gtk3Backend+ColorPicker.swift
stackotter: In this case it is guaranteed to be the same Swift object, because we just capture a weak `self` when setting the signal handler. But yeah I forgot that it's a subclass that you're using, so you'd have to cast.

I guess an unowned reference is fine because the parameter is literally a reference to the same instance, so it'll never be a nil reference.

Examples/Sources/ControlsExample/ControlsApp.swift
stackotter: The method? Only if it causes compilation issues, other it's probably harmless to leave it in

Sources/GtkBackend/GtkBackend+ColorPicker.swift
bbrk24: This has been updated now (and the GTK3 one as well).

Sources/WinUIBackend/WinUIBackend+ColorPicker.swift
bbrk24: To be fair, I didn't see any handling of this in `presentSheet(_:window:parentSheet:)` either.

Sources/WinUIBackend/WinUIBackend+ColorPicker.swift
bbrk24: Updated

Sources/Gtk3Backend/Gtk3Backend+ColorPicker.swift
```diff
@@ -0,0 +1,72 @@
...
+        _ colorPicker: Widget,
+        supportsOpacity: Bool,
+        environment: EnvironmentValues,
+        onChange: @escaping (SwiftCrossUI.Color.Resolved) -> Void
+    ) {
+        let colorButton = colorPicker as! CustomColorButton
+        colorButton.sensitive = environment.isEnabled
+        colorButton.useAlpha = supportsOpacity
+        colorButton.colorSet = {
+            let rgba = ($0 as! CustomColorButton).rgba
```
stackotter: Give the closure parameter a name given that this is a non-trivial closure (same for the other instance of this in GtkBackend

stackotter (changes requested):
Just one tiny requested change, but looks good other than that

stackotter (dismissed):
Thanks for fixing that up

Sources/WinUIBackend/WinUIBackend+ColorPicker.swift
stackotter: Ah yeah I mispoke, looks like only the alerts implementation supports nested sheets in WinUIBackend

stackotter (approved):

PR #741, opened 2026-09-03
Respect locale in Android DatePicker

### LINKED ISSUE #738: [AndroidBackend] Date pickers should respect environment.locale
Currently, they use the device locale.

bbrk24:
Fixes #738 

I was going to do more backends but this took long enough. Other backends can be separate issues.


stackotter (approved):
Thanks for resolving this. The caching system is nice

PR #750, opened 2026-09-05
Migrate GtkBackend gradients away from css to cairo

MiaKoring:
Thats all that happens. My brain wasn’t working properly, so this was quite the feverdream, but it turned out not to be hard at all xD

- LinearGradient renders basically the same as before
- RadialGradient renders differently, but more correctly. 
  The css based implementation didn’t support endradius, which it now does. 
  It now matches AppKitBackend way closer.

The commits are probably a bit scrambled, because I had two branches and sometimes picked the wrong one when switching back from main.

I decided to resolve the colors outside of the draw function, so environment doesn’t need to be captured.

This is just preparing for supporting gradient filled shapes.

I only tested with mac gtk


stackotter (approved):
The code looks great, I'll just test it out on Linux myself before merging

PR #760, opened 2026-09-10
BackendFeatures.FocusHandling and BackendFeatures.FocusDisabling for Gtk

MiaKoring:
# Bugfix:
When setting focus programmatically (“randomize focus” in ControlFocusabilityTest) the FocusState was often reset back to nil immediately because after gaining focus a different widget recieved unfocus.

Resolved by only passing .unfocus as override when FocusState is Optional(nil) or false

# Gtk
- Generated `EventControllerFocus`
- added `setFocus(_:)` to `CustomRootWidget` (basically `NSWindow.makeFirstResponder(_:)`)
- Widget
  - Added 2 cssblocks `focusCSS` and `focusWithinCSS` (focus-within is for textfields, focuscss for other widgets it works for
  - added computed property `root` returning the `CustomRootWidget` it’s contained in
  - added func `makeKey()` making a widget gain focus
  - added Gobject property `canFocus` for hierarchical focus control

# GtkBackend
- `GtkCustomButton`
  - made internal, because the type is needed for custom handling of focusEffect
  - added new css class “focusEffectDisabled” for use by the modifier
- Focus stuff
  - added `FocusStateManager` (currently in the same file cause it’s relatively small, will separate on request)
  - added protocol conformances

# Interesting stuff
TextField, SecureTextField, DatePicker and Picker use enter and exit signals, as they give focus to inner widgets. Thanks gtk for supporting hierarchical control, looking at you AppKit
All other widgets use the notifyIsFocus signal. None of the signals capture any reference types, only ObjectIdentifier, so we should be good.

FocusContainer is just a container with canFocus enabled or disabled, no custom overriding stuff like in AppKit.

focusEffectDisabled is a bit annoying, cause there are tons of different (pseudo)classes to override in css and I didn’t feel like that would’ve been right to do with tons of different CSSBlocks. CSSBlock should be updated to better handle these cases. I only added an additional block for TextField and SecureTextField, cause I feel like they were the most important after buttons and the ones working anyway. 
Buttons get the new focusEffectDisabled class set and removed.

I believe that’s it, not that special. I tested on mac and Linux so we should hopefully be good lol. I’ll create an issue for supporting focusEffectDisabled on the remaining widgets


Examples/Sources/ControlFocusabilityTest/ControlFocusabilityApp.swift
```diff
@@ -59,6 +59,7 @@ struct ControlFocusabilityApp: App {
     @State var isFlavorPickerFocusable = true
     @State var isDatePickerStyleFocusable = true
     @State var isDatePickerFocusable = true
+    @State var allFocusEffectDisabled: Bool = false
```
stackotter: `allFocusEffectsDisabled` sounds more correct I think

Examples/Sources/ControlFocusabilityTest/ControlFocusabilityApp.swift
```diff
@@ -76,6 +77,9 @@ struct ControlFocusabilityApp: App {
                         .padding(.bottom, 20)
                         .focused($focusState, equals: 0)
 
+                        Toggle("Disable all focuseffects", isOn: $allFocusEffectDisabled)
```
stackotter: `Disable all focus effects`

Maybe an accidental German compound noun? 😄

stackotter: I think this toggle should either go at the start or the end of the view so that it's not too confusing that it doesn't accept focus

Sources/Gtk/Widgets/Widget.swift
```diff
@@ -35,7 +35,37 @@ open class Widget: GObject {
...
+                """)
+        }
+    }
+
+    /// The focus related CSS rules applied directly to this widget.
+    public lazy var focusWithinCSS: CSSBlock =
+        CSSBlock(forClass: "\(customCSSClass):focus-within")
+    {
+        didSet {
+            guard oldValue != focusCSS else { return }
```
stackotter: This guard should check `focusWithinCSS` instead of `focusCSS`

Sources/Gtk/Widgets/Widget.swift
```diff
@@ -35,7 +35,37 @@ open class Widget: GObject {
     public lazy var css: CSSBlock = CSSBlock(forClass: customCSSClass) {
         didSet {
             guard oldValue != css else { return }
-            cssProvider.loadCss(from: css.stringRepresentation)
+            cssProvider.loadCss(from: """
+                    \(focusWithinCSS.stringRepresentation)
+                    \(focusCSS.stringRepresentation)
+                    \(css.stringRepresentation)
+                """)
+        }
```
stackotter: This is repeated 3 times and it should probably go into a private helper method. Maybe `reloadCSS`?

Sources/Gtk/Widgets/Widget.swift
```diff
@@ -128,6 +158,33 @@ open class Widget: GObject {
         controller.registerSignals()
     }
 
+    public var root: Gtk.CustomRootWidget? {
+        guard let ptr = gtk_widget_get_root(widgetPointer) else { return nil }
```
stackotter, line 158: I have a feeling that this is actually returning a window pointer and not a pointer to our CustomRootWidget. Our CustomRootWidget is just a subclass of Widget and it does not implement the Root interface. It just uses `Root` in its name because it's intended to be the root widget of a window, but the window is still the 'root' of the hierarchy according to Gtk. I also did not know that `Root` was a thing in Gtk; definitely makes the naming a bit confusing (although I think it still does an ok job at making the distinction between a root *widget* and the root of a widget hierarchy).

To resolve this, you probably have to either cast it to a window (making use of the assumption that SwiftCrossUI doesn't use other widget hierarchy roots) or create a `Root` protocol and an `AnyRoot` class that implements it. The protocol would just inherit from `GObject` with no other requirements, and then implement `setFocus(to:)` in an extension.

Sources/Gtk/Widgets/Widget.swift
```diff
@@ -128,6 +158,33 @@ open class Widget: GObject {
...
+        return CustomRootWidget(ptr)
+    }
+
+    /// Makes the widget the key view in the window it belongs to.
+    /// Equivalent to `NSWindow/makeFirstResponder(_)`.
+    public func makeKey() {
+        /// Wrap in g idle to make sure focus runs when free so we can know for sure it will get focused.
```
stackotter: What does this comment meant? What do you mean by free? (should also only be a double slash comment)

Sources/GtkBackend/Features/FocusHandling.swift
```diff
@@ -0,0 +1,109 @@
...
+
+    public func updateFocusContainer(
+        _ widget: Gtk.Widget,
+        focusability: Focusability
+    ) {
+        widget.canFocus = focusability != .disabled
+    }
+
+    public func setFocusEffectDisabled(on widget: Gtk.Widget, disabled: Bool) {
+        guard !(widget is GtkCustomButton) else {
```
stackotter: imo this would work better as an if/else. For some reason this guard feels a bit opposite to me, but it's hard to explain. Maybe it's because it feels more like special handling than a failure case/fallback/fast path?

Sources/GtkBackend/Features/FocusHandling.swift
```diff
@@ -0,0 +1,109 @@
...
+        let cssProperty = CSSProperty(key: "outline", value: "none")
+        if disabled {
+            if widget is Entry {
+                widget.focusWithinCSS.set(property: cssProperty)
+            }
+            widget.focusCSS.set(property: cssProperty)
+            return
+        }
+        widget.focusCSS = CSSBlock(forClass: widget.focusCSS.cssClass)
+        widget.focusWithinCSS = CSSBlock(forClass: widget.focusWithinCSS.cssClass)
```
stackotter, line 83: What is the purpose of this `disabled` case. Why does being disabled require inserting the property, while being enabled requires overriding the whole block? And what stops one of these approaches working in both cases? If the set vs override distinction isn't necessary then you could combine this logic so that you only set focusCSS once and focusWithinCSS once. If you need the Entry edgecase then you can place focusWithinCSS inside a `!disabled || widget is Entry` if statement.

The reasoning for the `Entry` edge case should also be documented with a comment

Sources/GtkBackend/Features/FocusHandling.swift
```diff
@@ -0,0 +1,109 @@
...
+        if let controller = widget.eventControllers
+            .first(where: { $0 is EventControllerFocus }) as? EventControllerFocus
+        {
+            focusController = controller
+        } else {
+            focusController = EventControllerFocus()
+            widget.addEventController(focusController)
+        }
+
+        // Some widgets focus is managed by descendants.
```
stackotter: `Some widgets have focusable children, which requires special handling.`

Sources/GtkBackend/Features/FocusHandling.swift
```diff
@@ -0,0 +1,109 @@
...
+
+                self.lastFocusedWidget = nil
+                observers.forEach { observer in
+                    observer.didLoseFocus()
+                }
+            }
+            return
+        }
+
+        focusController.notifyIsFocus = { [weak self] focusController, _ in
```
stackotter: Do `enter` and `leave` also work here? If so, I think it's best to streamline this and just handle everything that way so we don't have to have a special case for `Entry` and co

stackotter (changes requested):
Just found a couple of issues in the Gtk bindings changes, and maybe some ways to simplify things, but the base of the PR is remarkably simple (I love SCUI's backend system!)

stackotter (changes requested):
Just found a couple of issues in the Gtk bindings changes, and maybe some ways to simplify things, but the base of the PR is remarkably simple (I love SCUI's backend system!)

Sources/Gtk/Widgets/Widget.swift
MiaKoring: okay, but soon it’s going to disappear anyway 🤞

Sources/Gtk/Widgets/Widget.swift
MiaKoring: iirc when I first implemented things I ran into the issue that focus sometimes was lost wen other gtk things ran after. I have not encountered it again since adding idle.

g idle add moves something to be executed after everything higher priority.

Meaning it only gets focused after gtk is done drawing it.

Sources/GtkBackend/Features/FocusHandling.swift
MiaKoring: It’s not so much an edge case as it is one of many widget specific handling cases.

I’m replacing it with an empty block because that’s currently afaict the cleanest way to completely remove something. and as it has to get reloaded anyway there’s not really a difference performance wise.

I will improve that once I get to rewriting css block and bring support for the remaining widgets that currently don’t support it.

Examples/Sources/ControlFocusabilityTest/ControlFocusabilityApp.swift
MiaKoring: yup xD
so weird that english doesn’t do this

Sources/Gtk/Widgets/Widget.swift
MiaKoring: I just made it Window. No casting involved 😅

In this case the underlying type is GtkApplicationWindow, but GtkApplicationWindow inherits from window so it should be fine? and maybe it’s different with multiple windows. Is ApplicationWindow the primary window and other windows are just offsprings?

Sources/GtkBackend/Features/FocusHandling.swift
MiaKoring: I made it an if but kept the early return, I just think that’s nicer and pretty clear: Button does its thing, rest something else. Imo else would make it less readable due to additional indentation.

Sources/GtkBackend/Features/FocusHandling.swift
MiaKoring: fair, I made it separate cases because it’s more efficient memory wise. (2 closures vs 1, 2 self captures vs 1)

But enter and leave works too, and I’m okay with streamlining it.

Sources/Gtk/Widgets/Widget.swift
stackotter: There is a `GtkRoot` implementation that doesn't inherit from `GtkWindow`, and that's `GtkDragIcon`. Given that these bindings methods may be used for usecases outside of focus handling in future, I think that we should make sure that they're safe under all conditions. If you called this on a child of a drag icon and then used a window-specific method on the returned value, then you'd probably crash (hopefully via an assert, but perhaps not).

I think it's best to create a new `AnyRoot` class that just has your one root widget method, and then return an instance of that so people can only run the methods that we *know* are safe to run.

Sources/GtkBackend/Features/FocusHandling.swift
```diff
@@ -4,15 +4,10 @@ import CGtk
...
-            if focus == .unfocused {
-                widget.root?.setFocus(to: nil)
-            }
-            return
-        }
-
-        if focus == .focused {
+        if focus == .focused && ObjectIdentifier(widget) != lastFocusedWidget {
             widget.makeKey()
+        } else if focus == .unfocused {
```
stackotter, line 9: This refactor from guard did kinda change the logic slightly, the backend will now unfocus all widgets whenever it processes a widget with a target state of 'unfocused' even if the widget was already unfocused. Is that intended?

Sources/GtkBackend/Features/FocusHandling.swift
stackotter: Apologies, I completely misinterpreted this code that first time. I didn't realise that the last two lines were just clearing the blocks.

Unrelated to my initial comment, I believe that this should be an if/else scenario instead of an early return. It's one case for disabled and another for enabled.

stackotter (changes requested):
Thanks for addressing those comments, I've just got a few small suggestions this time.

Sources/GtkBackend/Features/FocusHandling.swift
MiaKoring: yup, that is intended, found that issue while testing.

(I unfortunately am not entirely sure anymore which issue that was, but I know it is correct that way)

PR #762, opened 2026-09-10
add deprecated Environment modifier and typealias

### LINKED ISSUE #716: Add deprecated EnvironmentObject modifier
We have no use for the EnvironmentObject modifier, but introducing a simple deprecated implementation of it that points people towards `@Environment(MyType.self) var store` would ease porting of existing SwiftUI code.

MiaKoring:
Resolves #716


stackotter (approved):
Thanks for resolving that! A nice simple fix

PR #763, opened 2026-09-10
fix calling onSubmit on focus loss

### LINKED ISSUE #713: AppKitBackend: onSubmit triggered when textfield loses focus
AppKitBackend's TextField implementation erroneously triggers `onSubmit` when the text field loses focus. This is because the underlying NSTextField's `action` gets triggered when the user hits enter OR when the field loses focus.

MiaKoring:
resolves #713


stackotter (approved):
Sweet!

PR #764, opened 2026-09-10
Adds linelimit support to dummybackend

### LINKED ISSUE #502: DummyBackend: Implement lineLimit modifier
DummyBackend currently doesn't appear to respect the `lineLimit` modifier

MiaKoring:
resolves #502 
first takes the min of lineCount an lineLimit.limit to, well, limit.
if it should reserve space it takes the max of the limited value and the limit.

I didn’t create a test cause it would feel a bit pointless to test a test. But there’s an issue for it and now it’s resolved lol


stackotter (approved):
Thanks!

PR #774, opened 2026-09-14
Remove Action

### LINKED ISSUE #767: AppKitBackend: Migrate away from onAction-based action handling
Many of AppKitBackend's control implementation (such as its button implementation) use `onAction` to receive actions (ie clicks in the case of buttons). This is a utility that I implemented early on and it's probably not very efficient. It uses object associations to store the actions for each control.

```swift
// TODO: Update all controls to use this style of action passing, seems way nicer
//   than the existing associated keys based approach. And probably more efficient too.
// Source: https://stackoverflow.com/a/36983811
final class Action: NSObject {
```

MiaKoring:
Resolves #767 
There were less things using the action class than I expected.
I kept the overall shape the same, just removing Action and replacing with closure and objc function.

I believe this also solves a memory leak we didn’t know. The toggle action captured a strong reference to the item and the item held a strong reference to the action. I made it unowned now because there should be now way it’s ever called while not alive, but I can change it to weak if we prefer to not even crash should this weird case happen.


stackotter (approved):
Looks good, thanks for cleaning that up!

PR #775, opened 2026-09-14
NSCustomButton now doesn't have button background anymore when not needed

### LINKED ISSUE #755: Remove NSButtonBackground of NSCustomButton when style doesn’t need it

MiaKoring:
resolves #755 
pretty self explanatory I believe. The background is not necessarily needed and therefore now only added when it is.


stackotter (approved):
Nice and simple!

PR #776, opened 2026-09-14
Gtk binding lint

### LINKED ISSUE #769: GtkCodeGen: Regenerate bindings in CI
This is a follow on from #259 (that issue mentioned this improvement, but it really should've been a separate issue)

We should regenerate bindings in CI to make sure that they've been generated correctly and that the generated files haven't been hand-edited.

MiaKoring:
resolves #769 

adds a new thing running on ubuntu.

I hope I did the only-run-on-changes thing correctly, because in my fork’s CI the not running anything else stuff didn’t exactly work. But it runs and fails if you add some bs in the generated files. I believe I got all files/directories relevant for gtk code gen. I didn’t include gyb/the full generate.sh in the latest change cause there’s some formatter weirdness going on (could probably be fixed by running format once it’s done)

I didn’t check changes to the submodules.


.github/workflows/build-test-and-docs.yml
```diff
@@ -141,6 +144,30 @@ jobs:
     - name: Run APITool stub generation
       run: ./Scripts/generate_porting_kit.sh
 
+  check-generated-gtk-bindings:
+    runs-on: ubuntu-24.04
+    needs: check_file_changes
+    if: ${{ needs.check_file_changes.outputs.gtk-bindings == 'true' }}
```
bbrk24, line 150: TIL you can have hyphens in output variable names.

.github/workflows/build-test-and-docs.yml
```diff
@@ -28,6 +28,8 @@ jobs:
             paths: 'Sources/WinUIBackend/ Sources/WinUIInterop/ vcpkg.json'
           - platform: apitool
             paths: 'APITool/ Scripts/analyze_api.sh Scripts/generate_porting_kit.sh'
+          - platform: gtk-bindings
+            paths: 'Sources/GtkCodeGen Sources/Gtk/Generated Sources/Gtk3/Generated Scripts/generate_gtk.sh Scripts/generate_gtk3.sh Scripts/generate_gtk4.sh Scripts/generate.sh'
```
bbrk24: Should `.gitmodules` be included in this list?

bbrk24 (commented):
> I hope I did the only-run-on-changes thing correctly, because in my fork’s CI the not running anything else stuff didn’t exactly work. 

We (intentionally) re-run everything if the workflow YAML itself changes, so it can't even be easily tested in this PR. That being said, the way you're handling it looks reasonably correct.

stackotter (approved):
Looks good! Thanks for implementing that. It'll help speed up Gtk-related code review quite a bit

PR #779, opened 2026-09-15
Made menus render full color

### LINKED ISSUE #777: AppKitBackend: Picker items appear dim
Picker items seem dim on macOS.

Image

MiaKoring:
resolves #777

NSMenuItems (the popup things of Picker and Menu) now render their full color, overriding appkit’s vibrancy stuff.

An exception are Text() in Menu, but that isn’t caused by vibrancy but the lack of an action, since in reality it’s just a button.

The benefit: it renders exactly like appkit normally would and supports colorScheme overrides. And it matches Menu buttons
The caveat: overriding the color doesn’t adapt just like the default, but that’s in the user’s hands.

The reasons: Using a non-adaptive NSColor makes AppKit apply vibrancy stuff, which makes our supplied white grey-ish.
Pretending to be adaptive with the environment dictated color makes the color render correctly, but in light mode it doesn’t change anymore for highlighted, which it does in appkit.

Therefore I decided to use .textColor if `suggestedForegroundColor` matches the colorscheme’s default and vibrancyOverride with the requested color for everything else. That way you get the correct behaviour without losing the ability to override it.


stackotter (dismissed):
The code looks good, and this fix appears to work basically as could as I would've hoped in practice! I think I misunderstood you when we were talking about this on Discord, cause I don't see anything wrong with this way of doing things (other than not making the hover color have correct contrast for custom colors, but that's a tiny issue imo). I think I hadn't realized that you had found a way to disable vibrancy for arbitrary colors

stackotter:
Happy to merge once this PR is updated to get the fixed CI from main

stackotter:
Oh no, a conflict has appeared 😅

stackotter (approved):

PR #780, opened 2026-09-15
Improve Gtk Binding check action

MiaKoring:
Now only checks whether there are changes in one of the two “Generated” directories.
I’m making this change, because it failed due to a change in Package.resolved in one of my PRs. And I don’t think that should happen.


stackotter (approved):
Thanks for fixing that up!

PR #783, opened 2026-09-17
Dummybackend splitup

MiaKoring:
I split up dummybackend following the same pattern as appkitbackend except Helpers still being DummyBackend namespaced.

There’s not really more to it. I hope it doesn’t take you too long to review haha


Sources/DummyBackend/Widgets/Button.swift
```diff
@@ -0,0 +1,8 @@
+import SwiftCrossUI
```
stackotter, line 1: Put a blank line after this import

Sources/DummyBackend/Widgets/Button.swift
```diff
@@ -0,0 +1,8 @@
...
+extension DummyBackend {
```
stackotter, line 3: I think widgets should go in their own `Widgets` directory instead of `Helpers`, because they're basically part of our fake UI framework (rather than being helper wrappers like the widget subclasses in other backends).

Sources/DummyBackend/Features/Core.swift
```diff
@@ -0,0 +1,26 @@
...
+
+    public func computeRootEnvironment(defaultEnvironment: EnvironmentValues) -> EnvironmentValues {
+        defaultEnvironment
+            .with(\.appPhase, appPhase)
+    }
+
+    public func setRootEnvironmentChangeHandler(
+        to action: @escaping @Sendable @MainActor () -> Void
+    ) {}
+
```
stackotter: Remove empty blank line

stackotter (changes requested):
I just have a few small requests, but other than those it looks great

stackotter (approved):

PR #784, opened 2026-09-17
DummyBackend test expansion

MiaKoring:
Follows #783 

### ViewGraphHelpers
- added `computeLayoutAndNode(_:)` for cases you need both the layout result and the `ViewGraphNode`.
- added `computeAndCommitExisting(_:)` for recomputing and commiting a node
#### DummyBackend.Widget extension
- added `first(where:)` returning nil if not found
- added `locateDescendant(where:)` as a throwing alternative
- added`filter(_:)` for retrieving all widgets matching something. (isn’t used yet, but thought it would be useful)
### TextRenderingHelpers
- added an easy way to get the default font
- added simplified access to `size(of text: …)`

### Updated tests to use helpers:
- `ForEachTests`
- `StackContextRendersDifferently`

### Added new tests
Added 2 new Tests for HStack and VStack layout with one axis unspecified proposed size.
I decided to use parameterized tests for HStack and VStack separately to get some savings without the test getting to confusing.


stackotter:
I'll hold off on reviewing this until #783 is merged, because otherwise I'm just reviewing #783 twice 😅

